PR #34, opened 2024-03-13
fix(tests): make all the tests pass

timraymond:
When cloning the repo, users expect to (or should expect to) be able to run `go test ./...` and run all the tests. If running all the tests on a fresh clone of main doesn't work, then it's hard to contribute. This ensures that all of the tests run with `go test ./... -short`. The short flag is used because there are some tests that are "costly" in terms of time or environment. This is a well-known flag in Go for bypassing these tests to provide a good onramp for new contributors.

Also, where one test was failing documentation has been improved and failing tests were fixed.


anubhabMajumdar (commented):
https://github.com/microsoft/retina/pull/35/files

You would need to remove some code from the `retina-test-image`  to get it working.

timraymond:
I've temporarily based this off #47 since I believe that this will fix CI here. Once that passes, and #47 merges, I will rebase off main. Disabling auto-merge as a result.

PR #38, opened 2024-03-14
feat: publish to ghcr

rbtr:


.github/workflows/images-ghcr.yaml
```diff
@@ -0,0 +1,172 @@
+name: Build and Publish Retina Container Images [GHCR]
```
anubhabMajumdar, line 1: Don't think we should maintain two `images*.yaml` files; I think we should update `IMAGE_REGISTRY`
in the make file to make ghrc.io the default -
https://github.com/microsoft/retina/blob/91de0eb612910a61568cb4f2a84affdcf6ce70ce/Makefile#L31

anubhabMajumdar (changes requested):

.github/workflows/images-ghcr.yaml
```diff
@@ -0,0 +1,172 @@
...
+      - name: Log in to registry
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
+
+      - name: Build/Push Images
+        shell: bash
+        run: |
+          set -euo pipefail
+          echo "TAG=$(make version)" >> $GITHUB_ENV
+          make retina-image-win \
+            IMAGE_REGISTRY=ghcr.io/microsoft/retina \
```
anubhabMajumdar, line 78: Can we use some bash command to figure this out dynamically?
Something like 
```
IMAGE_REGISTRY=ghrc.io/$(git remote get-url origin | sed -e 's/git@github.com://g' -e 's/\.git$//')
```

In the documentation we can recommend you fork our repository and this command would automatically set the repository to `<username>/retina`.

.github/workflows/images-ghcr.yaml
rbtr: updated

.github/workflows/images-ghcr.yaml
rbtr: added, and also updated the GHA to inject the org/repo from context so that forks will upload to _their_ GHP automatically

Makefile
```diff
@@ -188,16 +187,19 @@ retina-capture-workload: ## build the Retina capture workload
...
-KUBECTL_RETINA_IMAGE=kubectl-retina
-RETINA_OPERATOR_IMAGE=retina-operator
-RETINA_INTEGRATION_TEST_IMAGE=retina-integration-test
-RETINA_PROTO_IMAGE=retina-proto-gen
-RETINA_GO_GEN_IMAGE=retina-go-gen
-KAPINGER_IMAGE = kapinger
+IMAGE_REGISTRY	?= ghcr.io
+IMAGE_NAMESPACE	?= $(git remote get-url origin | sed -e 's/git@github.com://g' -e 's/\.git$//') # attempts to extract the upstream for image namespacing
+
+RETINA_BUILDER_IMAGE			= $(IMAGE_NAMESPACE)/retina-builder
```
anubhabMajumdar, line 193: This is more of a nit/opinion, but can we not have separate `IMAGE_NAMESPACE` and `IMAGE_RESGISTRY`?

My thinking is like this
```
IMAGE_REGISTRY ?= ghcr.io/$(git remote get-url origin | sed -e 's/git@github.com://g' -e 's/\.git$//')

RETINA_BUILDER_IMAGE = retina-builder
...
```
The benefit is any user who want's to use any other registry can update just the `IMAGE_REGISTRY` env variable and commands would simply work.

I think this will make pipelines simpler as well?

rbtr: wontfix in this change, we can evaluate simplifying the image name flow in the future

anubhabMajumdar (approved):
LGTM!

PR #41, opened 2024-03-14
fix: windows linter issue

nddq:
Some of the files the windows linter is complaining about are from linux-only plugins. Adding `_linux` suffix to each of them should resolve it for now.


rbtr (approved):

rbtr:
lint is passing so let's go 🚀

PR #43, opened 2024-03-14
fix: increase CodeQL timeout

nddq:
Increase timeout to accommodate codeQL runtime on windows.


rbtr (approved):
Windows CodeQL passed let's go 🚀

Issue #82, opened 2024-03-20
API Server Latency buckets are skewed in large scale clusters


vakalapa:
**Describe the bug**

On this large-scale cluster, all latencies are in the +inf category. Discussed increasing max bucket so that we have better information if latency is larger than 4.5ms (highest bucket currently).

![image](https://github.com/microsoft/retina/assets/9221639/cafb608e-2a0f-49b1-aafb-99a3d471dcaa)

Discussed making bucket width 1ms, starting at 0.5ms so that we don't get a lot of 0ms counts

cc @huntergregory  @anubhabMajumdar

rbtr:
@iamrajhans i see your linked PR, do you want this issue too?

iamrajhans:
@rbtr yeah

PR #96, opened 2024-03-20
fix: change bucket width and start value

iamrajhans:
Related issue #82

### DIFF
--- a/pkg/module/metrics/latency.go
+++ b/pkg/module/metrics/latency.go
@@ -35,8 +35,8 @@ const (
 	TTL                             time.Duration = 500 * time.Millisecond
 	LIMIT                           uint64        = 100000
 	// Bucket size.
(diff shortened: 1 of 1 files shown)


pkg/module/metrics/latency.go
```diff
@@ -35,8 +35,8 @@ const (
 	TTL                             time.Duration = 500 * time.Millisecond
 	LIMIT                           uint64        = 100000
 	// Bucket size.
-	start = 0
-	width = 0.5
+	start = 0.5
+	width = 1
```
anubhabMajumdar, line 39: Increasing the width will clump a lot of values into one bucket. I would recommend increasing the count to accommodate for results before `+inf`.

anubhabMajumdar (commented):
Thanks for the PR.

Issue #98, opened 2024-03-20
Duplicate import of flow library


anubhabMajumdar:
Duplicate import - https://github.com/microsoft/retina/blob/debc18823a6535deb9c3d4f6d18dd7aca3c2916c/pkg/module/metrics/latency.go#L15

Issue #93, opened 2024-03-20
Evaluate security context/caps 


rbtr:
Retina has CAP_NET_ADMIN, SYS_ADMIN, and others. 
Evaluate the caps and make sure we are adding the minimum required permissions

nddq:
I know that CAP_NET_ADMIN is needed for `packetparser`, SYS_ADMIN for root access I think is for loading the eBPF program into the kernel since we need root access, but if that's all we are doing with that CAP then I think we could use CAP_BPF instead?IPC_LOCK is used for mmap() calls and not sure about SYS_RESOURCE

PR #101, opened 2024-03-20
fix:fix duplicate import of flow library #98

Dibyanshuaman:


pkg/module/metrics/latency.go
```diff
@@ -12,7 +12,7 @@ import (
 	"time"
 
 	"github.com/cilium/cilium/api/v1/flow"
-	v1 "github.com/cilium/cilium/api/v1/flow"
+
```
rbtr, line 15: the linter will not like this empty line, it should be removed

rbtr (changes requested):

rbtr:
thanks for the contribution. please amend your commit message so that it is in [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) format, and add "fixes #93" to this PR description to link back to the issue that this fixes

pkg/module/metrics/latency.go
vakalapa: Also there are references of this v1 package name which needs to be corrected as well.

FZhg:
> thanks for the contribution. please amend your commit message so that it is in [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) format, and add "fixes #93" to this PR description to link back to the issue that this fixes

Hi @rbtr, it should fix #98

pkg/module/metrics/latency.go
```diff
@@ -11,7 +11,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/cilium/cilium/api/v1/flow"
```
anubhabMajumdar, line 14: I believe there are other references to `flow`, example - https://github.com/microsoft/retina/blob/96ff934a608fcce848d66c31ad8e54afbed9e00e/pkg/module/metrics/latency.go#L60

anubhabMajumdar (changes requested):

PR #109, opened 2024-03-21
feat: run e2e in azure cluster. 

jimassa:
This PR enables the execution of end-to-end integration tests for Retina within a Kubernetes cluster. While the current implementation is designed for an Azure Kubernetes Service (AKS) cluster, it also supports execution in other cloud environments.

Please note that these tests cannot be run on a local cluster such as Kind due to its lack of support for packetparser eBPF programs.

The gh action needs to be run manually on pr before being approved. 

Running tests can be found here:

https://github.com/microsoft/retina/actions/workflows/integration.yaml


.github/workflows/integration.yaml
```diff
@@ -1,74 +1,184 @@
-name: Integration Tests For Retina
+name: Build images and run E2E tests.
```
anubhabMajumdar, line 1: Since the images are built and pushed in the `Build` pipeline, why duplicate that effort? We can add steps to run it in `images.yaml` itself?

anubhabMajumdar (changes requested):

test/e2e/framework/kubernetes/install-retina-helm.go
```diff
@@ -58,6 +58,11 @@ func (i *InstallHelmChart) Run() error {
 	chart.Values["image"].(map[string]interface{})["pullPolicy"] = "Always"
 	chart.Values["operator"].(map[string]interface{})["tag"] = tag
 
+	// update helm chart to use images from acnrepo.azurecr.io when running e2e tests
+	chart.Values["image"].(map[string]interface{})["repository"] = "acnpublic.azurecr.io/microsoft/retina/retina-agent"
```
anubhabMajumdar, line 62: Do we need to hardcode the `IMAGE_REGISTRY/IMAGE_NAMESPACE`?

.github/workflows/integration.yaml
jimassa: if we run it in images.yaml, every pr will run this tests, we want it to only run when triggered by owners, but we want the build pipeline to run for every PR.

test/e2e/framework/kubernetes/install-retina-helm.go
jimassa: Fixed

.github/workflows/images.yaml
```diff
@@ -96,4 +96,4 @@ jobs:
           echo "TAG=$(make version)" >> $GITHUB_ENV
           make retina-operator-image \
             IMAGE_NAMESPACE=${{ github.repository }} \
-            PLATFORM=${{ matrix.platform }}/${{ matrix.arch }}
+            PLATFORM=${{ matrix.platform }}/${{ matrix.arch }}
```
anubhabMajumdar, line 99: Required?

.github/workflows/integration.yaml
anubhabMajumdar, line 1: File a Issue to remove duplicate code please.

Makefile
```diff
@@ -221,6 +221,7 @@ container-docker: buildx # util target to build container images using docker bu
 		-t $(IMAGE_REGISTRY)/$(IMAGE):$(TAG) \
 		$(CONTEXT_DIR)
 
+
```
anubhabMajumdar, line 224: Required?

test/e2e/framework/kubernetes/install-retina-helm.go
```diff
@@ -13,8 +13,10 @@ import (
 )
 
 const (
-	createTimeout = 240 * time.Second // windpws is slow
-	deleteTimeout = 60 * time.Second
+	createTimeout  = 240 * time.Second // windpws is slow
+	deleteTimeout  = 60 * time.Second
+	imageRegistry  = "acnpublic.azurecr.io"
+	imageNamespace = "microsoft/retina"
```
anubhabMajumdar, line 19: You are still hardcoding these values, is this necessary? No way to make them parameters?

.github/workflows/integration.yaml
jimassa: already field. https://github.com/microsoft/retina/issues/137

test/e2e/framework/kubernetes/install-retina-helm.go
jimassa: updated the code to set those form pipeline as env.

test/e2e/framework/kubernetes/install-retina-helm.go
```diff
@@ -42,6 +42,16 @@ func (i *InstallHelmChart) Run() error {
 
 	tag := os.Getenv(i.TagEnv)
 
+	imageRegistry := os.Getenv("IMAGE_REGISTRY")
```
anubhabMajumdar, line 45: Mat can comment on this more, but given it's a go code, I would prefer test flags than environment variables. This way we can validate immediately if correct flags were passed or not.

matmerr: No strong opinion, but I don't consider image registry to be a value that's frequently changed, so for those type of things I usually stick to env's, as it's easier to add to some env.json or .bashrc and forget about it. Other things that impact output or behavior of the test, yeah flags for sure

anubhabMajumdar: Image registry can have some default values and go test help can show the default values and validate stuff at the very start.

matmerr (approved):

test/e2e/framework/kubernetes/install-retina-helm.go
jimassa: It is changed into flags, for image-registry, image-namespace and tag provided by the user will provide those. Help or not providing the value will bring up the doc on expected flags.

anubhabMajumdar (approved):

PR #123, opened 2024-03-22
fix(enricher): Streamline the pipeline to send flows from plugins to enricher

anubhabMajumdar:
## Description

Currently we have to many ways to extract out the events from plugins, and each plugin is responsible to init/check/send the data to enricher. Instead, setup a tracker module to "track" events generated and decide what to do with it. Currently, we simply forward to enricher.

## Testing done:

- Enabled in local context mode
- Can observe TCP, DNS and Drop advanced metrics
-  Basic mode works as expected
- Unit tests passing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,8 @@ RETINA_DIR = $(REPO_ROOT)/controller
 KUBECTL_RETINA_BUILD_DIR = $(OUTPUT_DIR)/kubectl-retina
 OPERATOR_DIR=$(REPO_ROOT)/operator
 CAPTURE_WORKLOAD_DIR = $(REPO_ROOT)/captureworkload
+COVER_PKG ?= .
+
 
 KIND = /usr/local/bin/kind
 KIND_CLUSTER = retina-cluster
@@ -329,6 +331,7 @@ manifest:
 # Make sure the layer has only one directory.
 # the test DockerFile needs to build the scratch stage with all the output files 
 # and we will untar the archive and copy the files from scratch stage
+.PHONY: test-image
 test-image: ## build the retina container image for testing.
 	$(MAKE) container-docker \
 			PLATFORM=$(PLATFORM) \
@@ -338,12 +341,12 @@ test-image: ## build the retina container image for testing.
 			CONTEXT_DIR=$(REPO_ROOT) \
 			TAG=$(RETINA_PLATFORM_TAG)
 
-COVER_PKG ?= .
-
+.PHONY: test
 test: $(ENVTEST) # Run unit tests.
 	go build -o test-summary ./test/utsummary/main.go
 	CGO_ENABLED=0 KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use -p path)" go test -tags=unit -skip=TestE2E* -coverprofile=coverage.out -v -json ./... | ./test-summary --progress --verbose
 
+.PHONY: test-e2e
 coverage: # Code coverage.
 #	go generate ./... && go test -tags=unit -coverprofile=coverage.out.tmp ./...
 	cat coverage.out | grep -v "_bpf.go\|_bpfel_x86.go\|_bpfel_arm64.go|_generated.go|mock_" | grep -v mock > coveragenew.out
--- a/Makefile
+++ b/Makefile
@@ -346,7 +346,7 @@ test: $(ENVTEST) # Run unit tests.
 	go build -o test-summary ./test/utsummary/main.go
 	CGO_ENABLED=0 KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use -p path)" go test -tags=unit -skip=TestE2E* -coverprofile=coverage.out -v -json ./... | ./test-summary --progress --verbose
 
-.PHONY: test-e2e
+.PHONY: coverage
 coverage: # Code coverage.
 #	go generate ./... && go test -tags=unit -coverprofile=coverage.out.tmp ./...
 	cat coverage.out | grep -v "_bpf.go\|_bpfel_x86.go\|_bpfel_arm64.go|_generated.go|mock_" | grep -v mock > coveragenew.out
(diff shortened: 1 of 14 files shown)


pkg/track/events.go
```diff
@@ -0,0 +1,74 @@
...
+
+var (
+	once sync.Once
+	t    *Track
+)
+
+func New() *Track {
+	once.Do(func() {
+		var e *enricher.Enricher
+		if e = enricher.Instance(); e == nil {
```
nddq, line 35: will cause the old error when we tried to compare a nil interface with nil (this will never be true)

anubhabMajumdar: Instance returns struct pointer, not interface - https://github.com/microsoft/retina/blob/f93df8d2ed54ca8fb77bf40d3dbb314c6275aea0/pkg/enricher/enricher.go#L61

anubhabMajumdar: Also, enricher should absolutely be a singleton for the entire lifecycle of Retina. It's a key piece of the control plane that needs the ringbuffer to be set and all events to be funneled to the same. The enricher also serves the enriched data to endpoint as metrics. All of this is predicated on a single enricher instance.

pkg/track/events.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"context"
+	"sync"
+
+	v1 "github.com/cilium/cilium/pkg/hubble/api/v1"
+	"github.com/microsoft/retina/pkg/enricher"
+	"github.com/microsoft/retina/pkg/log"
+)
+
+type Track struct {
```
timraymond, line 15: Can you doc what a `Track` is intended to be?

pkg/track/events.go
```diff
@@ -0,0 +1,74 @@
...
+	ctx             context.Context
```
timraymond, line 16: `context.Context` shouldn't be stored in a struct: see: https://pkg.go.dev/context

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.

timraymond (changes requested):

pkg/track/events.go
anubhabMajumdar: Yes, `Start` take a context and lifecycle is tied to it. This variable is not actually needed, what I do need it the `cancelFunc` (in lieu of a separate done channel) to make sure I stop the track before I close the external channel.

pkg/track/events.go
anubhabMajumdar: +1

Makefile
```diff
@@ -338,12 +341,12 @@ test-image: ## build the retina container image for testing.
...
 			TAG=$(RETINA_PLATFORM_TAG)
 
-COVER_PKG ?= .
-
+.PHONY: test
 test: $(ENVTEST) # Run unit tests.
 	go build -o test-summary ./test/utsummary/main.go
 	CGO_ENABLED=0 KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use -p path)" go test -tags=unit -skip=TestE2E* -coverprofile=coverage.out -v -json ./... | ./test-summary --progress --verbose
 
+.PHONY: test-e2e
```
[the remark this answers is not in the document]
anubhabMajumdar: +1

pkg/plugin/dns/dns_linux_test.go
```diff
@@ -146,26 +137,31 @@ func TestRequestEventHandler(t *testing.T) {
...
-		utils.DNSType_QUERY, 0, event.DNSName, []string{event.QType}, 0, []string{},
-	)).Times(1)
-	d.enricher = mockEnricher
-
 	d.eventHandler(event)
 	after := value(c)
 	assert.Equal(t, after-before, float64(1))
+
+	// Test External channel.
+	em := EventMatched(utils.DNSType_QUERY, 0, "test.com", []string{"A"}, 0, []string{})
```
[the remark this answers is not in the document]
anubhabMajumdar: +1

pkg/plugin/dns/dns_linux_test.go
```diff
@@ -146,26 +137,31 @@ func TestRequestEventHandler(t *testing.T) {
...
+	select {
+	case ev := <-exCh:
+		assert.Assert(t, em.Matches(ev))
+	case <-time.After(3 * time.Second):
```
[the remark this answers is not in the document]
anubhabMajumdar: +1

pkg/plugin/dns/dns_linux_test.go
```diff
@@ -196,15 +192,18 @@ func TestResponseEventHandler(t *testing.T) {
...
 	d.eventHandler(event)
 	after := value(c)
 	assert.Equal(t, after-before, float64(1))
+
+	// Test External channel.
+	em := EventMatched(utils.DNSType_RESPONSE, 0, "test.com", []string{"A"}, 2, []string{"1.1.1.1", "2.2.2.2"})
+	select {
+	case ev := <-exCh:
+		assert.Assert(t, em.Matches(ev), "Expected event to match")
+	case <-time.After(3 * time.Second):
```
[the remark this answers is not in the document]
anubhabMajumdar: +1

pkg/plugin/dropreason/dropreason_linux_test.go
```diff
@@ -316,7 +310,8 @@ func TestDropReasonReadData_WithEmptyPerfArray(t *testing.T) {
 
 	// Start the drop reason routine in a goroutine
 	go func() {
-		dr.readEventArrayData()
+		err := dr.readEventArrayData()
+		assert.Nil(t, err, "Expected error but got nil")
```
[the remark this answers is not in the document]
anubhabMajumdar: You are right, I have reworked them.

pkg/track/events.go
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package track
```
anubhabMajumdar, line 4: We need to refactor the name to `TrackEvents`.

timraymond:
@anubhabMajumdar Looks like stalebot killed this, but I see you manually closed this as well. What was the reason here?

PR #125, opened 2024-03-22
chore: package kubectl retina as tar.gz (zip for windows) via goreleaser

nddq:
Publishing plugins to Krew requires the plugins to be packaged as `.zip` or `.tar.gz` archives. 

Sample release: https://github.com/nddq/retina/releases/tag/v0.0.3

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -26,8 +26,7 @@ builds:
     main: cli/main.go
 
 archives:
-  - format: tar.gz
-    name_template: "{{ .Binary }}_{{ .Version }}"
+  - name_template: "{{ .Binary }}-v{{ .Version }}"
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -26,9 +26,12 @@ builds:
     main: cli/main.go
 
 archives:
-  - format: binary
+  - format: tar.gz
     name_template: "{{ .Binary }}_{{ .Version }}"
```
rbtr, line 30: I noticed after I added this that this is `_` instead of `-`, can you correct it here?

rbtr: I think we should stick the `v` prefix back in also, not sure why goreleaser strips that for archives

rbtr (approved):

Issue #128, opened 2024-03-23
can't build binary by `make retina-binary` 


nayihz:
**Describe the bug**
```bash
# make retina-binary
package command-line-arguments
        imports github.com/microsoft/retina/pkg/plugin/packetforward
        imports github.com/microsoft/retina/pkg/plugin/packetforward/_cprog: C source files not allowed when not using cgo or SWIG: packetforward.c
package command-line-arguments
        imports github.com/microsoft/retina/pkg/plugin/packetforward
        imports github.com/microsoft/retina/pkg/plugin/packetforward/_cprog: C source files not allowed when not using cgo or SWIG: packetforward.c
prog.go:12:2: no required module provides package github.com/golang/mock/mockgen/model: go.mod file not found in current directory or any parent directory; see 'go help modules'
prog.go:14:2: no required module provides package github.com/microsoft/retina/pkg/plugin/packetforward: go.mod file not found in current directory or any parent directory; see 'go help modules'
2024/03/23 12:35:28 Loading input failed: exit status 1
exit status 1
pkg/plugin/packetforward/types_linux.go:31: running "go": exit status 1
```
**To Reproduce**
Just run `make retina-binary` on main branch.

**Expected behavior**
Create an `output` directory and build retina binary successfully. 

**Screenshots**

**Platform (please complete the following information):**

- OS: Ubuntu
- Kubernetes Version: [e.g. 1.22]
- Host: [e.g. AKS, KIND, self-host, etc]
- Retina Version:

**Additional context**

nddq:
I believe you have to disable `CGO_ENABLED` for this, so adding `CGO_ENABLED=0` before `make retina-binary` should work

PR #129, opened 2024-03-23
fix: failed to build binary by make retina-binary

nayihz:
Addresses issue https://github.com/microsoft/retina/issues/128


nayihz:

rbtr:
@nayihz while I believe that this fixes the issue, I am not convinced it is the correct solution. Why don't we encounter this issue when we build in our container? https://github.com/microsoft/retina/blob/main/controller/Dockerfile.controller#L38

timraymond:
TIL about `//go:build ignore`, thanks for the contribution 😄 This looks pretty good to me as-is... I was able to build this with `make retina-binary` without any additional ENVs (we had previously been setting `CGO_ENABLED=0`, but this seems better). I did find that there's a few object files left behind from compiling BPF... would you be interested in adding those to `.gitignore` as well? They are:

- `pkg/plugin/dropreason/kprobe_bpfel_x86.o`
- `pkg/plugin/filter/filter_bpfel_x86.o`
- `pkg/plugin/packetforward/packetforward_bpfel_x86.o`
- `pkg/plugin/packetparser/packetparser_bpfel_x86.o`

timraymond:
@rbtr It's because we set `CGO_ENABLED=0` here: https://github.com/microsoft/retina/blob/cf23ecb9a5ed25c787d1f6ec819518749dfb7774/controller/Dockerfile.controller#L30 . The go toolchain skips over its distaste for C here when CGO is disabled: https://github.com/golang/go/blob/e7bdc8819a19f06321d300719224abb2d8567b4a/src/cmd/go/internal/load/pkg.go#L2060-L2063

rbtr:
> @rbtr It's because we set `CGO_ENABLED=0` here:

@timraymond We also disable CGO in the Make target to build the binary, though:
https://github.com/microsoft/retina/blob/cf23ecb9a5ed25c787d1f6ec819518749dfb7774/Makefile#L148-L149

rbtr:
Actually if I'm reading your link correctly
`len(p.CFiles) > 0 && !p.UsesCgo()` means that we get this error when CGO **is disabled** and there are C files present, right? So the unexpected behavior seems to be that in the Dockerfile, CGO remains enabled despite the env var being set

nayihz:
> > @rbtr It's because we set `CGO_ENABLED=0` here:
> 
> @timraymond We also disable CGO in the Make target to build the binary, though:
> 
> https://github.com/microsoft/retina/blob/cf23ecb9a5ed25c787d1f6ec819518749dfb7774/Makefile#L148-L149

`export CGO_ENABLED=0` doesn't seem to take effect in makefile. We can build successfully when we add `CGO_ENABLED=0` before `go build` command.
```makefile
# makefile
retina-binary: ## build the Retina binary
	go generate ./...
#export CGO_ENABLED=0
	CGO_ENABLED=0 go build -v -o $(RETINA_BUILD_DIR)/retina$(EXE_EXT) -gcflags="-dwarflocationlists=true" -ldflags "-X main.version=$(TAG) -X main.applicationInsightsID=$(APP_INSIGHTS_ID)" $(RETINA_DIR)/main.go
```

nayihz:
Modifed the makefile to enable `CGO_ENABLED=0` and omit ".o" files. 
PTAL.

rbtr:
I'm afraid we need those .o files:
> Error: 10.40 pkg/plugin/filter/filter_bpfel_arm64.go:119:12: pattern filter_bpfel_arm64.o: no matching files found
Error: 10.40 pkg/plugin/dropreason/kprobe_bpfel_arm64.go:190:12: pattern kprobe_bpfel_arm64.o: no matching files found
Error: 10.40 pkg/plugin/packetforward/packetforward_bpfel_arm64.go:125:12: pattern packetforward_bpfel_arm64.o: no matching files found
Error: 10.40 pkg/plugin/packetparser/packetparser_bpfel_arm64.go:159:12: pattern packetparser_bpfel_arm64.o: no matching files found

https://github.com/microsoft/retina/actions/runs/8429451457/job/23089687153?pr=129

anubhabMajumdar (changes requested):
Thanks for the PR. Can you please remove the generated object files (not delete them from the repository, just the updated ones from your PR)? I think they don't need to be updated as part of the PR.

anubhabMajumdar (approved):

rbtr:
@nayihz can you please sign and sign-off your commits? `git commit -S -s --amend --no-edit` should do it

nayihz:
> @nayihz can you please sign and sign-off your commits? `git commit -S -s --amend --no-edit` should do it

Done.

rbtr:
Sorry @nayihz, you have correctly `Signed-off-By` in the commit message message, but it looks like you're missing the cryptographic signature on the commit itself.

This `Verified` in the commits page is what I'm looking for:
image 

It is a merge requirement that all commits be signed.
See #197

nayihz:
> This `Verified` in the commits page is what I'm looking for:

Done.  PTAL.

rbtr (approved):

rbtr:
looks good, thanks!

PR #157, opened 2024-03-27
fix(helm): remove duplicate nodeSelector

ronalddb89:
This duplicate will lead to problems using automated CD tools like fluxCD:

`2024-03-27T14:57:44.805Z error HelmRelease/retina.retina - Reconciler error Helm install failed: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors:
  line 119: mapping key "nodeSelector" already defined at line 27`


rbtr (approved):
lgtm, thanks for the fix!

PR #158, opened 2024-03-27
fix(grafana): repair node selection & metrics name

aslafy-z:
- Update 'dns' & 'clusters' Grafana dashboards to fix node selection
- Update 'pod-level' Grafana dashboard to fix metrics names, pod selection and datasource templating
- Update datasource variable to DS_PROMETHEUS convention

Note: The pod-level grafana dashboard still have some old metrics to update.

### DIFF
--- a/deploy/grafana/dashboards/clusters.json
+++ b/deploy/grafana/dashboards/clusters.json
@@ -107,7 +107,7 @@
     {
       "datasource": {
         "type": "prometheus",
-        "uid": "${datasource}"
+        "uid": "${DS_PROMETHEUS}"
       },
       "fieldConfig": {
         "defaults": {
@@ -167,11 +167,11 @@
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "${datasource}"
+            "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "sum by (cluster) (rate(networkobservability_forward_bytes[$__rate_interval]))",
+          "expr": "sum by (cluster) (rate(networkobservability_forward_bytes{instance=~\"($Nodes):[0-9]+\"}[$__rate_interval]))",
           "format": "table",
           "instant": true,
           "legendFormat": "__auto",
@@ -201,7 +201,7 @@
     {
       "datasource": {
         "type": "prometheus",
-        "uid": "${datasource}"
+        "uid": "${DS_PROMETHEUS}"
       },
       "fieldConfig": {
         "defaults": {
@@ -276,10 +276,10 @@
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "${datasource}"
+            "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "sum by (cluster) (rate(networkobservability_forward_bytes[$__rate_interval]))",
+          "expr": "sum by (cluster) (rate(networkobservability_forward_bytes{instance=~\"($Nodes):[0-9]+\"}[$__rate_interval]))",
           "legendFormat": "__auto",
           "range": true,
           "refId": "A"
@@ -291,7 +291,7 @@
     {
       "datasource": {
         "type": "prometheus",
-        "uid": "${datasource}"
+        "uid": "${DS_PROMETHEUS}"
       },
       "fieldConfig": {
         "defaults": {
@@ -351,11 +351,11 @@
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "${datasource}"
+            "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "sum by (cluster) (rate(networkobservability_drop_bytes[$__rate_interval]))",
+          "expr": "sum by (cluster) (rate(networkobservability_drop_bytes{instance=~\"($Nodes):[0-9]+\"}[$__rate_interval]))",
           "format": "table",
           "instant": true,
           "legendFormat": "__auto",
@@ -385,7 +385,7 @@
     {
       "datasource": {
         "type": "prometheus",
-        "uid": "${datasource}"
+        "uid": "${DS_PROMETHEUS}"
       },
       "description": "",
       "fieldConfig": {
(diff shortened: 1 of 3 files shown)


vakalapa:
@huntergregory  to review.

aslafy-z:
Please have a look @vakalapa @huntergregory @rbtr

rbtr:
hey @aslafy-z, thanks for working on this fix. I see that you have the DCO "signed-off-by" on all your commits, but [we also need a cryptographic sig](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--Sltkeyidgt) to be able to guarantee origin. Would you update these with a signature? Here's how: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
I'm looking for this "Verified" tag commits in the PR once you've done that:
image

aslafy-z:
@rbtr I just rebased, squashed and signed my commits :)

rbtr (approved):
lgtm, just need final sign-off from @huntergregory

deploy/grafana/dashboards/pod-level.json
huntergregory, line 1: Turns out that this dashboard is broken and not importable #271 (at least on my grafana setup). If you have a working version, would you actually be able to export it for sharing externally?
image

deploy/grafana/dashboards/pod-level.json
```diff
@@ -1062,25 +1062,39 @@
...
         "name": "pod",
         "options": [],
         "query": {
-          "query": "query_result(retina_adv_forward_count)",
+          "query": "query_result(networkobservability_adv_forward_count)",
           "refId": "StandardVariableQuery"
         },
         "refresh": 1,
-        "regex": "/.*_podname=\"([^\"]*).*/",
+        "regex": "/.*podname=\"([^\"]*).*/",
```
huntergregory, line 1097: nice catch. You must have been using the advanced local-context metric mode. Just noting how this prompted initial thoughts on #344

deploy/grafana/dashboards/clusters.json
```diff
@@ -107,7 +107,7 @@
     {
       "datasource": {
         "type": "prometheus",
-        "uid": "${datasource}"
+        "uid": "${DS_PROMETHEUS}"
```
huntergregory, line 110: > Update datasource variable to DS_PROMETHEUS convention

Do you have a link to this convention? I just glanced at the [top dashboard on Grafana.com](https://grafana.com/grafana/dashboards/1860-node-exporter-full/), and it uses `datasource` rather than `DS_PROMETHEUS` as the variable. Seems the same for the built-in dashboards in Azure's managed Grafana.

I'm also afraid that this might be a breaking change to someone's existing dashboard setup.

huntergregory (changes requested):
Hi @aslafy-z, sorry for the delay. I missed the notifications from March 😕. Thanks for your PR and the interest in improving these dashboards.

Do you mind updating the PR description with details/examples as needed for the bugs/fixes? Also, if you have a working pod-level dashboard, could you help fix #271?

Added more details in the comment, but I don't think it would make sense to filter by node in the "Fleet View". I'm also not sure that we can/should change `datasource` to `DS_PROMETHEUS`.

deploy/grafana/dashboards/clusters.json
aslafy-z: I've observed this "naming convention" widely used in recent years. If the dashboard needs to incorporate another type of datasource in the future, the name will clearly indicate the type.

aslafy-z: I can revert the change if you prefer.

deploy/grafana/dashboards/clusters.json
```diff
@@ -811,10 +811,10 @@
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "${datasource}"
+            "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "sum(rate(networkobservability_forward_count{direction=\"egress\", cluster=\"$cluster\", instance=~\"$Nodes\"}[$__rate_interval]))",
+          "expr": "sum(rate(networkobservability_forward_count{direction=\"egress\", cluster=\"$cluster\", instance=~\"($Nodes):[0-9]+\"}[$__rate_interval]))",
```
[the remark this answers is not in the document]
aslafy-z: The $Nodes variable has one or multiple node ips formated like `1.1.1.1` or `1.1.1.1|2.2.2.2`.
The instance label however has the node:port `1.1.1.1:1234`.
This edit makes it possible to select multiple nodes.

deploy/grafana/dashboards/pod-level.json
aslafy-z: I moved out from my previous job and has no access to a cluster where I can install retina right now. I'll try on a kind when back with my personal laptop in a few days and see how it goes.

deploy/grafana/dashboards/clusters.json
```diff
@@ -3717,17 +3717,17 @@
...
+        "definition": "label_values(kube_node_info{cluster=\"$cluster\"},internal_ip)",
         "hide": 0,
         "includeAll": true,
         "label": "Nodes",
         "multi": true,
         "name": "Nodes",
         "options": [],
         "query": {
-          "query": "label_values(kube_node_info{cluster=\"$cluster\"},node)",
+          "query": "label_values(kube_node_info{cluster=\"$cluster\"},internal_ip)",
```
cmergenthaler, line 3730: I think we could use named capture groups here to separate displayed value and used value, e.g.
``` json
       "query": {
          "qryType": 3,
          "query": "query_result(kube_node_info)",
          "refId": "PrometheusVariableQueryEditor-VariableQuery"
        },
        "refresh": 2,
        "regex": "/node=\"(?<text>[^\"]+)|internal_ip=\"(?<value>[^\"]+)/g",
```

This would allow users to select nodes based on names but filter panels by underlying IPs

huntergregory (changes requested):
The dashboard files have moved since #432. Could you please apply the changes in the new files? Also, I think @cmergenthaler makes a good suggestion about displaying the node names.

If changing `datasource.uid` would break someone's dashboard when updating the dashboard to the new version, then I would prefer we not change it. Either way, it would be nice to keep this PR's scope smaller and make a change to `datasource.uid` in another PR (there are also test files that depend on that value).

deploy/grafana/dashboards/clusters.json
huntergregory: I see. Added some thoughts here https://github.com/microsoft/retina/pull/158#pullrequestreview-2226013402

aslafy-z:
@rbtr @huntergregory @cmergenthaler
My priorities are shifting, and I'm now working on another project without any Azure clusters on hand. Please feel free to take over if you're able.

rbtr:
@ibezrukavyi

PR #159, opened 2024-03-27
feat: cache Go's build artifacts in dockerfiles

nddq:
This update caches Go’s build artifacts, speeding up image builds and avoiding rebuilds for unchanged packages, thus streamlining development.

### DIFF
--- a/controller/Dockerfile.controller
+++ b/controller/Dockerfile.controller
@@ -35,11 +35,10 @@ ARG VERSION
 # Set this if you want to enable Application Insights telemetry.
 ARG APP_INSIGHTS_ID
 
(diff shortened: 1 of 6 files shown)


rbtr (dismissed):

rbtr:
@nddq you need to amend and sign all previous commits (or squash to a single signed commit)

rbtr:
wonder if we could do this to CodeQL somehow so that the Windows build would take less than 90 minutes? 😭

nddq:
> @nddq you need to amend and sign all previous commits (or squash to a single signed commit)

i did add a `-S` to every `git commit` though? Is it not showing up?

rbtr:
> i did add a `-S` to every `git commit` though? Is it not showing up?

not on `cf4c52c92544e268f589997041bf5678c1844ebb`
you can see which ones have the `Verified` badge here https://github.com/microsoft/retina/pull/159/commits

rbtr (approved):

PR #162, opened 2024-03-27
feat: add container cleanup cronjob

rbtr:
Prunes any non-release image tags from the container registry daily at 00:00Z.
"non-release" tags include:
- SemVer prelease such as `:v1.2.3-6f5089b0` suffixed tags
- bare commit tags such as `:6f5089b0`

"release" tags include only bare SemVer such as `:v1.2.3`, specifically matching the RegEx `"^v(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)$"`, and will be skipped

### DIFF
--- a/.github/workflows/container-cleanup.yaml
+++ b/.github/workflows/container-cleanup.yaml
@@ -1,0 +1,33 @@
+name: Container Registry Cleanup
+
+on:
+  schedule:
+    - cron: "0 0 * * *" # every day at midnight
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  container-cleanup:
+    name: delete-images
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        image:
+          - "retina/charts/retina"
+          - "retina/kubectl-retina"
+          - "retina/retina-agent"
+          - "retina/retina-init"
+          - "retina/retina-operator"
+    steps:
+      # This is a fork of the official actions/delete-package-versions which adds GHCR image tag support
+      # https://github.com/actions/delete-package-versions/pull/104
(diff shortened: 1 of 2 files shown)


.github/workflows/container-cleanup.yaml
```diff
@@ -0,0 +1,33 @@
...
+        image:
+          - "retina/charts/retina"
+          - "retina/kubectl-retina"
+          - "retina/retina-agent"
+          - "retina/retina-init"
+          - "retina/retina-operator"
+    steps:
+      # This is a fork of the official actions/delete-package-versions which adds GHCR image tag support
+      # https://github.com/actions/delete-package-versions/pull/104
+      - uses: port-of-rotterdam-dtis/delete-package-versions@bf25fb8df311fdcbeac67ba2e5153495d76415a8
```
jimassa, line 27: Is it a good idea to use an unmerged fork for this? The changes in the fork seem trivial, but are there any safety concerns?

rbtr: If it were referenced by version, it would be very unsafe. Ref by the [git SHA is the recommended way to use 3rd party actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions). We could fork it ourselves if we're worried about a malicious SHA1 collision but that seems unlikely.

rbtr: will try and get some traction with the official action to merge changes from the fork
https://github.com/actions/delete-package-versions/issues/88#issuecomment-2024071330

rbtr: this was discussed and accepted until such time as GH acks and updates the official action

timraymond (approved):

Issue #163, opened 2024-03-27
`make helm-install` does not install the latest image 


parkjeongryul:
**Describe the bug**

The latest release is currently v0.0.2, but`make helm-install` with v0.0.2 code will install v0.0.1 image.

**To Reproduce**

```
$ git clone https://github.com/microsoft/retina.git
$ make helm-install
$ k get ds -n kube-system retina-agent -o yaml | grep image:
        image: ghcr.io/microsoft/retina/retina-agent:v0.0.1
      - image: ghcr.io/microsoft/retina/retina-init:v0.0.1
```

**Expected behavior**

Install latest image of current repo.

**Screenshots**

**Platform (please complete the following information):**

I don't think this information is needed for this issue.

**Additional context**

rbtr:
thanks for the report, this is fixed by https://github.com/microsoft/retina/pull/142 which will also update the documentation to recommend the OCI helm install rather than the Make targets

parkjeongryul:
> thanks for the report, this is fixed by https://github.com/microsoft/retina/pull/142 which will also update the documentation to recommend the OCI helm install rather than the Make targets

I see it was a known issue, thanks for sharing.

PR #142, opened 2024-03-26
docs: update instructions for helm install to use GHCR

rbtr:
Updates installation guide in docs and README to reflect that the Helm Chart is now published on GHCR and may be installed directly without cloning the repo or running the Make commands.

Follow up to the fix to #107 (#127) and documents it as requested [here](https://github.com/microsoft/retina/pull/127#issuecomment-2018747660).

fixes #163

PR #164, opened 2024-03-27
fix: `make helm-install` does not install the latest(v0.0.2) image

parkjeongryul:
fixes: #163


rbtr:
potentially duplicate of https://github.com/microsoft/retina/pull/142

PR #168, opened 2024-03-28
fix: remove template variable args from oci install instructions

rbtr:
Removes leftover chart args from the Make instructions that shouldn't be in the quickstart OCI instructions.


peterj (approved):
You were faster than me :)

rbtr:
@peterj you were busy streaming 😉

timraymond (approved):

PR #198, opened 2024-03-29
feat: packet traces feature for Retina CLI

nddq:
# Description

Readding this WIP trace feature for Retina CLI to the repo. Also remove a bunch of Apache License since we are using MIT now.

## Related Issue

Will open an issue to keep track of development.


test/plugin/dns/main_linux.go
```diff
@@ -1,17 +1,3 @@
-// Copyright 2022 The Inspektor Gadget authors
```
rbtr, line 1: this one probably needs to stay

test/utsummary/main.go
```diff
@@ -1,17 +1,3 @@
-// Copyright 2019 The Go Cloud Development Kit Authors
```
rbtr, line 1: this one too

rbtr (commented):
A couple of these are (the original?) 3rd party licenses and should stay. The license of the file supersedes the default license of the repo

docs/CRDs/TracesConfiguration.md
```diff
@@ -0,0 +1,70 @@
+# TraceConfiguration
+
+> **Note:** This feature is currently under experimental development.
```
vakalapa, line 3: add a comment here that this is not implemented in bold headings ?

vakalapa (changes requested):

PR #261, opened 2024-04-12
feat(helm): Improve resource configuration in chart

Flasheh:
# Description

Currently the resources for the retina operator are non-configurable. And the agent only has limits configurable which sets the requests at the same value. With the default cpu limit for the agent being 500m, this seems excessive for request.

I have added what I think are sane defaults, but if you think other values are appropriate let me know.

## Related Issue

N/A

## Checklist

N/A

## Additional Notes

N/A

---

### DIFF
--- a/deploy/manifests/controller/helm/retina/values.yaml
+++ b/deploy/manifests/controller/helm/retina/values.yaml
@@ -122,7 +122,7 @@ resources:
     cpu: "500m"
   requests:
     memory: "300Mi"
-    cpu: "100m"
+    cpu: "500m"
 
 ## @param nodeSelector [object] Node labels for pod assignment
 ## Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/
--- a/deploy/manifests/controller/helm/retina/values.yaml
+++ b/deploy/manifests/controller/helm/retina/values.yaml
@@ -16,6 +16,13 @@ operator:
   capture:
     debug: "true"
     jobNumLimit: 0
+  resources:
+    limits:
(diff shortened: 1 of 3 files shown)


Flasheh:

deploy/manifests/controller/helm/retina/values.yaml
```diff
@@ -113,6 +120,9 @@ resources:
   limits:
     memory: "300Mi"
     cpu: "500m"
+  requests:
+    memory: "300Mi"
+    cpu: "100m"
```
rbtr, line 125: This is a behavior change, as the Pods will now deploy with a QoS class of "Burstable" instead of the [previous class of "Guaranteed"](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-guaranteed) since Requests = Limits when Requests aren't specified.

> [Note: If you specify a limit for a resource, but do not specify any request, and no admission-time mechanism has applied a default request for that resource, then Kubernetes copies the limit you specified and uses it as the requested value for the resource.](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits)

Can we keep them equal but configurable, to maintain the "Guaranteed" QoS by default?
```suggestion
    cpu: "500m"
```

rbtr (changes requested):
Thanks @Flasheh for this enhancement! I only have one point of feedback

rbtr (approved):

rbtr:
LGTM, thanks!

PR #272, opened 2024-04-16
fix: add secondary IPs for enrichment

### LINKED ISSUE #224: Enrich flows based on non-primary Pod IPs
Currently retina only enriches flows based on the primary IP of the Pod, since we only add primary IP to the `ipToEpKey` map in cache, we should support adding all IPs so we can get the EP for secondary IPs as well.

https://github.com/microsoft/retina/blob/5cfb7efcc24b7fa48ff951e9f560c3ebe58f5af3/pkg/controllers/cache/cache.go#L204-L208

FZhg:
# Description

## Related Issue
Fixes #224 

## Checklist

image

## Additional Notes
Both the previous implementation and this PR don't clear stale IP-endpoint bindings immediately from the ipToEpKey Map.
E.g. When endpoint A changes from IP address "1.1.1.1" to "2.2.2.2", the stale binding between the IP address "1.1.1.1" and endpoint A is still in the ipToEpKey map until the IP address "1.1.1.1" gets reassigned to another pod.
However, this shouldn't be a problem. The DNS is supposed to provide up-to-date IP addresses for the sender and receiver so that any flow event will always contain the latest IP addresses.

### DIFF
--- a/pkg/common/endpoint.go
+++ b/pkg/common/endpoint.go
@@ -5,7 +5,13 @@ package common
 import (
 	"fmt"
 	"net"
+
+	"github.com/pkg/errors"
 )
+
+var ErrNoIPFoundForEndpoint = errors.New("no IP found for endpoint")
+
+var ErrNoPrimaryIPFoundEndpoint = errors.New("no primary IP found for endpoint")
 
 func NewRetinaEndpoint(name, namespace string, ips *IPAddresses) *RetinaEndpoint {
 	return &RetinaEndpoint{
@@ -52,7 +58,21 @@ func (ep *RetinaEndpoint) DeepCopy() interface{} {
 	return newEp
 }
 
-func (ep *RetinaEndpoint) IPs() *IPAddresses {
+func (ep *RetinaEndpoint) IPs() ([]string, error) {
+	ep.RLock()
+	defer ep.RUnlock()
+
+	if ep.ips != nil {
+		ips := ep.ips.GetIPs()
+		if len(ips) > 0 {
+			return ips, nil
+		}
+	}
+
+	return []string{}, errors.Wrap(ErrNoIPFoundForEndpoint, ep.Key())
+}
+
+func (ep *RetinaEndpoint) NetIPs() *IPAddresses {
 	ep.RLock()
 	defer ep.RUnlock()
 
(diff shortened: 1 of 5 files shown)


FZhg:
@rbtr Hi Evan, Could you give me another approval for running the CI tests? Thanks!

rbtr:
hey @FZhg it looks like your branch has conflicts, can you rebase it and resolve them? Please update it against latest `main` also, your diff is showing other changes than just yours and I think a rebase would resolve this.

FZhg:
> hey @FZhg it looks like your branch has conflicts, can you resolve them? Please update it against latest `main` also, your diff is showing other changes than just yours and I think a rebase (`git fetch --all && git rebase upstream/main` or equivalent based on your repository config) would resolve this.

Thanks for your help! Problem solved.

pkg/common/endpoint.go
```diff
@@ -1,12 +1,25 @@
 // Copyright (c) Microsoft Corporation.
-// Licensed under the MIT license.
+// Licensed under the MIT licensep.
```
rbtr, line 2:
```suggestion
// Licensed under the MIT license.
```

pkg/common/endpoint.go
```diff
@@ -1,12 +1,25 @@
...
 import (
+	"errors"
 	"fmt"
 	"net"
 )
 
+var ErrNoIPFoundForEndpoint = errors.New("no IP found for endpoint")
+
+func noIPFoundForEndpointError(endpointKey string) error {
+	return fmt.Errorf("%w %s", ErrNoIPFoundForEndpoint, endpointKey)
```
rbtr, line 14: instead of string fmt, let's wrap

```suggestion
	return errors.Wrap(ErrNoIPFoundForEndpoint, endpointKey)
```
applies to all of this pattern in this change. We wrap so that comparisons via `errors.Is` and `errors.As` in the calling code will work to tell us if a child error has one of your typed errors in its heritage. 
With fmt, `errors.Is(err, ErrNoIPFoundForEndpoint)` is always false, because fmt.Errorf is making a new, unique error every time.

pkg/common/endpoint.go
```diff
@@ -129,32 +156,32 @@ func (ep *RetinaEndpoint) SetAnnotations(annotations map[string]string) {
...
-		pip := e.ips.PrimaryIP()
+	if ep.ips != nil {
+		pip := ep.ips.PrimaryIP()
 		if pip != "" {
 			return pip, nil
 		}
 	}
 
-	return "", fmt.Errorf("no primary IP found for endpoint %s", e.Key())
+	return "", noPrimaryIPFoundForEndpointError(ep.Key())
```
rbtr, line 170: incidentally, if we wrap errors, we can do that here instead of needing a helper func

rbtr (commented):
some initial feedback on the typed errors introduced here

pkg/common/endpoint.go
FZhg: Thanks for your feedback~ :smiley: I have applied the `errors.Wrap` method and squashed this PR. Looking forward for any more suggestions 

P.s.
With the `%w` verb, the `fmt.Errorf` wraps the errors as well. But the helper functions are unnecessary.
image

rbtr: Ah you're right about `%w`, I'm sorry! I give this feedback often (`fmt`->`errors`) and didn't process the fmt verbs here 🙈

nddq (approved):
lgtm, thank!

snguyen64 (approved):

snguyen64:
lgmt, thank you for your contribution

PR #275, opened 2024-04-16
fix: capture name flag

rbtr:
# Description

The `-n,--name` flag on `retina capture download` conflicts with the inherited `-n,--namespace` flag from `cli-runtime`. 

This change removes the shorthand `-n` from the `--name` flag.

Additionally, it: 
- changes the `--name` flag to be "Persistent" on the `capture` command instead of only declared on some subcommands
- marks the `--name` flag "Required" and removes unnecessary manual checks that the value is set
- respect the `--name` flag during `create` instead of generating a randstring

## Related Issue

## Checklist

## Additional Notes

---

### DIFF
--- a/cli/cmd/capture/create.go
+++ b/cli/cmd/capture/create.go
@@ -22,7 +22,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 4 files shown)


cli/cmd/capture/create.go
```diff
@@ -174,10 +173,9 @@ func deleteSecret(kubeClient kubernetes.Interface, secretName *string) error {
 }
 
 func createCaptureF(kubeClient kubernetes.Interface) (*retinav1alpha1.Capture, error) {
-	captureName := fmt.Sprintf("retina-capture-%s", utilrand.String(5))
```
huntergregory, line 177: Should we keep auto-generation functionality?

huntergregory (commented):
Should we actually keep the option to auto-generate the name for `capture create`?

Auto-gen might reduce cognitive load? Like if you're doing multiple captures, you don't have to worry about updating the name each time

cli/cmd/capture/create.go
rbtr: You have to use the name for `download` and `delete` so I think the UX is better if it also has to be provided during the `create`

huntergregory: Makes sense. Also, if you're forced to set the name, perhaps it'll be easier to identify which capture is which in a list of captures

huntergregory (approved):
Looks good!

What happens if you run `capture create` twice with the same name? Maybe we could check for or introduce a safeguard in a follow-up PR, so that the `capture create` errors out before overwriting an existing capture.

rbtr:
> What happens if you run `capture create` twice with the same name?

Yeah, currently it allows you to overwrite the job. This is probably acceptable behavior since the command blocks until it gets either an error or the capture is complete (unless `--no-wait` is also set)?

PR #310, opened 2024-04-25
chore: build tags for test files

huntergregory:
# Description

Might be best practice to skip building test files unless their build tag is set. This change doesn't seem to impact _go.mod_

## Checklist


rbtr:
these files are already excluded from the build because they are only referenced in `*_test.go`

Issue #311, opened 2024-04-25
retina-agent pod failed when enabled tcpretrans plugins: invalid memory address or nil pointer dereference


wenhuwang:
**Describe the bug**
Retina-agent can run normally when the tcpretrans plugin is turned off, but fails when the tcpretrans plugin is turned on. The error log is as follows:
```
ts=2024-04-25T07:14:26.112Z level=warn caller=tcpretrans/tcpretrans_linux.go:91 msg="tcpretrans plugin does not have a gadget context" goversion=go1.21.9 os=linux arch=amd64 numcores=48 hostname=node4 podname=retina-agent-ckhc4 version=v0.0.8 apiserver=https://10.68.0.1:443 plugins=packetforward,linuxutil,dns,tcpretrans
ts=2024-04-25T07:14:26.112Z level=info caller=tcpretrans/tcpretrans_linux.go:59 msg="Initialized tcpretrans plugin" goversion=go1.21.9 os=linux arch=amd64 numcores=48 hostname=node4 podname=retina-agent-ckhc4 version=v0.0.8 apiserver=https://10.68.0.1:443 plugins=packetforward,linuxutil,dns,tcpretrans
ts=2024-04-25T07:14:26.112Z level=info caller=pluginmanager/pluginmanager.go:122 msg="Reconciled plugin" goversion=go1.21.9 os=linux arch=amd64 numcores=48 hostname=node4 podname=retina-agent-ckhc4 version=v0.0.8 apiserver=https://10.68.0.1:443 plugins=packetforward,linuxutil,dns,tcpretrans name=tcpretrans
ts=2024-04-25T07:14:26.112Z level=info caller=pluginmanager/pluginmanager.go:173 msg="starting plugin tcpretrans" goversion=go1.21.9 os=linux arch=amd64 numcores=48 hostname=node4 podname=retina-agent-ckhc4 version=v0.0.8 apiserver=https://10.68.0.1:443 plugins=packetforward,linuxutil,dns,tcpretrans
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1a84d40]

goroutine 725 [running]:
github.com/cilium/ebpf.(*MapSpec).Compatible(0x25bcf60?, 0x0)
	/go/pkg/mod/github.com/cilium/ebpf@v0.14.0/map.go:196 +0x40
[...]

PR #322, opened 2024-04-28
fix: the ebpf program in the tcpretrans plugin failed to load

wenhuwang:
# Description
Fix the issues of abnormal status of retina-agent caused by opening tcpretrans plugin. The retina-agent nil pointer is because the [Tracer.socketEnricherMap](https://github.com/inspektor-gadget/inspektor-gadget/blob/main/pkg/gadgets/trace/tcpretrans/tracer/tracer.go#L41) field of the tcpretrans plugin is nil.

## Related Issue

#311 

## Checklist

---


rbtr:
@wenhuwang this doesn't seem to fully resolve the problem. Instead of a nil pointer, now there is an error instantiating the socketEnricher:
> failed to reconcile plugin tcpretrans: failed to new socketEnricher: read BPF iterator: checking if current pid namespace is host pid namespace: host.Init() must be called before calling isHostNamespace()

rbtr (commented):
needs to be tested

rbtr (changes requested):
needs to be tested

wenhuwang:
> @wenhuwang this doesn't seem to fully resolve the problem. Instead of a nil pointer, now there is an error instantiating the socketEnricher:
> 
> > failed to reconcile plugin tcpretrans: failed to new socketEnricher: read BPF iterator: checking if current pid namespace is host pid namespace: host.Init() must be called before calling isHostNamespace()

@rbtr Sorry, I'm on vacation recently. I have tested it locally before. I will verify this problem in a few days.

wenhuwang:
I have not reproduced this problem in my local environment. 
> failed to reconcile plugin tcpretrans: failed to new socketEnricher: read BPF iterator: checking if current pid namespace is host pid namespace: host.Init() must be called before calling isHostNamespace()

Platform (please complete the following information):
OS: Ubuntu 18.04.5 LTS
Kubernetes Version: 1.22.2
Host: self-host
Kernel Version: 5.10.87-051087-generic
Retina Version: v0.0.8

Do you have any specific steps to reproduce the problem? @rbtr

pkg/plugin/tcpretrans/tcpretrans_linux.go
```diff
@@ -56,6 +58,12 @@ func (t *tcpretrans) Init() error {
 	// Create tracer. In this case no parameters are passed.
 	t.tracer = &tracer.Tracer{}
```
anubhabMajumdar, line 59: I believe IG also requires `host.Init()` call before tracer is initialized; similar to https://github.com/microsoft/retina/blob/0dbffcbf9726710702121f71553d0af397d250b2/pkg/plugin/dns/dns_linux.go#L55 . Can you add that as well?

anubhabMajumdar (commented):
Thanks for the change. Can you add details as to how you verified this change?

wenhuwang:
> Thanks for the change. Can you add details as to how you verified this change?
1. Before adding this change, the retina-agent would panic because of the nil pointer.
2. After adding this change, retina-agent runs normally and can get tcpretrans plugin metrics.
3. When troubleshooting this problem through the logs, I can only see the "Initialized tcpretrans plugin" log but not the "failed to new socketEnricher" related logs.

> failed to reconcile plugin tcpretrans: failed to new socketEnricher: read BPF iterator: checking if current pid namespace is host pid namespace: host.Init() must be called before calling isHostNamespace()

pkg/plugin/tcpretrans/tcpretrans_linux.go
wenhuwang: OK, that's what i thought

rbtr (approved):
lgtm, thanks @wenhuwang

Issue #319, opened 2024-04-25
Repeated logs in plugins


nddq:
![image](https://github.com/microsoft/retina/assets/28567936/41e3e4c1-66a1-4f0f-a940-7c3eac955c76)
We are seeing repeated logs for the plugins, might be a good first issue to clean the up. Also need to investigate why dropreason's `Generate` is being called twice implied from the logs

nddq:
ah, we are invoking `Generate` twice for each plugin, once in `Reconcile`
![image](https://github.com/microsoft/retina/assets/28567936/9edb724d-20b8-48ef-a83f-3f5d94979f00)
and another one in `Compile` for `packetparser` and `dropreason`
![image](https://github.com/microsoft/retina/assets/28567936/e1d12742-5569-4374-b942-7d3abe5e7a74)

PR #342, opened 2024-05-03
fix: only invoke Generate once for plugins when reconciling in plugin manager

nddq:
# Description

Fixes issue mentioned in https://github.com/microsoft/retina/issues/319

## Related Issue

https://github.com/microsoft/retina/issues/319

### DIFF
--- a/pkg/plugin/dropreason/dropreason_linux_test.go
+++ b/pkg/plugin/dropreason/dropreason_linux_test.go
@@ -103,7 +103,7 @@ func TestCompile(t *testing.T) {
 
 	err = p.Generate(context.Background())
 	if err != nil {
-		t.Fatalf("Expected no error. Error: %+v", err)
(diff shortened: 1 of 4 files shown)


rbtr (approved):

pkg/plugin/dropreason/dropreason_linux_test.go
```diff
@@ -101,6 +101,11 @@ func TestCompile(t *testing.T) {
 		t.Fatalf("Expected no error. Error: %+v", err)
 	}
 
+	err = p.Generate(context.Background())
+	if err != nil {
+		t.Fatalf("Expected no error. Error: %+v", err)
```
timraymond, line 106: I typically just do this, but nbd either way:

```suggestion
		t.Fatal("Expected no error. Error:", err)
```

rbtr (approved):

PR #354, opened 2024-05-07
docs: Document minimum linux kernel version

timraymond:
After much experimentation, the minimum required Linux kernel version is v5.4. This documents this fact in the intro documentation so that user expectations are set early.


rbtr (dismissed):

rbtr:
ugh markdownlint

snguyen64 (approved):

PR #161, opened 2024-03-27
chore(utils): remove host endianness check

hainenber:
Remove incorrect host endianness check

Closes #139

Issue #95, opened 2024-03-20
Refactor usages of `unsafe`


rbtr:
There are usages of unsafe which could be made safe
https://github.com/microsoft/retina/blob/078d15b3612dc2abacd77b01dc9e6413dd3dbb2a/pkg/utils/utils_linux.go#L41

may be refactored as
```go
func htons(i uint16) uint16 {
  b := make([]byte, 2)
  binary.BigEndian.PutUint16(b, i)
  return binary.BigEndian.Uint16(b)
}
```

Troy-Butler:
Hi, could you please assign this issue to me?

vakalapa:
@Troy-Butler Assigned the issue to you.

PR #443, opened 2024-06-05
fix(#139): do not check for host endianness in IP conversion

### LINKED ISSUE #139: Don't check endieness on the host
**Describe the bug**
Currently Retina checks host endianness to enrich port/IP addresses. That is not correct, network is big-endian. Handle conversion in bpf code.

**To Reproduce**

Ref: https://github.com/microsoft/retina/blob/61e115278c478191910d3038e07e7137a96b65d2/pkg/utils/utils_linux.go#L78

**Expected behavior**
Network is big-endian. Handle conversion as such bpf code.

**Platform (please complete the following information):**

- OS: Linux
- Kubernetes Version: N/A
- Host: Linux/all-arch
- Retina Version: v0.0.1

rectified95:
# Description

Function `determineEndian()` always returns the same value since the host byte order is always little-endian, so it can be skipped.
What's more, the byte order of IP address values passed by eBPF programs doesn't need to be changed, because Golang's `net.IP` type expects big-endian byte arrays.

**Details:**

We always enter this if-branch `binary.LittleEndian.PutUint32(ip, nn)` which looks like it performs conversion TO little-endian, while it actually expects to receive input already IN little-endian.
Note, the input number's bytes are kept in the same order - this function merely converts a uint32 to a byte array. The result is we convert a big-endian number to a big-endian byte array.

```
func (littleEndian) PutUint32(b []byte, v uint32) {
	_ = b[3] // early bounds check to guarantee safety of writes below
	b[0] = byte(v)
	b[1] = byte(v >> 8)
	b[2] = byte(v >> 16)
	b[3] = byte(v >> 24)
}
```
Crucially, `net.IP` expects big-endian byte arrays. For example, IP `15.14.0.0` (`0x0F0E0000` in host order, and `0x00000E0F` in network order) should be passed in as `[0F, 0E, 00, 00]` (most significant byte at the lowest memory address) and so doesn't need inverting. Example here: https://play.golang.com/p/ykiP6mLB_gG

In summary, we currently do not perform byte order conversion and shouldn't start to, either in GO or eBPF code.
## Related Issue

Fixes #139 

This PR looks almost identical to https://github.com/microsoft/retina/pull/161 which was rejected because it didn't include byte order conversion in eBPF, but turns out that's not needed. It also uses the correct uint-to-array conversion functions.

## Checklist


nddq (approved):

nddq:
Relates to #95

anubhabMajumdar (approved):

PR #531, opened 2024-07-04
fix: operator deployment ignores custom namespace

### LINKED ISSUE #492: Operator deployment in different namespace than kube-system.
### Discussed in https://github.com/microsoft/retina/discussions/487

<div type='discussions-op-text'>

<sup>Originally posted by **illarios** June 19, 2024</sup>
As retina is able to be deployed in other namespaces and I can see that the retina-operator-config gets its namespace from the values file:

https://github.com/microsoft/retina/blob/316ad80b4829a464ed60daa50459df4ec27aab92/deploy/manifests/controller/helm/retina/templates/operator.yaml#L258C1-L263C37

I would suggest the same to happen with the operator itself, as the case now is that if you choose another namespace for deployment of retina the operator cannot start without its configuration which apparently exists in the different namespace specified in the values file.

You can see here that the operator's deployment namespace is not configurable:

https://github.com/microsoft/retina/blob/316ad80b4829a464ed60daa50459df4ec27aab92/deploy/manifests/controller/helm/retina/templates/operator.yaml#L6</div>

IMO this is actually a bug - some of the resources in the Helm chart respect the namespace set in the Values, but some have it hardcoded to kube-system, breaking the deployment.

timmhirsens:
# Description

This fixes the operator deployment actually using the namespace defined in the helm values. Without this fix, the deployment will fail, since the configuration ConfigMap for the operator is deployed in the specified namespace while the operator itself is deployed in `kube-system`.

## Related Issue

Fixes #492 

## Checklist

### DIFF
--- a/deploy/hubble/manifests/controller/helm/retina/templates/operator/clusterrolebinding.yaml
+++ b/deploy/hubble/manifests/controller/helm/retina/templates/operator/clusterrolebinding.yaml
@@ -17,6 +17,6 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: retina-operator
-  namespace: kube-system
+  namespace: {{ .Values.namespace }}
 
 {{- end -}}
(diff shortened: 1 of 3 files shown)


timmhirsens:

rbtr (approved):
lgtm, thanks!

PR #549, opened 2024-07-18
feat: Improve controller Dockerfile caching

timraymond:
There's no reason to continually do things like fetching eBPF compilation dependencies, rebuild eBPF, fetch Go dependencies, etc. over and over.  Most of these things will not change frequently, so deserve to be aggressively cached by Docker layers. The things that are more likely to change can then have a much faster build process.

This entailed reducing some of the intermediate image sprawl into a single "bins" image so that cache layers could be reused.

### DIFF
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -21,7 +21,7 @@ FROM --platform=$TARGETPLATFORM mcr.microsoft.com/windows/servercore@sha256:4595
 # build stages
 
 # intermediate go generate stage
-FROM golang AS bins
+FROM golang AS intermediate 
 ARG APP_INSIGHTS_ID # set to enable AI telemetry
 ARG GOARCH=amd64 # default to amd64
 ARG GOOS=linux # default to linux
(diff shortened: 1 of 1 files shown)


controller/Dockerfile
```diff
@@ -21,56 +21,31 @@ FROM --platform=$TARGETPLATFORM mcr.microsoft.com/windows/servercore@sha256:4595
...
-FROM intermediate AS init-bin
-ARG APP_INSIGHTS_ID # set to enable AI telemetry
-ARG GOARCH=amd64 # default to amd64
-ARG GOOS=linux # default to linux
-ARG VERSION
-ENV CGO_ENABLED=0
-ENV GOARCH=${GOARCH}
-ENV GOOS=${GOOS}
 RUN --mount=type=cache,target="/root/.cache/go-build" go build -v -o /go/bin/retina/initretina -ldflags "-X main.version="$VERSION" -X main.applicationInsightsID="$APP_INSIGHTS_ID"" init/retina/main_linux.go
+RUN --mount=type=cache,target="/root/.cache/go-build" go build -v -o /go/bin/retina/captureworkload -ldflags "-X main.version="$VERSION" -X main.applicationInsightsID="$APP_INSIGHTS_ID"" captureworkload/main.go
```
rbtr, line 48: could this be slower since it will serially build all the binaries instead of being parallelizable separate images? if part what you're trying to improve is redundant dep downloads and rebuilds, could we move those upstream to the `golang` image and still have separate parallel final bin builds?

timraymond: @rbtr I did end up switching this. As mentioned in the commit, there wasn't a huge benefit to parallel builds (about 2s). But it wasn't worse, and hopefully the effect is compounded in CI builds.

rbtr (approved):

nddq:
@timraymond can you resolve the merge conflict so we can get this in?

timraymond:
@nddq Thanks for the heads up. I wish the merge queue would ping when there's actionable problems for the author to fix.

rectified95:
@timraymond THis would be very useful to everyone - can you resolve the conflict and merge?

timraymond:
Ack. Will prioritize

nddq (approved):
🚀

rbtr:
github says these commits are from a timposter

timraymond:
@rbtr argh, good catch. New environment and didn't have commit.gpgSign set in the fresh clone. Done.

rbtr (approved):

controller/Dockerfile
```diff
@@ -128,7 +133,7 @@ ENTRYPOINT ["./retina/controller"]
 
 
 # agent final image for windows 
-FROM ${OS_VERSION} AS agent-win
+FROM ${OS_VERSION} as agent-win
```
nddq, line 136:
```suggestion
FROM ${OS_VERSION} AS agent-win
```

timraymond: ARGH… I stared at those lines for like 20 minutes trying to figure out where the conflict was and it was capitalization *in* the line 🤦

timraymond:
RCA of build failures found. Just need an equivalent for windows.

nddq:
@timraymond can you attach the before and after build time?

rectified95:
> Before (luke-warm cache): make retina-image BUILDX_ACTION=--load 3.04s user 1.30s system 2% cpu 2:34.11 total
> 
> Before (hot cache): make retina-image BUILDX_ACTION=--load 3.05s user 1.44s system 3% cpu 2:12.67 total
> 
> After (luke-warm cache): make retina-image BUILDX_ACTION=--load 2.70s user 0.95s system 8% cpu 45.541 total
> 
> After (hot cache): make retina-image BUILDX_ACTION=--load 2.69s user 1.16s system 51% cpu 7.474 total

Building the retina agent image takes 60s for (another 60s for operator) - are you really seeing 2s? :O

timraymond:
@rectified95 That's the output from `time`, so total wall clock is all the way to the right. It's about 45s if you need to fetch Go modules. For example, if I change some log messages around (i.e. just the Go logic), rebuild time is about ~10s.

rbtr (approved):

PR #562, opened 2024-07-24
feat: pktmon client recreation + remove ltsc2019 in azp

matmerr:
# Description
Occasionally the grpc stream experiences transient errors, which are mitigated by recreating the client observer stream. Requires further RCA, but this avoids crashloopbackoff, and gracefully logs to continue

## Related Issue

## Checklist

## Additional Notes

---

### DIFF
--- a/pkg/plugin/windows/pktmon/pktmon_plugin_windows.go
+++ b/pkg/plugin/windows/pktmon/pktmon_plugin_windows.go
@@ -21,6 +21,7 @@ import (
 	"go.uber.org/zap/zapio"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/status"
 )
 
 var (
@@ -130,6 +131,44 @@ func (p *Plugin) Start(ctx context.Context) error {
 	}()
 
 	var str observerv1.Observer_GetFlowsClient
+	str, err := p.GetFlowsClient(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get flow client from observer: %w", err)
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("pktmon context cancelled: %w", ctx.Err())
+		default:
+			err := p.GetFlow(str)
+			if err != nil {
+				p.l.Error("failed to get flow from observer", zap.Error(err))
+				// seeing status errors, recreate client
+				// experiencing sporadic cases of this
+
+				// and then client gets stuck in loop, where crashing and recreating is immediate mitigation.
+				// Instead, we will try to recreate the client and stream
+				if _, ok := status.FromError(err); ok {
+					var clientErr error
+					clientErr = str.CloseSend()
+					if clientErr != nil {
+						return fmt.Errorf("failed to close observer after after failure: %w, parent: %w", clientErr, err)
+					}
+					str, clientErr = p.GetFlowsClient(ctx)
+					if clientErr != nil {
+						return fmt.Errorf("failed to get flow client from observer after failure: %w, parent: %w", clientErr, err)
+					}
+				} else {
+					return fmt.Errorf("failed to get flow from observer: %w", err)
+				}
+			}
+		}
+	}
+}
+
+func (p *Plugin) GetFlowsClient(ctx context.Context) (observerv1.Observer_GetFlowsClient, error) {
+	var str observerv1.Observer_GetFlowsClient
 	fn := func() error {
 		p.l.Info("creating pktmon client")
 		client, err := NewClient()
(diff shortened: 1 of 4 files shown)


pkg/plugin/windows/pktmon/pktmon_plugin_windows.go
```diff
@@ -129,6 +130,44 @@ func (p *Plugin) Start(ctx context.Context) error {
...
+			if err != nil {
+				p.l.Error("failed to get flow from observer", zap.Error(err))
+				// seeing status errors, recreate client
+				// experiencing sporadic cases of this
+
+				// and then client gets stuck in loop, where crashing and recreating is immediate mitigation.
+				// Instead, we will try to recreate the client and stream
+				if _, ok := status.FromError(err); ok {
+					var clientErr error
+					clientErr = str.CloseSend()
```
anubhabMajumdar, line 154: Is this strictly required? If current client instance is malfunctioning and this fails, plugin bails and takes down entire agent, and new agent will basically perform L158 to mitigate.

matmerr: not required, effectively a noop since this a recv stream not send stream, removing

timraymond (commented):
I feel like `StartStreamingRPC` isn't doing enough which is making the `err != nil` case do too much. If we're writing a client to deal with getting flows from gRPC, then the user of that shouldn't have to deal with all of gRPC's baggage. There's also spread-out knowledge between `plugin` and `GRPCClient` about how to deal with flows and present them as something useful to the rest of the application.

I think I'd do something like put the contents of `GetFlow` into `StartStreamingRPC` (and rename it `GetFlow` since that's what it does). Then make that return a `<-chan Event` and an `error` and take a `context`. The deal with the caller is that you get things sent on that channel until you cancel the context (at which point the channel is closed).

The idea would really be to make the client a little more specialized for our use case to make using it easier (and less error-prone).

pkg/plugin/windows/pktmon/pktmon_plugin_windows.go
```diff
@@ -111,103 +116,135 @@ func (p *Plugin) RunPktMonServer() error {
...
+	defer cancel()
+
+	err := p.StartStream(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to setup pktmon stream: %w", err)
 	}
 
-	// Write the event to the external channel.
-	if p.externalChannel != nil {
+	for {
```
huntergregory, line 225: It seems like GetFlow functionality has changed? This infinite loop seems new, and the function now keeps getting flows until one of them is nil or an error is reached?

pkg/plugin/windows/pktmon/pktmon_plugin_windows.go
```diff
@@ -111,103 +116,135 @@ func (p *Plugin) RunPktMonServer() error {
...
+						metrics.LostEventsCounter.WithLabelValues(utils.ExternalChannel, string(Name)).Inc()
+					}
+				}
+			}
+		}
 	}()
 
-	var str observerv1.Observer_GetFlowsClient
+	// run the getflows loop
+	for {
```
huntergregory, line 166: Here’s the second infinite loop for getting flows. Remove one of the infinite loops or rename GetFlow to represent the new logic?

huntergregory (commented):
Just curious about a new infinite loop

pkg/plugin/windows/pktmon/pktmon_plugin_windows.go
matmerr: the idea here is that p.GetFlow contains a child context, which is represents the lifetime of the streaming rpc. when that rpc bails due to an rpc error, that child context is canceled, and returns here to the caller. if it is an rpc error (which happens a decent amount on windows), restart the streaming rpc via calling GetFlows again. if it's not a known rpc (status) error, bail out of the plugin and crash

pkg/plugin/windows/pktmon/pktmon_plugin_windows.go
matmerr: yep exactly, all of the rpc stream management is here, and parent pktmon plugin is no longer responsible for grpc stream management

pkg/plugin/windows/pktmon/pktmon_plugin_windows.go
```diff
@@ -111,103 +116,135 @@ func (p *Plugin) RunPktMonServer() error {
...
+			if fl == nil {
+				p.l.Error("received nil flow, flow proto mismatch from client/server?")
+				return nil
+			}
+
+			ev := &v1.Event{
+				Event:     fl,
+				Timestamp: fl.GetTime(),
+			}
+			eventCh <- ev
```
anubhabMajumdar, line 245: Why can't we send this to `externalChannel` or `enricher` directly?

pkg/plugin/windows/pktmon/pktmon_plugin_windows.go
huntergregory: Makes sense. I feel like adding a couple comments and renaming GetFlow (e.g. make it plural) might help others who read the code later, but I wouldn’t block the PR on it

anubhabMajumdar (approved):

PR #609, opened 2024-08-15
tool(dev): update dev container,  makefile, remove object files

snguyen64:
# Description
Updates to the dev container to use kind and kind registry for local testing.
Removes generated files including mocks
Update mock generates to be consistent
Update linting to 1.60.3 for workflow and local lint and update deprecated linters. 

## Related Issue

## Checklist

## Additional Notes

---


nddq:
hmm pretty sure we needed those empty object files, so the linter doesn't complain 🤔

.devcontainer/installMoreTools.sh
```diff
@@ -6,3 +6,5 @@ sudo apt-get update && sudo apt-get install -y lsb-release wget software-propert
 # Install LLVM 14
 export LLVM_VERSION=14
 curl -sL https://apt.llvm.org/llvm.sh | sudo bash -s "$LLVM_VERSION"
+echo 'export PATH=$PATH:/usr/lib/llvm-14/bin' >> ~/.profile
```
rbtr, line 9: this differs from how we install llvm in the bulid containers, should we make them consistent?

rbtr: should the devcontainer be the same base mariner image as the build container?

hack/tools/tools.go
```diff
@@ -13,7 +13,6 @@ $ go install ...
 package tools
 
 import (
-	_ "github.com/golangci/golangci-lint/cmd/golangci-lint"
```
rbtr, line 16: why?

Makefile
```diff
@@ -524,6 +524,24 @@ quick-build:
 	$(MAKE) retina-image PLATFORM=linux/amd64 BUILDX_ACTION=--push
 	$(MAKE) retina-operator-image PLATFORM=linux/amd64 BUILDX_ACTION=--push
 
+.PHONY: quick-build-devcontainer
+quick-build-devcontainer: # buildx push doesnt work for some reason in kind
```
rbtr, line 528: instead of `*-devcontainer` targets, maybe we could have a `REGISTRY=local` flag or something like that?

Makefile
```diff
@@ -106,8 +106,8 @@ $(GOFUMPT): $(TOOLS_DIR)/go.mod
 
 gofumpt: $(GOFUMPT) ## Build gofumpt
 
-$(GOLANGCI_LINT): $(TOOLS_DIR)/go.mod
-	cd $(TOOLS_DIR); go mod download; go build -tags=tools -o bin/golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
+$(GOLANGCI_LINT): ## Get the golangci-lint binary to match gh workflow version instead of building for consistency
+	cd $(TOOLS_DIR); curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(TOOLS_BIN_DIR) v1.55.2
```
rbtr, line 110: this will have the same problem of drifting away from the workflow, won't it? you've just moved where the dupe hardcoded version is

.github/workflows/golangci-lint.yaml
```diff
@@ -28,6 +28,18 @@ jobs:
         if: env.IS_NOT_MERGE_GROUP
         with:
           go-version-file: go.mod
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: install dependencies and generate
```
rbtr, line 33: great idea

hack/tools/tools.go
rbtr: oh i see the makefile changes now. not convinced that that's better, though

Makefile
snguyen64: I actually ran into a few problems when trying to figure out this linting issue:
The linting results were drastically different from workflow even with the linting config.

So I downgrading hack/tools golangci-lint to 1.55 from 1.60 but then running make lint produced some missing dependency even after go mod tidy. 
Then i just downloaded the binary for 1.55.2 in this command - which actually worked and produced the same results as the workflow

I can try upgrading both, but with 1.60 i noticed that the //nolint wasn't working for our c file references in our go code and i havent figured that out. And we would need to update our linter config since some have been deprecated

.devcontainer/installMoreTools.sh
snguyen64: hm im actually not too familiar with the pros and cons, i was thinking jammy would be more known and easy to develop on, but i dont have much experience with mariner. I can update this if you think it is better, then we can be more consistent

Makefile
snguyen64: i had these separate to not affect existing commands, and some of the steps here are pretty much tied to devcontainer (like the registry:port) 
I can update these though if you think it is better

Makefile
snguyen64: nvm, upgraded version, set cgo=0 and updated linter config to remove deprecated value resolved the mismatch

timraymond: Not sure if it's happening in this instance, but @snguyen64 you should know that golangci-lint is sensitive to the version of Go that it was built with. Weird things have happened in the past where golangci-lint was built with a different version of Go than the module itself.

.devcontainer/installMoreTools.sh
timraymond: I think that the closer the dev / prod impedence match is, the more likely you are to catch compatibility issues. I'd say go with mariner if it's usable in dev.

Makefile
rbtr: i prefer some kind of flag/option rather than duplicate targets if everything else is identical, but i guess that can be less discoverable also. no strong opinion, your call

timraymond: My preference is for dev-related things to have a fast alias and prod-build things to have more flags and knobs. Yes, there's always shell aliases, but it is nice when the standard dev experience doesn't require creating aliases to be productive. It's easy to put whatever flags / tuning are needed in CI, so I'm not as bothered by that (and +1 to the discoverability problem for devex).

.devcontainer/installMoreTools.sh
snguyen64: I did mess around for a little bit with this image, seems that some devcontainer features (docker-in-docker, kind, kubernetes-cli, and github-cli) don't work by default so we would have to set that up 
Im not sure if it's worth it to figure out the configuration for the above missing features 
If you guys think it's worth it then i can play around with it a bit more and figure something out

In this current setup, there is docker-in-docker and kind which works -> if we were to use mariner im not sure if this would work - im thinking it would need to mount to a socket?

I was just envisioning it as installing dependencies, running out make commands that generate retina images, and deploying to kind cluster, so i don't see as much importance in matching this container image with the retina container image

nddq:
@snguyen64 i just remember one thing, we might need those. o files for vendoring as well since they are embedded in the generated ebpf wrapper files. `go mod vendor` will not success IIRC, but you can verify it

PR #283, opened 2024-04-17
feat: conntrack for data aggregation

nddq:
# Description

Implement a _naive_ conntrack map. Related to https://github.com/microsoft/retina/issues/138. Smaller PRs will be open from this PR for ease of review.

- [ ] https://github.com/microsoft/retina/pull/610
- [ ] https://github.com/microsoft/retina/pull/590 (Optional)
- [ ] https://github.com/microsoft/retina/pull/624
- [ ] packets and bytes count per direction per flow (Planned)

## Related Issue

If this pull request is related to any issue, please mention it here. Additionally, make sure that the issue is assigned to you before submitting this pull request.

## Checklist

- [x] I have read the [contributing documentation](https://retina.sh/docs/contributing).
- [x] I signed and signed-off the commits (`git commit -S -s ...`). See [this documentation](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) on signing commits.
- [x] I have correctly attributed the author(s) of the code.
- [x] I have tested the changes locally.
- [x] I have followed the project's style guidelines.
- [ ] I have updated the documentation, if necessary.
- [ ] I have added tests, if applicable.

## Screenshots (if applicable) or Testing Completed

Please add any relevant screenshots or GIFs to showcase the changes made.

## Additional Notes

Add any additional notes or context about the pull request here.

---

Please refer to the [CONTRIBUTING.md](../CONTRIBUTING.md) file for more information on how to contribute to this project.

PR #610, opened 2024-08-15
feat(conntrack): bpf code for conntrack + conntrack GC

nddq:
# Description

Part of https://github.com/microsoft/retina/pull/283, broken down for ease of review. Testing results will be added in https://github.com/microsoft/retina/pull/624
## Related Issue

## Checklist

## Additional Notes

---


pkg/plugin/lib/common/libbpf/_include/asm/barrier.h
nddq, line 1: All files in `_include` were added from https://github.com/libbpf/libbpf for compilation so they are not needed for review

pkg/plugin/conntrack/_cprog/conntrack.h
```diff
@@ -0,0 +1,57 @@
...
+#define CT_REPORT_INTERVAL 30
+// Define the maximum number of connections that can be stored in the conntrack table
+#define CT_MAP_SIZE 262144
+
+enum tcp_flags {
+    TCP_FIN = 0x01,
+    TCP_SYN = 0x02,
+    TCP_RST = 0x04,
+    TCP_PSH = 0x08,
+    TCP_ACK = 0x10,
```
rectified95, line 34: Was there a particular reason for these hex values? If you continue the series as powers of two, you'll be able to store them as a 2-byte bitfield

pkg/plugin/conntrack/_cprog/conntrack.h
```diff
@@ -0,0 +1,57 @@
...
+    TCP_URG = 0x20,
+    TCP_ECE = 0x40,
+    TCP_CWR = 0x80,
+    TCP_NS  = 0x100
+};
+
+enum ct_packet_dir {
```
rectified95, line 41: For all these enums, I suggest coalescing into bit fields.
(optionally for extra compactness could consider stuffing them all in one large type.

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,297 @@
...
+    __u32 dst_ip;
+    __u16 src_port;
+    __u16 dst_port;
+    __u8 proto;
+};
+/**
+ * The structure representing a connection in the connection tracking map.
+ */
+struct ct_entry {
+    __u32 lifetime; // lifetime stores the time when the connection should be removed from the map.
```
rectified95, line 24: is this a unix timestamp or some internal representation? If the former, we should make a project-wide decision whether we care about https://en.wikipedia.org/wiki/Year_2038_problem
and stick to either 32 or 64 bits everywhere

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,297 @@
...
+ * @arg key The key to be used to create the new connection.
+ * @arg flags The flags of the packet.
+ * @arg observation_point The point in the network stack where the packet is observed.
+ * @arg timeout The timeout for the connection.
+ */
+static __always_inline bool _ct_create_new_tcp_connection(struct ct_v4_key key, __u8 flags, enum obs_point observation_point, __u64 timeout) {
+    struct ct_entry new_value;
+    __builtin_memset(&new_value, 0, sizeof(struct ct_entry));
+    __u64 now = bpf_mono_now();
+    new_value.lifetime = now + timeout;
```
rectified95, line 78: you're storing 64bit values in a 32bit field - so `lifetime` does look like a unix timestamp. 
Add an explicit cast here to avoid a compiler warning if you're keeping the smaller field. I'd personally be ok with this - this gives Retina 80 years before overflow lol 
Let me ask the team offline.

pkg/plugin/conntrack/_cprog/conntrack.c
rbtr: don't we need to use 64 bit time since that's what the kernel uses?

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    // The packet is not a SYN packet and the connection corresponding to this packet is not found.
+    // This might be because of an ongoing connection that started before Retina started tracking connections.
+    // Therefore we would have missed the SYN packet. A conntrack entry will be created with best effort.
+    struct ct_entry new_value;
+    __builtin_memset(&new_value, 0, sizeof(struct ct_entry));
+    __u64 now = bpf_mono_now();
+    new_value.lifetime = now + CT_CONNECTION_LIFETIME_TCP;
+    new_value.is_closing = (flags & (TCP_FIN | TCP_RST)) ? true : false;
```
rectified95, line 124: simpler: 
`new_value.is_closing = (flags & (TCP_FIN | TCP_RST)) != 0x0`

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,297 @@
...
+    }
+    
+    // The connection is not found in the forward direction. Check the reply direction.
+    struct ct_v4_key reverse_key;
+    __builtin_memset(&reverse_key, 0, sizeof(struct ct_v4_key));
+    reverse_key.src_ip = key.dst_ip;
+    reverse_key.dst_ip = key.src_ip;
+    reverse_key.src_port = key.dst_port;
+    reverse_key.dst_port = key.src_port;
+    reverse_key.proto = key.proto;
```
rectified95, line 243: can you instead do:
`reverse_key = key`?

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,297 @@
...
+ */
+static __always_inline __attribute__((unused)) bool ct_is_reply_packet(struct ct_v4_key key) {    
+    // Lookup the connection in the map.
+    struct ct_entry *entry = bpf_map_lookup_elem(&retina_conntrack_map, &key);
+    if (entry) {
+        // We return false here because we found the connection in the forward direction
+        // meaning that the packet is coming from the initiator of the connection and therefore not a reply packet.
+        return false;
+    } else {
+        return true;
```
rectified95, line 269: nit: `return entry == NULL`

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,297 @@
...
+        return entry->traffic_direction;
+    }
+    // Construct the reverse key.
+    struct ct_v4_key reverse_key;
+    __builtin_memset(&reverse_key, 0, sizeof(struct ct_v4_key));
+    reverse_key.src_ip = key.dst_ip;
+    reverse_key.dst_ip = key.src_ip;
+    reverse_key.src_port = key.dst_port;
+    reverse_key.dst_port = key.src_port;
+    reverse_key.proto = key.proto;
```
rectified95, line 290: nit: maybe extract to a function, since this is also done elsewhere

pkg/plugin/conntrack/_cprog/conntrack.h
```diff
@@ -0,0 +1,57 @@
...
+// Define how long a TCP connection should be kept in the table
+#define CT_CONNECTION_LIFETIME_TCP 360 
+// Define how long a non-TCP connection should be kept in the table
+#define CT_CONNECTION_LIFETIME_NONTCP 60
+// Define how long a TCP connection should be kept alive after receiving the first SYN
+#define CT_SYN_TIMEOUT 60
+// Define the interval at which a packet should be sent to the userspace
+#define CT_REPORT_INTERVAL 30
+// Define the maximum number of connections that can be stored in the conntrack table
+#define CT_MAP_SIZE 262144
```
rectified95, line 27: how did we derive this value?

pkg/plugin/conntrack/conntrack_linux.go
```diff
@@ -0,0 +1,129 @@
+// Copyright (c) Microsoft Corporation.
```
anubhabMajumdar, line 1: Can you add a `Snapshot` function that prints the current state of `conntrack` map? I want to validate it's updated as expected and can be used to debug retina issues in future.

pkg/plugin/conntrack/_cprog/conntrack.h
nddq: image
I picked this number initially because it aligns with native conntrack table size, but we can always modify this to something more suitable

anubhabMajumdar (commented):
Need some detailed `Testing Done` for this piece of code. I suggest adding a `Snapshot` function in `conntrack_linux.go` to test from userspace.

pkg/plugin/conntrack/conntrack_linux.go
nddq: i have some debug logs that would print out entries in conntrack every GC cycle, are you saying that we should have a CLI that should be able to do that instead?

pkg/plugin/conntrack/_cprog/conntrack.c
nddq: `lifetime` counts from VM boot time. We derive the actual wall clock timestamp in userspace. @rbtr worked on this 🙂

pkg/plugin/conntrack/_cprog/conntrack.c
nddq: `bpf_ktime_get_boot_ns` officially returns u64, so i think I'll go ahead making everything time related u64

pkg/plugin/conntrack/conntrack_linux.go
anubhabMajumdar: Yeah, add that function so that we can add it to a CLI debug command. The snapshot of `conntrack` can be useful to debug bpf issue.

pkg/plugin/conntrack/_cprog/conntrack.c
rectified95: @rbtr Yea, this is basically fine until the year _**2106** if we use unsigned 32-bit_. The kernel and some projects might want to support practically infinite timestamps - it's a judgement call whether we need to. My vote is not really

rectified95: @nddq The name `lifetime` suggests duration - if this is a timestamp I'd rename to something else, eg. `conn_eviction_time` or similar

pkg/plugin/conntrack/_cprog/conntrack.c
rectified95: Let's wait on others' feedback. ConnTrack introduces many timestamp values, and IMO it'd be wasteful to have several new half-empty 64-bit fields passed around.
The fact that the kernel uses 64-bit doesn't necessarily mean our internal representation needs to, as it can be safely cast down to 32-bit.

rectified95: And as above, regardless of what type is used in the end, you should be checking for overflow when using the sum operator

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,297 @@
...
+}
+
+/**
+ * Create a new TCP connection.
+ * @arg key The key to be used to create the new connection.
+ * @arg flags The flags of the packet.
+ * @arg observation_point The point in the network stack where the packet is observed.
+ * @arg timeout The timeout for the connection.
+ */
+static __always_inline bool _ct_create_new_tcp_connection(struct ct_v4_key key, __u8 flags, enum obs_point observation_point, __u64 timeout) {
```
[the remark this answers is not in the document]
nddq: just my personal preference 🙂 not sure which other files you are referring to though, im seeing similar format for function defs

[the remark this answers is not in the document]
rectified95: Totally, just thought I saw it done that way, but maybe have been my previous ebpf project hah

pkg/plugin/conntrack/_cprog/conntrack.h
nddq: not sure what you mean by this, can you add a suggestion?

pkg/plugin/conntrack/_cprog/conntrack.h
rectified95: Oh I misread the values - they're fine and indeed consecutive powers of two.
Still, in line with my other comments and how you do it in `pkg/plugin/conntrack/types_linux.go` these can be converted into constants to avoid wasting space of the underlying 4-byte type.

pkg/plugin/conntrack/_cprog/conntrack.h
rectified95: Convert all these enums with few values into a more compact representation like you do for some other types - these can and should all be bit masks :)

rectified95: `ct_packet_dir` `ct_traffic_dir` and `obs_point` can have all of their values represented with 1+2+2=5 bits, but you're using 3*32=66 bits instead, so over 12x redundancy

rectified95: The first-level optimization is to use a single byte (8 bit) variable for each enum when passing these as parameters - that might be equally readable as separate enums.

The second-level optimization I meant, is that all these 3 enums can be at once represented with 2 bytes: total of 9 values can occupy a single bit each in a 16bit half-word (uint16_t).

Theird-level optimization, is that if that variables of these enums types can never have more than one value at once (unlike a packet having many tcp flags set, direction is only one per packet etc..) then you don't even need a 1 bit per value, but they can be coded as combinations: eg obs_point would only need 2 bits, rather than 4.

Lastly, all these types could be stored in a single vector, just as a CPU uses 64bit word and interprets sub-sections of them as instruction op code, parameters, registers etc. I'd do this with some discernment keeping in mind which types are semantically related tho. Happy to chat offline line if it 's still unclear :)

pkg/plugin/conntrack/conntrack_linux.go
nddq: imagewould something like this useful? planning to pivot from this to adding a debug CLI as well

pkg/plugin/conntrack/_cprog/conntrack.h
nddq: thanks for the explanation! I think I'll go with the first level, which is to make everything a mask, which alleviate the needs for enums as well as still keeping the code readable 🙂

anubhabMajumdar (approved):

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,317 @@
...
+ * @arg flags The flags of the packet.
+ * @arg observation_point The point in the network stack where the packet is observed.
+ * @arg timeout The timeout for the connection.
+ */
+static __always_inline bool _ct_create_new_tcp_connection(struct ct_v4_key key, __u8 flags, __u8 observation_point, __u64 timeout) {
+    struct ct_entry new_value;
+    __builtin_memset(&new_value, 0, sizeof(struct ct_entry));
+    __u64 now = bpf_mono_now();
+    // Check for overflow
+    if (timeout > UINT64_MAX - now) {
```
rectified95, line 94: should be UINT32_MAX since that's the type of `eviction_time`

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,317 @@
...
+ * @arg key The key to be used to create the new connection.
+ * @arg flags The flags of the packet.
+ * @arg observation_point The point in the network stack where the packet is observed.
+ */
+static __always_inline bool _ct_handle_udp_connection(struct ct_v4_key key, __u8 flags, __u8 observation_point) {
+    struct ct_entry new_value;
+    __builtin_memset(&new_value, 0, sizeof(struct ct_entry));
+    __u64 now = bpf_mono_now();
+    // Check for overflow
+    if (CT_CONNECTION_LIFETIME_NONTCP > UINT64_MAX - now) {
```
rectified95, line 115: same as above

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,317 @@
...
+}
+
+/**
+ * Process a packet and update the connection tracking map.
+ * @arg key The key to be used to lookup the connection in the map.
+ * @arg flags The flags of the packet.
+ * @arg observation_point The point in the network stack where the packet is observed.
+ * Returns true if the packet should be report to userspace. False otherwise.
+ */
+static __always_inline __attribute__((unused)) bool ct_process_packet(struct ct_v4_key key, __u8 flags, __u8 observation_point) {
```
rectified95, line 259: `key` can be passed as a pointer like you do in other functions to avoid copying

rectified95 (commented):
Nice job!

pkg/plugin/conntrack/_cprog/conntrack.c
nddq: actually, my initial plan was to pass in the pointer to the packet into conntrack and then construct the key within `ct_process_packet`. That way we don't have to do the subsequent function call to get the traffic_direction and is_reply
image
and instead, we can do it within `ct_process_packet`. However, we will need to relocate the `packet` struct to conntrack since cycle dependency is not allowed, i think. Curious about your opinion

nddq: example in the bigger PR https://github.com/microsoft/retina/pull/283/files#diff-42413aecd6af8d8432f0b375f105d75e593d077574c202a2ab83a99c46a101cd

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,317 @@
...
+/**
+ * Check if a packet is a reply packet to a connection.
+ * @arg key The key to be used to check if the packet is a reply packet.
+ */
+static __always_inline __attribute__((unused)) bool ct_is_reply_packet(struct ct_v4_key key) {    
+    // Lookup the connection in the map.
+    struct ct_entry *entry = bpf_map_lookup_elem(&retina_conntrack_map, &key);
+    // We return false here because we found the connection in the send direction
+    // meaning that the packet is coming from the initiator of the connection and therefore not a reply packet.
+    return entry == NULL;
```
rectified95, line 294: nit: avoid allocating extra stack space for local variable by combining these two lines in the return statement

pkg/plugin/conntrack/_cprog/conntrack.c
rectified95: I didn't understand - why cannot we pass a pointer to they input key here - this function on;y seems to be accessing its values.

Where would we have a cyclical dependency? (Do you mean the eBPF verifier fails on that?)

nddq: IIRC, and i could be wrong, there was some issue with the verifier when passing the key pointer from packetparser to conntrack, i can test it again to verify.
The cyclical dependency is applied to the statement where i wanted to pass the _packet_ pointer from packetparser to conntrack. Since packetparser is already importing conntrack, conntrack can't import packetparser back to get the `packet` struct def. The reason why I think we should be passing the packet pointer to conntrack is because so that we don't have to do subsequent map look up for `is_reply` and `traffic_direction`, but instead set in immediately in `ct_process_packet`

pkg/plugin/conntrack/_cprog/conntrack.c
nddq:
```suggestion
    return bpf_map_lookup_elem(&retina_conntrack_map, &key) == NULL;
```
like this?

rectified95: yup! sorry, I'll learn the 'suggest change' feature soon :D

PR #620, opened 2024-08-19
fix: Implement a dynamic temporary cache that tracks network interfaces that do not support the ethtool operation.

xiaozhiche320:
# Description

1.Implement a dynamic temporary cache that tracks network interfaces that do not support the ethtool operation.
2.Add new test cases for the new changings

## Related Issue

## Checklist

## Additional Notes

---

### DIFF
--- a/pkg/plugin/linuxutil/ethtool_handle_linux.go
+++ b/pkg/plugin/linuxutil/ethtool_handle_linux.go
@@ -1,0 +1,50 @@
+package linuxutil
+
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+
+	lru "github.com/hashicorp/golang-lru/v2"
+
+	"github.com/microsoft/retina/pkg/log"
+	"go.uber.org/zap"
+)
+
+type CachedEthtool struct {
+	EthtoolInterface
+	unsupported *lru.Cache[string, struct{}]
+	l           *log.ZapLogger
+}
+
+func NewCachedEthtool(ethHandle EthtoolInterface, opts *EthtoolOpts) *CachedEthtool {
+	cache, err := lru.New[string, struct{}](int(opts.limit))
+	if err != nil {
+		log.Logger().Error("failed to create LRU cache: ", zap.Error(err))
+	}
+
+	return &CachedEthtool{
+		EthtoolInterface: ethHandle,
+		unsupported:      cache,
+		l:                log.Logger().Named(string("EthtoolReader")),
+	}
+}
+
+var errskip = errors.New("skip interface")
+
+func (ce *CachedEthtool) Stats(intf string) (map[string]uint64, error) {
+	// Skip unsupported interfaces
+	if _, ok := ce.unsupported.Get(intf); ok {
+		return nil, errskip
+	}
+
+	ifaceStats, err := ce.EthtoolInterface.Stats(intf)
+	if err != nil {
+		if strings.Contains(err.Error(), "operation not supported") {
+			ce.unsupported.Add(intf, struct{}{})
+			return nil, errors.Wrap(err, "interface not supported while retrieving stats")
+		}
+		return nil, errors.Wrap(err, "failed to retrieve interface stats")
+	}
+	return ifaceStats, nil
+}
--- a/pkg/plugin/linuxutil/ethtool_stats_linux.go
+++ b/pkg/plugin/linuxutil/ethtool_stats_linux.go
@@ -3,6 +3,7 @@
 package linuxutil
 
 import (
+	"errors"
 	"net"
 	"strings"
 
@@ -28,11 +29,13 @@ func NewEthtoolReader(opts *EthtoolOpts, ethHandle EthtoolInterface) *EthtoolRea
 			return nil
 		}
 	}
+	// Construct a cached ethtool handle
+	CachedEthHandle := NewCachedEthtool(ethHandle, opts)
 	return &EthtoolReader{
 		l:         log.Logger().Named(string("EthtoolReader")),
 		opts:      opts,
 		data:      &EthtoolStats{},
-		ethHandle: ethHandle,
+		ethHandle: CachedEthHandle,
 	}
 }
 
@@ -48,8 +51,6 @@ func (er *EthtoolReader) readAndUpdate() error {
 }
 
 func (er *EthtoolReader) readInterfaceStats() error {
-	// ethtool section
-
 	ifaces, err := net.Interfaces()
 	if err != nil {
 		er.l.Error("Error while getting all interfaces: %v\n", zap.Error(err))
@@ -67,10 +68,15 @@ func (er *EthtoolReader) readInterfaceStats() error {
 		if strings.Contains(i.Name, "lo") || strings.Contains(i.Name, "cbr0") {
 			continue
 		}
+
 		// Retrieve tx from eth0
 		ifaceStats, err := er.ethHandle.Stats(i.Name)
 		if err != nil {
-			er.l.Error("Error while getting ethtool:", zap.String("ifacename", i.Name), zap.Error(err))
+			if errors.Is(err, errskip) {
+				er.l.Debug("Skipping unsupported interface", zap.String("ifacename", i.Name))
+			} else {
+				er.l.Error("Error while getting ethtool:", zap.String("ifacename", i.Name), zap.Error(err))
+			}
 			continue
 		}
 
@@ -79,6 +85,7 @@ func (er *EthtoolReader) readInterfaceStats() error {
 		er.data.stats[i.Name] = tempMap
 
 		er.l.Debug("Processed ethtool Stats ", zap.String("ifacename", i.Name))
+
 	}
 
 	return nil
--- a/pkg/plugin/linuxutil/ethtool_stats_linux_test.go
+++ b/pkg/plugin/linuxutil/ethtool_stats_linux_test.go
@@ -1,7 +1,10 @@
 package linuxutil
 
 import (
+	"errors"
 	"testing"
+
+	lru "github.com/hashicorp/golang-lru/v2"
 
 	"github.com/microsoft/retina/pkg/log"
 	"github.com/microsoft/retina/pkg/metrics"
@@ -15,12 +18,18 @@ var (
 	MockCounterVec *metrics.MockICounterVec
 )
 
+var (
+	errInterfaceNotSupported = errors.New("operation not supported")
+	errOther                 = errors.New("other error")
+)
+
 func TestNewEthtool(t *testing.T) {
 	log.SetupZapLogger(log.GetDefaultLogOpts())
 
 	opts := &EthtoolOpts{
 		errOrDropKeysOnly: false,
 		addZeroVal:        false,
+		limit:             10,
 	}
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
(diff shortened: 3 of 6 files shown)


rbtr:
@xiaozhiche320 please review the contributing guidelines, in particular [signing and signing-off](https://github.com/microsoft/retina/blob/main/CONTRIBUTING.md#cryptographic-signing-of-commits), and also update or remove the boilerplate in the description so that it is accurate.

pkg/plugin/linuxutil/ethtool_stats_linux.go
```diff
@@ -67,10 +73,22 @@ func (er *EthtoolReader) readInterfaceStats() error {
...
 		ifaceStats, err := er.ethHandle.Stats(i.Name)
 		if err != nil {
 			er.l.Error("Error while getting ethtool:", zap.String("ifacename", i.Name), zap.Error(err))
+
+			if len(er.unsupported) < er.limit {
+				er.unsupported[i.Name] = true
+			} else {
+				er.l.Warn("Reached limit of unsupported interfaces")
+			}
 			continue
```
timraymond, line 92: What we're doing here is basically caching this error. I'd rather see a wrapper for `EthtoolInterface` that performs this caching, so that the handling is unchanged (it's still an error).

pkg/plugin/linuxutil/ethtool_stats_linux.go
```diff
@@ -12,11 +12,15 @@ import (
...
 type EthtoolReader struct {
-	l         *log.ZapLogger
-	opts      *EthtoolOpts
-	data      *EthtoolStats
-	ethHandle EthtoolInterface
+	l           *log.ZapLogger
+	opts        *EthtoolOpts
+	data        *EthtoolStats
+	ethHandle   EthtoolInterface
+	unsupported map[string]bool
```
timraymond, line 22: `map[string]struct{}` is the right way to use a map as a set. `map[string]bool` incurs overhead to store all those bools. `map[string]struct{}` says "I want to use the map machinery, but I don't actually care about storing anything." You have to use the two-return form of map access to test for the existence of keys like this:

```go
_, ok := er.unsupported[key]
```

And you assign with:

```go
er.unsupported["something"] = struct{}{}
```

pkg/plugin/linuxutil/ethtool_stats_linux.go
```diff
@@ -12,11 +12,15 @@ import (
...
+	limit       int
```
timraymond, line 23: nit: this should probably be a uint

pkg/plugin/linuxutil/ethtool_stats_linux.go
```diff
@@ -29,10 +33,12 @@ func NewEthtoolReader(opts *EthtoolOpts, ethHandle EthtoolInterface) *EthtoolRea
...
-		l:         log.Logger().Named(string("EthtoolReader")),
-		opts:      opts,
-		data:      &EthtoolStats{},
-		ethHandle: ethHandle,
+		l:           log.Logger().Named(string("EthtoolReader")),
+		opts:        opts,
+		data:        &EthtoolStats{},
+		ethHandle:   ethHandle,
+		unsupported: make(map[string]bool),
+		limit:       limitValue,
```
timraymond, line 41: Can't this be pulled in from `opts`? Why not make it tunable and set defaults at the config level?

pkg/plugin/linuxutil/ethtool_stats_linux_test.go
```diff
@@ -69,32 +74,71 @@ func TestReadInterfaceStats(t *testing.T) {
...
 			wantErr: false,
 		},
+		{
+			name: "test unsported interface",
+			opts: &EthtoolOpts{
+				errOrDropKeysOnly: false,
+				addZeroVal:        false,
+			},
+			statsReturn: nil,
+			statErr:     errInterfaceNotSupported,
```
timraymond, line 84: IMO, just define the error in-place and disable the linter. It makes it harder to read the test otherwise when the error is defined out-of-scope (but definitely follow the linter's advice outside of the test).

pkg/plugin/linuxutil/ethtool_stats_linux_test.go
```diff
@@ -69,32 +74,71 @@ func TestReadInterfaceStats(t *testing.T) {
...
+			ethReader.updateMetrics()
+		}
+
+		if tt.statErr != nil && errors.Is(tt.statErr, errInterfaceNotSupported) {
+			assert.NotEqual(t, nil, ethReader.unsupported, "unsupported map should not be nil")
+			assert.NotEmpty(t, ethReader.unsupported, "unsupported map should contain interfaces")
+		}
+
+		globalUnsupportedCache = ethReader.unsupported
+		fmt.Println("Current unsupported cache: ", ethReader.unsupported, "Global unsupported cache: ", globalUnsupportedCache)
```
timraymond, line 140: Generally, you shouldn't use `fmt.Println` in tests. `t.Log` is better since the output is more orderly.

timraymond (changes requested):
I have a couple thoughts here, see below. Also, what is the motivation for these changes? Is there some performance concern that this improves? Couldn't find a linked issue or explanation in the PR description.

pkg/plugin/linuxutil/ethtool_stats_linux.go
```diff
@@ -67,10 +73,22 @@ func (er *EthtoolReader) readInterfaceStats() error {
 		if strings.Contains(i.Name, "lo") || strings.Contains(i.Name, "cbr0") {
 			continue
 		}
+
+		if er.unsupported[i.Name] {
+			er.l.Info("Skip Unsupported interface", zap.String("ifacename", i.Name))
```
vakalapa, line 78: Lets not write to Info, this will create a lot of logs, make it debug ?

pkg/plugin/linuxutil/ethtool_stats_linux.go
vakalapa: I would recommend using a proper LRU cache here, in case the interface is deleted, we should be able to reuse those limits, or else at some point we will hit the limit and never process any new interfaces.

timraymond: Yeah @vakalapa is right. LRU is the way to go for this. My comment is more for general Go knowledge in this case.

pkg/plugin/linuxutil/types_linux.go
```diff
@@ -102,6 +102,9 @@ type EthtoolOpts struct {
 
 	// when true will include all keys with value 0
 	addZeroVal bool
+
+	// Configurable limit for unsupported interfaces cache
+	limit uint
```
nddq, line 107: we should remove this limit field since we are not using it moving to lru cache

pkg/plugin/linuxutil/ethtool_handle_linux.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		ce.unsupported.Add(intf, struct{}{})
+		return nil, errors.Wrap(err, "error while getting interface stats")
+	}
+	return ifaceStats, nil
+}
+
+func (ce *CachedEthtool) Close() {
+	ce.ethHandle.Close()
+}
```
timraymond, line 49: You can embed EthtoolInterface and eliminate this (and also guarantee that it will continue to implement `EthtoolInterface`.

```go
type CachedEthtool struct {
       EthtoolInterface
       // everything else
}

File: pkg/plugin/linuxutil/ethtool_stats_linux_test.go [hunk 10]
```diff
@@ -39,15 +41,16 @@ func TestNewEthtoolWithNil(t *testing.T) {
...
 	}
 
 	ethReader := NewEthtoolReader(opts, nil)
 	assert.NotNil(t, ethReader)
 }
 
-var globalUnsupportedCache = make(map[string]bool)
-
 func TestReadInterfaceStats(t *testing.T) {
+	globalCache, _ := lru.New[string, struct{}](10)
```
timraymond, line 52: Always worth checking the error, even in tests.

pkg/plugin/linuxutil/ethtool_handle_linux.go
```diff
@@ -0,0 +1,45 @@
...
+var errskip = errors.New("skip interface")
+
+func (ce *CachedEthtool) Stats(intf string) (map[string]uint64, error) {
+	// Skip unsupported interfaces
+	if _, ok := ce.unsupported.Get(intf); ok {
+		return nil, errskip
+	}
+
+	ifaceStats, err := ce.EthtoolInterface.Stats(intf)
+	if err != nil {
```
nddq, line 40: if this error out, does this means that the interface does not support the operation, or could it be because of other reason as well?

pkg/plugin/linuxutil/ethtool_handle_linux.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+type CachedEthtool struct {
+	EthtoolInterface
+	unsupported *lru.Cache[string, struct{}]
+	l           *log.ZapLogger
+}
+
+func NewCachedEthtool(ethHandle EthtoolInterface, opts *EthtoolOpts) *CachedEthtool {
+	cache, err := lru.New[string, struct{}](int(opts.limit))
```
nddq, line 19: make opts.limit an int variable and remove the int casting imo

pkg/plugin/linuxutil/ethtool_stats_linux.go
nddq: @timraymond just curious, why would this being an uint better than int? (besides the fact that the limit can't be a negative number 🙂)

pkg/plugin/linuxutil/ethtool_stats_linux.go
```diff
@@ -28,11 +29,13 @@ func NewEthtoolReader(opts *EthtoolOpts, ethHandle EthtoolInterface) *EthtoolRea
 			return nil
 		}
 	}
+	// Construct a cached ethtool handle
+	CachedEthHandle := NewCachedEthtool(ethHandle, opts)
 	return &EthtoolReader{
 		l:         log.Logger().Named(string("EthtoolReader")),
 		opts:      opts,
 		data:      &EthtoolStats{},
```
matmerr, line 37: can we remove data struct here since all ethtool stats are coming from handle? not blocking, but maybe future consideration

matmerr (approved):

timraymond (approved):

pkg/plugin/linuxutil/ethtool_stats_linux.go
timraymond: @nddq Yeah, that's the only reason. It says, "this would be nonsense if it was negative so we're disallowing that" (because it would be). Robert Griesemer at one point on the Go mailing list had an interesting opinion to default to `int` because its behavior is less surprising when doing arithmetic that would result in negative numbers. In this case though, the field should be immutable, so we don't have those concerns. But it's really a nit.

nddq (approved):

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -318,10 +318,10 @@ func (p *packetParser) clean(tcnl ITc, tcIngressObj *tc.Object, tcEgressObj *tc.
 	// Warning, not error. Clean is best effort.
 	if tcnl != nil {
 		if err := getQdisc(tcnl).Delete(tcEgressObj); err != nil && !errors.Is(err, tc.ErrNoArg) {
```
anubhabMajumdar, line 320: We should check if the error is because the`veth` was deleted. If that's the case, it's not an error.

PR #624, opened 2024-08-20
feat(conntrack): conntrack integration with packetparser

nddq:
# Description

Part 2 of https://github.com/microsoft/retina/pull/610
## Related Issue

## Checklist

Output from debug CLI tool:
image
Hubble flow logs:
image

## Additional Notes

---


pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,295 @@
...
+/**
+    * The structure representing a connection in the connection tracking map.
+ **/
+struct ct_entry {
+    __u32 lifetime; // lifetime stores the time when the connection should be removed from the map.
+    /*
+        * traffic_direction indicates the direction of the connection in relation to the host. 
+        * If the connection is initiated from within the host, the traffic_direction is egress. Otherwise, the traffic_direction is ingress.
+    */
+    enum ct_traffic_dir traffic_direction;
```
rectified95, line 29: Seeing that you're using a single byte for `flags_seen_forward_dir` please do the same for all other enums which only have a handful of values, eg `enum ct_traffic_dir` - this will allow us re-order the fields to avoid introducing unnecessary padding and bloating this `ct_entry` struct.
If possible, do this also for all the other enums in the `packet` struct in packet parser

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -0,0 +1,295 @@
...
+        * flags_seen_*_dir stores the flags seen in the forward and reply direction.
+    */
+    __u8  flags_seen_forward_dir;
+    __u8  flags_seen_reply_dir;
+    /*
+        * last_report_*_dir stores the time when the last packet event was reported in the forward and reply direction respectively.
+    */
+    __u32 last_report_forward_dir;
+    __u32 last_report_reply_dir;
+    __u8 is_closing; // is_closing indicates if the connection is closing.
```
rectified95, line 40: nit: for consistency use bool here?

pkg/plugin/conntrack/_cprog/conntrack.h
```diff
@@ -0,0 +1,57 @@
...
+    TCP_CWR = 0x80,
+    TCP_NS  = 0x100
+};
+
+enum ct_packet_dir {
+    CT_FORWARD = 0,
+    CT_REPLY,
+};
+
+enum ct_traffic_dir {
```
rectified95, line 46: I strongly advise against introducing new enums like these - a single byte is enough.
Sadly, unlike in C++ where you can specify the underlying type, in C we can't really :(
Enum values are great for readability but cause terrible struct bloat and the ebpf programs on in the hot path of our dataplane.

nddq: yea i tried looking up and see if we could define the type for enum in C since it is a whopping 4 bytes by default and came to the same result as well (Copilot lied about this btw 🥲). Will change this to a byte and then add comments for clarification.

pkg/plugin/conntrack/_cprog/conntrack.c
nddq: i have been thinking about this, how much of a performance would we trade off here for readability here do you think? Since I know you have been doing some similar testing with dropreason struct

rectified95: I'm still working on exact numbers, but I don't think we should be trading off any performance on the hot path - especially since we previously saw that _at our current scale limit, the bpf perf event array was dropping events_ - this is exactly the thing that will prevent that.
Kernel code will often be slightly less readable the simple Go code - I'd just add a comment listing all the values of the enum, and also comment mid-line when passing them into functions for readability.

pkg/plugin/conntrack/_cprog/conntrack.h
timraymond: Can we `#define` some bytes instead? Not as pretty and lacking guarantees, but it's at least a little more self-documenting

```c
#define TRAFFIC_DIRECTION_UNKNOWN 0x00
#define TRAFFIC_DIRECTION_INGRESS 0x01
#define TRAFFIC_DIRECTION_EGRESS 0x02

**Comment by nddq** [comment]
fyi can everyone that is reviewing the bpf code do it here instead? https://github.com/microsoft/retina/pull/610 thanks 🙂

File: pkg/plugin/packetparser/_cprog/packetparser.c [hunk 4]
```diff
@@ -5,39 +5,40 @@
...
 	__u16 psh;
 	__u16 urg;
-	__u32 tsval; // TCP timestamp value
-	__u32 tsecr; // TCP timestamp echo reply
 };
 
 struct packet
 {
-	// 5 tuple.
+	__u64 ts; // timestamp in nanoseconds
```
rectified95, line 31: maybe rename to sth like `t_nsec` to disambiguate from the other timestamps that are 32-bit

pkg/plugin/packetparser/_cprog/packetparser.c
```diff
@@ -5,39 +5,40 @@
...
+	__u64 bytes; // packet size in bytes
```
rectified95, line 32: this can be __u32, which comes from the length field of `sk_buff` (and is an overkill anyway - packets cannot be this big, even jumbo frames: https://stackoverflow.com/questions/2613734/maximum-packet-size-for-a-tcp-connection)

rectified95: (edited above with extra context)

pkg/plugin/packetparser/_cprog/packetparser.c
```diff
@@ -221,21 +227,49 @@ static void parse(struct __sk_buff *skb, direction d)
...
+	struct ct_v4_key key;
+	__builtin_memset(&key, 0, sizeof(struct ct_v4_key));
+	key.src_ip = p.src_ip;
+	key.dst_ip = p.dst_ip;
+	key.src_port = p.src_port;
+	key.dst_port = p.dst_port;
+	key.proto = p.proto;
+
+	// Process the packet in ct
+	bool report __attribute__((unused));
```
rectified95, line 248: What does 'unused' mean here why split out the declaration from assignment below?

pkg/plugin/packetparser/_cprog/packetparser.c
```diff
@@ -5,39 +5,40 @@
...
 
 char __license[] SEC("license") = "Dual MIT/GPL";
 
-
-
 struct tcpmetadata {
 	__u32 seq; // TCP sequence number
 	__u32 ack_num; // TCP ack number
+	__u32 tsval; // TCP timestamp value
+	__u32 tsecr; // TCP timestamp echo reply
```
rectified95, line 19: move these fields to the beginning so that you can do a block memory copy when assigning values to this type in `parse()` at `packetparser.c@205`

pkg/plugin/packetparser/_cprog/packetparser.c
```diff
@@ -192,6 +195,9 @@ static void parse(struct __sk_buff *skb, direction d)
 		p.src_port = tcp->source;
 		p.dst_port = tcp->dest;
 
+		// Get all TCP flags.
+		flags = (tcp->fin << 0) | (tcp->syn << 1) | (tcp->rst << 2) | (tcp->psh << 3) | (tcp->ack << 4) | (tcp->urg << 5) | (tcp->ece << 6) | (tcp->cwr << 7);
```
rectified95, line 199: funny we're doing it better than the kernel here lol

rectified95: Seriously though, could out tcp_metadata struct use 1 bit per each of these

pkg/plugin/packetparser/_cprog/packetparser.c
nddq: just to silence the compiler mostly, since we are using this var in an `ifdef` block (which we know for sure that it will be define lol)
image

nddq:
@rectified95 let's open a new PR to address non-conntrack packetparser related improvements that we could make?

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -84,17 +114,16 @@ static __always_inline __u8 _ct_get_traffic_direction(__u8 observation_point) {
...
- * @arg timeout The timeout for the connection.
  */
-static __always_inline bool _ct_create_new_tcp_connection(struct ct_v4_key key, __u8 flags, __u8 observation_point, __u64 timeout) {
+static __always_inline bool _ct_create_new_tcp_connection(struct ct_v4_key key, __u8 flags, __u8 observation_point) {
     struct ct_entry new_value;
     __builtin_memset(&new_value, 0, sizeof(struct ct_entry));
     __u64 now = bpf_mono_now();
     // Check for overflow
-    if (timeout > UINT64_MAX - now) {
+    if (CT_SYN_TIMEOUT > UINT32_MAX - now) {
```
nddq, line 123: @rectified95 this should still be `UINT64_MAX` right? since `now` is a _u64, although we are going to downcast it to _u32 when setting it to `eviction_time` anyway

rectified95: Looking at Cilium's use of `bpf_mono_now` they're inconsistently using 32 and 64-bit variables.
https://github.com/search?q=repo%3Acilium%2Fcilium%20bpf_mono_now&type=code

Is there a specific reason you're using this function? It looks like it returns its value in 'jiffies' which are linux units related to cpu ticks 1-10ms long?

rectified95: Actually, its definitiona depends on whether  `jiffies`is defined, it also uses bpf_ktime_get_sec conditionally
https://github.com/cilium/cilium/blob/510b222f50256c1e876409387c4b93087586256b/bpf/lib/time.h#L14

rectified95: why don't you just use that instead directly? it returns seconds and you can stick to __u32

nddq: i think it only returns value in jiffies if `defined(HAVE_JIFFIES) && defined(ENABLE_JIFFIES) && KERNEL_HZ != 1` right? This is that function definition in Retina now
```
#define NSEC_PER_SEC	(1000ULL * 1000ULL * 1000UL)
#define bpf_ktime_get_sec()	\
	({ __u64 __x = bpf_ktime_get_boot_ns() / NSEC_PER_SEC; __x; })
# define bpf_mono_now()		bpf_ktime_get_sec()
```

nddq: it is in `conntrack.h`

rectified95: To your original question - you should be checking for overflow using the bounds of the final storage type of eviction_tyime, meaning u32

pkg/bpf/setup_linux.go
```diff
@@ -75,5 +76,19 @@ func Setup(l *zap.Logger) error {
 		return errors.Wrap(err, "failed to initialize filter map")
 	}
 	l.Info("Filter map initialized successfully", zap.String("path", plugincommon.MapPath), zap.String("Map name", plugincommon.FilterMapName))
+
+	// Delete existing conntrack map file.
+	err = os.Remove(plugincommon.MapPath + "/" + plugincommon.ConntrackMapName)
```
MikeZappa87, line 81: This attempts to remove the map file regardless of if its present or not then handles the error if it's a specific type or not. The code above does the same, does it make sense to check if the file exists and then delete it? Perhaps a reason exists for this approach?

nddq: we can definitely do an `os.Stat` first, but then we will have to handle the err returned from it, wondering if there is a cleaner way to do this in Go 🤔

rectified95: Can this map not exist in any scenario? If so, then let's just handle two errors separately: failure to remove existing map, and failure to delete non-existent object. Then it doesn't matter whether we check for existence first or not, and can do either way

nddq: these maps will not exist when Retina is installed for the first time

vakalapa (approved):

pkg/plugin/packetparser/_cprog/packetparser.c
rectified95: not a blocker, but you liked and resolved the comment without making any changes :D

pkg/plugin/packetparser/_cprog/packetparser.c
rectified95: resolved without making the change - bad rebase/merge?

nddq: oops 🥲

nddq: oh wait no, i moved this struct to conntrack.c, and it has the right unit

nddq: image

pkg/plugin/packetparser/_cprog/packetparser.c
```diff
@@ -221,21 +191,39 @@ static void parse(struct __sk_buff *skb, direction d)
 
 		p.src_port = udp->source;
 		p.dst_port = udp->dest;
+
+		p.flags = 1;
```
rectified95, line 195: maybe a short note here vs just in PR

pkg/plugin/conntrack/_cprog/conntrack.h
```diff
@@ -13,7 +13,7 @@
 	({ __u64 __x = bpf_ktime_get_boot_ns() / NSEC_PER_SEC; __x; })
 # define bpf_mono_now()		bpf_ktime_get_sec()
 
-#define UINT64_MAX 18446744073709551615ULL
+#define UINT32_MAX 4294967295U
```
rectified95, line 16: see if this works: https://stackoverflow.com/questions/56975861/error-compiling-ebpf-c-code-out-of-kernel-tree

rectified95 (approved):
left some nits, lgtm

pkg/plugin/conntrack/_cprog/conntrack.h
nddq: ok so this is the error if i try to include the header file with `UNIT32_MAX` 
image

pkg/utils/flow_utils.go
```diff
@@ -31,22 +31,15 @@ const (
 // 2 is from host to network and 3 is from network to host.
 // ts is the timestamp in nanoseconds.
 func ToFlow(
+	l *log.ZapLogger,
```
matmerr, line 34: is there any extra reasoning to passing logger here besides warn? this was pretty concise, requires refactoring usage downstream just for a warn

nddq: There is no reason to pass the logger here. From some of the pprof we collected, this actually incurred quite a bit of memory usage.

PR #637, opened 2024-08-23
Optimize packet struct in DropReason plugin.

rectified95:
base - 56bytes
```
uint32_t src_ip;
uint32_t dst_ip;
uint16_t src_port;
uint16_t dst_port;
uint8_t proto;
uint64_t skb_len;
direction_type direction;
struct metrics_map_key key;
uint64_t ts; // timestamp in nanoseconds
bool in_filtermap;
```
48B - reorder `in_filtermap`
```
uint32_t src_ip;
    uint32_t dst_ip;
    uint16_t src_port;
    uint16_t dst_port;
    uint8_t proto;
    bool in_filtermap;
    uint64_t skb_len;
    direction_type direction;
    struct metrics_map_key key;
    uint64_t ts; // timestamp in nanoseconds
```
40B - remove `direction` + its padding
```
uint32_t src_ip;
uint32_t dst_ip;
uint16_t src_port;
uint16_t dst_port;
uint8_t proto;
bool in_filtermap;
uint64_t skb_len;
struct metrics_map_key key;
uint64_t ts; // timestamp in nanoseconds
```
32B - shrink `skb_len` to match `sk_buff`; 
         shrink `drop_type` to uint16_t from uint32_t
         split up `metrics_map_ke`y within `struct packet` to remove nested struct padding
```
__u32 src_ip;
__u32 dst_ip;
__u16 src_port;
__u16 dst_port;
__u32 skb_len;
__u32 return_val;
__u16 drop_type;
__u8 proto;
bool in_filtermap;
__u64 ts; // timestamp in nanoseconds
```


pkg/plugin/dropreason/kprobe_bpfel_x86.o
nddq, line 1: can you make these .o files empty?

pkg/plugin/dropreason/dropreason_linux_test.go
```diff
@@ -261,8 +261,8 @@ func TestDropReasonReadDataPodLevelEnabled(t *testing.T) {
 	mockedPerfReader := mocks.NewMockIPerfReader(ctrl)
 	menricher := enricher.NewMockEnricherInterface(ctrl) //nolint:typecheck
 
-	// create a rawSample slice and fill it with 56 bytes of data. 56 is the size of the rawSample in perf.Record (kprobePacket)
-	rawSample := make([]byte, 56)
+	// create a rawSample slice and fill it with 32 bytes of data. 32 is the size of the rawSample in perf.Record (kprobePacket)
+	rawSample := make([]byte, 32)
```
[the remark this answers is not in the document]
rectified95: oh yeah for sure, didn't even think of it when updating:
https://dlintw.github.io/gobyexample/public/memory-and-sizeof.html

pkg/plugin/dropreason/kprobe_bpfel_x86.o
rectified95: hmm, i didn't notice it was empty before - why are we checing them in? typically, the .o is checked in to do away with compiler differences for freshly checkout repos

nddq: these two reasons basically
- https://github.com/microsoft/retina/pull/269
- https://github.com/microsoft/retina/pull/334

timraymond: It's too easy to miss that these files are getting touched. We *should* remove them. It is *dangerous* to have unreviewable code even if it's *supposed* to get overwritten by a build step (what happens if it doesn't one time?)

rectified95: I removed them from this PR. @nddq / @timraymond why are these needed by `go mod vendor`?
Since they need to be empty, perhaps we can just `touch` them right before vendoring?

nddq: these are embedded in the go wrapper files generated from ebpf code. `go mod vendor` will not complete sucessfully if these are missing

nddq:
@rectified95 can you run `go generate` for both amd64 and arm64 to re generated the wrapper files?

Issue #284, opened 2024-04-18
prevent the potential for misconfiguration of MetricsInterval


huntergregory:
In the agent's ConfigMap, `MetricsInterval` is meant to be a number in seconds:
https://github.com/microsoft/retina/blob/30a128b985bc99fc8686ef21afa1cc7358dc7dfd/pkg/config/config.go#L54

But the type for `MetricsInterval` is `time.Duration`:
https://github.com/microsoft/retina/blob/30a128b985bc99fc8686ef21afa1cc7358dc7dfd/pkg/config/config.go#L21
so someone could configure `MetricsInterval: 5s`, and unknowingly end up with an actual interval of 150 years 😬

rbtr:
viper supports duration as config, so we should always expect this value to come in as a valid duration and not turn it in to one later

huntergregory:
Should we deprecate this field and replace it with say `MetricsIntervalDuration`? If we reuse the same field, if someone continues to use the current config of `MetricsInterval: 10`, they would then get a 10 nanosecond interval

ritwikranjan:
To address the potential misconfiguration of `MetricsInterval`, I agree with @huntergregory about introducing a new field in the YAML configuration, such as `MetricsIntervalDuration`. This new field would take precedence over the existing `MetricsInterval`, but we would include a conditional statement to maintain backward compatibility. This way, if `MetricsIntervalDuration` is present, it will be used; otherwise, we will fall back to `MetricsInterval`.

Here's a rough outline of the plan:

- Introduce the new `MetricsIntervalDuration` field in the YAML configuration.
- Implement a conditional check to prioritize `MetricsIntervalDuration` if it is set.
- Ensure that if `MetricsIntervalDuration` is not set, the existing `MetricsInterval` field will be used. We can also add a log line here to highlight that it is deprecated.

Please let me know your thoughts on this approach!

rbtr:
@ritwikranjan plan looks good

PR #638, opened 2024-08-23
fix: prevent potential misconfiguration of metricsInterval (#284)

ritwikranjan:
# Description

I have added a new field `metricsIntervalDuration` of type `time.Duration`. It will super cede the `metricsInterval` configuration of the same type. The older key is not removed to keep backward compatibility. The change is done to prevent potential misconfiguration that can happen by assigning a duration to `metricsInterval` when it is converted to seconds.

## Related Issue

Issue #284 

## Checklist

- [X] I have read the [contributing documentation](https://retina.sh/docs/contributing).

The logs showing that retina agent runs successffully after making the changes.
retina-after-chnage

This shows the log line that `metricsInterval` should not be used.
retina-using-metrics-interval

## Additional Notes

I was not able to us the `zap` logger as it is probably not initialized when the `GetConfig` is called or I might be missing how to use it so i used the go `log` library. Do let me know if we need to change that.

---


deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -86,8 +86,8 @@ enabledPlugin_win: '["hnsstats"]'
 
 enableTelemetry: true
 
-# Interval, in seconds, to scrape/publish metrics.
-metricsInterval: 10
+# Interval, in duration, to scrape/publish metrics.
+metricsIntervalDuration: "10s"
```
timraymond, line 90: ❤️ thanks for making this a duration

pkg/config/config.go
```diff
@@ -95,8 +97,13 @@ func GetConfig(cfgFilename string) (*Config, error) {
 	if err != nil {
 		return nil, fmt.Errorf("fatal error config file: %s", err)
 	}
-	// Convert to second.
-	config.MetricsInterval = config.MetricsInterval * time.Second
+
+	if config.MetricsIntervalDuration != 0 {
+		config.MetricsInterval = config.MetricsIntervalDuration
+	} else {
```
timraymond, line 103: I guess this is bug-for-bug compatible, but setting neither field would generate the deprecation warning unless I missed a presence check somewhere up the call stack.

pkg/config/config.go
```diff
@@ -55,6 +56,7 @@ type Config struct {
 	LogLevel                 string        `yaml:"logLevel"`
 	EnabledPlugin            []string      `yaml:"enabledPlugin"`
 	MetricsInterval          time.Duration `yaml:"metricsInterval"`
```
rbtr, line 58: go ahead and un-export this field and throw a `//deprecated` on it so it doesn't get used anymore

pkg/config/config.go
```diff
@@ -55,6 +56,7 @@ type Config struct {
...
+	MetricsIntervalDuration  time.Duration `yaml:"metricsIntervalDuration"`
```
rbtr, line 59: or this one, whichever one shouldn't be referenced outside of the config

pkg/config/config.go
ritwikranjan: No, I missed it completely, Will fix it thanks!

pkg/config/config.go
```diff
@@ -95,8 +97,13 @@ func GetConfig(cfgFilename string) (*Config, error) {
 	if err != nil {
 		return nil, fmt.Errorf("fatal error config file: %s", err)
 	}
-	// Convert to second.
-	config.MetricsInterval = config.MetricsInterval * time.Second
+
+	if config.MetricsIntervalDuration != 0 {
+		config.MetricsInterval = config.MetricsIntervalDuration
+	} else {
+		config.MetricsInterval *= time.Second
```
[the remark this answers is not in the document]
ritwikranjan: Sorry, new to the language, do you have any suggestion to improve it in someway?

pkg/config/config.go
ritwikranjan: Inside the go code `MetricsIntervalDuration` will not be referenced however in the yaml `metricsInterval` is deprecated, we should in future make the field like:
```
MetricsInterval    time.Duration    `yaml:"metricsIntervalDuration"`
```

pkg/config/config.go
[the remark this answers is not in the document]
rbtr: No, this is correct and [allowed](https://go.dev/ref/spec#Operators_and_punctuation). My comment is more that Go rarely has these sort of abbreviated syntax and usually prefers just writing the whole thing out.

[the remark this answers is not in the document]
timraymond: @rbtr I had the same thought as well when I read this... I had to stop and think whether arithmetic assignment operators were legal.

timraymond (dismissed):

rbtr (approved):

PR #698, opened 2024-09-03
feat: Add the ability to overlay configuration

timraymond:
In certain deployments, it's desirable to have one set of configuration pre-set and another set of overrides to that configuration. In this overlay configuration, only certain fields should be overridable--additional provided fields should be ignored.

This modifies `GetConfig` in a backwards-compatible way to accept a variadic list of configuration overlays. These overlays take a filesystem path and a set of allowed fields that are permitted from those overlay configs. It also introduces `internal` types for performing filtering of these configuration files.

These are currently not used. A subsequent pull request will integrate this work.


rbtr:
I'm hearing... drop-ins? 🙂

anubhabMajumdar (approved):

anubhabMajumdar (approved):

PR #720, opened 2024-09-10
test: Retina e2e scale test

alexcastilio:
# Description

Create Retina E2E Scale tests using test/e2e framework.

## Related Issue

## Checklist

## Additional Notes

---

### DIFF
--- a/test/e2e/framework/kubernetes/create-namespace.go
+++ b/test/e2e/framework/kubernetes/create-namespace.go
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"context"
 	"fmt"
+	"time"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -11,7 +12,36 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
-func CreateNamespace(kubeconfigpath, namespace string) error {
+type CreateNamespace struct {
+	Namespace          string
+	KubeConfigFilePath string
+}
+
+func (c *CreateNamespace) Run() error {
+	return CreateNamespaceFn(c.KubeConfigFilePath, c.Namespace)
+}
+
+func (c *CreateNamespace) Stop() error {
+	return nil
+}
+
+func (c *CreateNamespace) Prevalidate() error {
+	return nil
+}
+
+func (c *CreateNamespace) getNamespace() *v1.Namespace {
+	return &v1.Namespace{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Namespace",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: c.Namespace,
+		},
+	}
+}
+
+func CreateNamespaceFn(kubeconfigpath, namespace string) error {
 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigpath)
 	if err != nil {
 		return fmt.Errorf("error building kubeconfig: %w", err)
@@ -22,7 +52,7 @@ func CreateNamespace(kubeconfigpath, namespace string) error {
 		return fmt.Errorf("error creating Kubernetes client: %w", err)
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeoutSeconds*time.Second)
 	defer cancel()
 
 	_, err = clientset.CoreV1().Namespaces().Create(ctx, &v1.Namespace{
@@ -30,11 +60,11 @@ func CreateNamespace(kubeconfigpath, namespace string) error {
 			Name: namespace,
 		},
 	}, metav1.CreateOptions{})
-
 	if err != nil && !errors.IsAlreadyExists(err) {
-		return fmt.Errorf("error creating namespace: %w", err)
-	} else {
-		fmt.Printf("Namespace '%s' created successfully.\n", namespace)
-		return nil
+		return fmt.Errorf("failed to create namespace \"%s\": %w", namespace, err)
 	}
+
+	fmt.Printf("Namespace \"%s\" created.\n", namespace)
+
+	return nil
 }
--- a/test/e2e/framework/scaletest/create-network-policies.go
+++ b/test/e2e/framework/scaletest/create-network-policies.go
@@ -1,0 +1,158 @@
+package scaletest
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	e2ekubernetes "github.com/microsoft/retina/test/e2e/framework/kubernetes"
+
+	"github.com/microsoft/retina/test/e2e/framework/scaletest/templates"
+	netv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type CreateNetworkPolicies struct {
+	KubeConfigFilePath          string
+	Namespace                   string
+	NumNetworkPolicies          int
+	NumUnappliedNetworkPolicies int
+	NumSharedLabelsPerPod       int
+}
+
+// Useful when wanting to do parameter checking, for example
+// if a parameter length is known to be required less than 80 characters,
+// do this here so we don't find out later on when we run the step
+// when possible, try to avoid making external calls, this should be fast and simple
+func (c *CreateNetworkPolicies) Prevalidate() error {
+	return nil
+}
+
+// Primary step where test logic is executed
+// Returning an error will cause the test to fail
+func (c *CreateNetworkPolicies) Run() error {
+
+	config, err := clientcmd.BuildConfigFromFlags("", c.KubeConfigFilePath)
+	if err != nil {
+		return fmt.Errorf("error building kubeconfig: %w", err)
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return fmt.Errorf("error creating Kubernetes client: %w", err)
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeoutSeconds*time.Second)
+	defer cancel()
+
+	networkPolicies := c.generateNetworkPolicies(c.NumNetworkPolicies)
+
+	for _, np := range networkPolicies {
+		e2ekubernetes.CreateResource(ctx, np, clientset)
+	}
+
+	return nil
+}
+
+// Require for background steps
+func (c *CreateNetworkPolicies) Stop() error {
+	return nil
+}
+
+func (c *CreateNetworkPolicies) generateNetworkPolicies(numPolicies int) []runtime.Object {
+	objs := []runtime.Object{}
+	for i := 0; i < numPolicies; i++ {
+		name := fmt.Sprintf("policy-%05d", i)
+
+		template := templates.NetworkPolicy.DeepCopy()
+
+		template.Name = name
+		template.Namespace = c.Namespace
+
+		valNum := i
+		if valNum >= c.NumSharedLabelsPerPod-2 {
+			valNum = c.NumSharedLabelsPerPod - 2
+		}
+
+		template.Spec.PodSelector.MatchLabels[fmt.Sprintf("shared-lab-%05d", valNum)] = "val"
+
+		ingressNum := valNum + 1
+		template.Spec.Ingress = []netv1.NetworkPolicyIngressRule{
+			{
+				From: []netv1.NetworkPolicyPeer{
+					{
+						PodSelector: &metav1.LabelSelector{
+							MatchLabels: map[string]string{
+								fmt.Sprintf("shared-lab-%05d", ingressNum): "val",
+							},
+						},
+					},
+				},
+			},
+		}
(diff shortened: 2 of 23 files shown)


test/e2e/framework/types/runner.go
```diff
@@ -2,27 +2,61 @@ package types
...
 
-func (r *Runner) Run() {
+func (r *runner) Run() {
 	if r.t.Failed() {
 		return
 	}
 	require.NoError(r.t, r.Job.Run())
 }
+
+func RunnerWithTimeout(r Runner, t time.Duration) runnerFunc {
```
timraymond, line 41: This is probably better done with `context.Context`

test/e2e/framework/types/runner.go
```diff
@@ -2,27 +2,61 @@ package types
...
+	}()
+
+	return func() {
+		go func() {
+			r.Run()
+			runComplete <- struct{}{}
+		}()
+		select {
+		case <-timeout:
+			panic("Timeout")
```
timraymond, line 57: It's better to `testing.(*T).Fatal` in tests than outright panic. The effect is similar (it calls `runtime.Goexit()` under the hood which is panic-adjacent).

test/e2e/scale_test.go
```diff
@@ -0,0 +1,65 @@
...
+	createTestInfra := types.NewRunner(t, jobs.CreateTestInfra(subID, clusterName, location, kubeConfigFilePath))
+	createTestInfra.Run()
+
+	// Hacky way to ensure that the test infra is deleted even if the test panics
+	defer func() {
+		if r := recover(); r != nil {
+			t.Logf("Recovered in TestE2ERetina_Scale, %v", r)
+			_ = jobs.DeleteTestInfra(subID, clusterName, location).Run()
+		}
+	}()
```
timraymond, line 56: Aforementioned panic-avoidance strategies would prevent you from having to do these strange gyrations. Remember that panics are not a control-flow mechanism.

You can replace this with `testing.(*T).Cleanup`

test/e2e/scale_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// Install Retina
+	installRetina := types.NewRunner(t, jobs.InstallRetina(kubeConfigFilePath, chartPath))
+	types.RunnerWithTimeout(installRetina, 5*time.Minute).Run()
```
timraymond, line 60: You should pull out the timeout from `testing.(*T).Deadline` and create a `context.WithDeadline` using that. That puts control back in the user's hands for how long they're willing to wait. They would ultimately control this via the `go test -timeout` flag. It would also respect `<C-c>`

timraymond (commented):
I realize that this is a draft, and I ordinarily don't review drafts since things are subject to change... but I figured these would be good suggestions early-on if you're planning to write some more here.

test/e2e/scale_test.go
alexcastilio: Thank you very much for this comment. I was looking for a way to deal with `<C-c>` and `go test -timeout` but had not yet found a good solution. I'll do that.

test/e2e/jobs/jobs.go
```diff
@@ -62,6 +62,20 @@ func DeleteTestInfra(subID, clusterName, location string) *types.Job {
 	return job
 }
 
+func InstallRetina(kubeConfigFilePath, chartPath string) *types.Job {
```
anubhabMajumdar, line 65: Why do you need this? Can you refactor `InstallAndTestRetinaBasicMetrics`.

alexcastilio: I preferred not to change `InstallAndTestRetinaBasicMetrics` since this might be in use by someone else and the scope seems to be different. 

The plan for the new pipeline is to Create Infra, Install Retina, Scale Up and Run Tests, so my thinking is to create different jobs for each stage so each job could be reused somewhere else. So in the future instead of using `InstallAndTestRetinaBasicMetrics`, one could [re-]use `InstallRetina`, `Test Basic Metrics`, and perhaps add some other set of tests. Does that make sense?

test/e2e/framework/types/runner.go
```diff
@@ -20,9 +21,20 @@ func NewRunner(t *testing.T, job *Job) *Runner {
...
-	require.NoError(r.t, r.Job.Run())
+	runComplete := make(chan struct{})
+
+	go func() {
+		require.NoError(r.t, r.Job.Run())
+		runComplete <- struct{}{}
+	}()
+	select {
+	case <-ctx.Done():
+		r.t.Fatal("Timeout")
```
matmerr, line 36: add that this is a go test runner timeout, could be ambiguous with sporadic infra timeouts that could happen

timraymond: +1. The message you leave in `t.Fatal` is the message you will read in a cryptic jumble of log output. Be kind to yourself and always make it as descriptive as you can.

test/e2e/framework/scaletest/validate-options.go
```diff
@@ -0,0 +1,99 @@
...
+		return errors.New("NumSharedLabelsPerPod must be at least 3 when NumNetworkPolicies > 0 because of the way Network Policies are generated")
+	}
+
+	return nil
+}
+
+// Returning an error will cause the test to fail
+func (po *ValidateAndPrintOptions) Run() error {
+
+	log.Printf(`Starting to scale with folowing options:
```
matmerr, line 41: Use `%+v` formatting directive to print struct keys and values

```suggestion
	log.Printf(`Starting to scale with following options %+v", po.Options)
```

test/e2e/framework/scaletest/validate-nodes.go
```diff
@@ -0,0 +1,65 @@
...
+	defaultTimeoutSeconds = 300
+)
+
+type ValidateNumOfNodes struct {
+	NumNodesRequired   int
+	Label              map[string]string
+	KubeConfigFilePath string
+}
+
+// Useful when wanting to do parameter checking, for example
```
matmerr, line 25: this comment is a bit redundant when copy pasted to every struct :)

test/e2e/framework/scaletest/templates/networkpolicy.go
```diff
@@ -0,0 +1,24 @@
+package templates
+
+import (
+	netv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	NetworkPolicy = netv1.NetworkPolicy{
```
matmerr, line 9: +1 this doesn't necessarily seem scale test specific, let's put it in a more common directory and/or reuse with existing create deny netpol steps

test/e2e/framework/scaletest/templates/kapinger-rbac.go
```diff
@@ -0,0 +1,45 @@
+package templates
+
+import (
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	KapingerClusterRole = rbacv1.ClusterRole{
```
matmerr, line 9: +same for this, let's stick to one way of deploying kapinger and tweak that if we need to

test/e2e/framework/kubernetes/create-namespace.go
```diff
@@ -0,0 +1,62 @@
...
+	v1 "k8s.io/api/core/v1"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type CreateNamespace struct {
+	Namespace          string
+	KubeConfigFilePath string
+	DryRun             bool
```
matmerr, line 17: Do we have any strong use cases for dry run? In a lot of these cases so far all core step code is gated by this conditional, and when dry run is set to true, it just pulls kubeconfig

For what it's worth the prevalidate step is effectively a dry run

matmerr (dismissed):

test/e2e/framework/scaletest/create-resources.go
```diff
@@ -0,0 +1,163 @@
...
+	e2ekubernetes "github.com/microsoft/retina/test/e2e/framework/kubernetes"
+	"github.com/microsoft/retina/test/e2e/framework/scaletest/templates"
+	yaml "gopkg.in/yaml.v2"
+	v1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type CreateResources struct {
```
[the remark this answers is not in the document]
alexcastilio: It's being reused inside Run() method (line 66). This step puts together the generation of all objects required for this scale test and then it reuses the CreateResource function you mention to apply that to the cluster.

What would you suggest to reuse it in a better way?

test/e2e/framework/scaletest/create-network-policies.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"github.com/microsoft/retina/test/e2e/framework/scaletest/templates"
+	netv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type CreateNetworkPolicies struct {
```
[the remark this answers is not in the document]
alexcastilio: Yes, that could be done with some refactoring.

kubernetes.CreateDenyAllNetworkPolicy step creates a specific NetworkPolicy (NP) that is encapsulated in a way where the only Namespace and one podSelector label could be provided.

scaletest.CreateNetworkPolicies step creates a number of NPs that are also specific with some podSelector labels needed for the scale-tests.

The first one would need to have more fields to allow more customization on generated NP. But it would also require to check and change the code everywhere it's currently being used. 

If it were done that way, how should this step that generates a single NP be reused to generate many NPs? Inside the ScaleTest function that generates the job (file test/e2e/jobs/scale.go) add a loop to add this same Step several times? Or a new step should be created encapsulating these multiple NP creations inside of it? (I'm not sure if the framework currently allows creation/reuse of steps inside of steps.)

Anyway, if this is the best approach, maybe this could be done in a separate refactoring task. Do you agree?

test/e2e/framework/scaletest/validate-nodes.go
alexcastilio: This comment comes from the interface. I use a plugin to automatically create the methods from the interface that I want to implement, so it also gets the comment from it :)

test/e2e/framework/scaletest/validate-options.go
alexcastilio: ok

test/e2e/framework/scaletest/create-resources.go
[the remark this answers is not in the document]
matmerr: My thought stems from the fact that creating/destroying Kubernetes resources is a core component of these suites, and if there was a way of making this CreateResources step able to be shared amongst other suites.

In other such cases we've seen people create resources that require pod/job/daemonset to be ready, but the creating resources aspect of the test just creates the resource in apiserver and moves on without waiting for job or pod, which then results in faulty tests. Then it is necessary to add a wait for pod ready like the added step in this pr and/or verify stable cluster, etc.

I don't have a clear solution for this at the moment so I won't block, but it seems everybody has different way of creating test resources, so just starting a dialog around the topic

test/e2e/framework/kubernetes/create-namespace.go
matmerr: cc: @huntergregory for his thoughts

test/e2e/framework/scaletest/create-network-policies.go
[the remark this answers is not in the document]
matmerr: I see the reasoning, while there might be a way to accommodate both scenarios, that may be out of scope for this pr given current limitations. Existing cases are verifying granular drop behavior unlike this suite, can shelve for later

test/e2e/framework/kubernetes/create-namespace.go
huntergregory: I think we can remove this for simplicity

huntergregory: This was primarily used to calculate how many IP sets et cetera that npm will create

test/e2e/framework/scaletest/templates/networkpolicy.go
alexcastilio: I can put it in the folder `test/e2e/framework/kubernetes/templates` and existing create deny netpol step can be changed in another task to reuse it and perform changes needed. Is that ok?

test/e2e/framework/scaletest/templates/kapinger-rbac.go
alexcastilio: Done

test/e2e/framework/kubernetes/create-namespace.go
alexcastilio: DryRun removed

ibezrukavyi (dismissed):

ibezrukavyi (approved):

ibezrukavyi (approved):

ibezrukavyi (dismissed):

ibezrukavyi (approved):

ibezrukavyi (approved):

Issue #766, opened 2024-09-20
[Docs] Capture installation docs are not right or have wrong links


vakalapa:
https://retina.sh/docs/Captures/

Here the link is not showing up properly and documentation for captures needs to vetted fully again. 

<img width="430" alt="image" src="https://github.com/user-attachments/assets/38a6fd1d-bb28-4e3f-a66d-7bb25f33cf18">

Links for installation are also incorrect, needs to focus on krew install commands.

PR #832, opened 2024-10-08
chore: Streamline Usability of Captures

kamilprz:
# Description

This PR aims to improve the usability of the `retina capture create` command.

New defaults have been set, eliminating the need for required flags. The command can now be ran with just `kubectl retina capture create`.

The documentation received another overhaul - restructuring how flags are displayed into a more readable table and curating the examples which are shown.

## Related Issue

Issue related to updating usability of captures
https://github.com/microsoft/retina/issues/766

## Checklist

## Additional Notes

---

### DIFF
--- a/cli/cmd/capture/create.go
+++ b/cli/cmd/capture/create.go
@@ -32,32 +32,45 @@ import (
 var (
 	configFlags *genericclioptions.ConfigFlags
 
-	duration           time.Duration
-	maxSize            int
-	packetSize         int
-	nodeSelectors      string
-	podSelectors       string
-	namespaceSelectors string
-	nodeNames          string
-	hostPath           string
-	pvc                string
 	blobUpload         string
-	s3Region           string
-	s3Endpoint         string
-	s3Bucket           string
-	s3Path             string
-	s3AccessKeyID      string
-	s3SecretAccessKey  string
-	tcpdumpFilter      string
+	debug              bool
+	duration           time.Duration
 	excludeFilter      string
+	hostPath           string
 	includeFilter      string
 	includeMetadata    bool
-	namespace          string
 	jobNumLimit        int
+	maxSize            int
+	namespace          string
+	namespaceSelectors string
+	nodeNames          string
+	nodeSelectors      string
+	nowait             bool
+	packetSize         int
+	podSelectors       string
+	pvc                string
+	s3AccessKeyID      string
+	s3Bucket           string
+	s3Endpoint         string
+	s3Path             string
+	s3Region           string
+	s3SecretAccessKey  string
+	tcpdumpFilter      string
+)
 
-	nowait bool
-
-	debug bool
+const (
+	DefaultDebug           bool          = false
+	DefaultDuration        time.Duration = 1 * time.Minute
+	DefaultHostPath        string        = "/mnt/retina/captures"
+	DefaultIncludeMetadata bool          = true
+	DefaultJobNumLimit     int           = 0
+	DefaultMaxSize         int           = 100
+	DefaultNodeSelectors   string        = "kubernetes.io/os=linux"
+	DefaultNowait          bool          = true
+	DefaultPacketSize      int           = 0
+	DefaultS3Path          string        = "retina/captures"
+	DefaultWaitPeriod      time.Duration = 1 * time.Minute
+	DefaultWaitTimeout     time.Duration = 5 * time.Minute
 )
 
 var createExample = templates.Examples(i18n.T(`
@@ -91,14 +104,9 @@ var createExample = templates.Examples(i18n.T(`
 			--s3-secret-access-key "your-secret-access-key"
 		`))
 
-const (
-	defaultWaitTimeout time.Duration = 5 * time.Minute
-	defaultWaitPeriod  time.Duration = 1 * time.Minute
-)
-
 var createCapture = &cobra.Command{
 	Use:     "create",
-	Short:   "create a Retina Capture",
+	Short:   "Create a Retina Capture",
 	Example: createExample,
 	RunE: func(*cobra.Command, []string) error {
 		kubeConfig, err := configFlags.ToRESTConfig()
@@ -244,6 +252,14 @@ func createCaptureF(kubeClient kubernetes.Interface) (*retinav1alpha1.Capture, e
 		capture.Spec.CaptureConfiguration.CaptureOption.Duration = &metav1.Duration{Duration: duration}
 	}
 
+	if namespaceSelectors != "" || podSelectors != "" {
+		// if node selector is using the default value (aka hasn't been set by user), set it to nil to prevent clash with namespace and pod selector
+		if nodeSelectors == DefaultNodeSelectors {
+			retinacmd.Logger.Info("Overriding default node selectors value and setting it to nil. Using namespace and pod selectors. To use node selector, please remove namespace and pod selectors.")
+			nodeSelectors = ""
+		}
+	}
+
 	nodeSelectorLabelsMap, err := labels.ConvertSelectorToLabelsMap(nodeSelectors)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


pkg/capture/constants/create_capture_defaults.go
```diff
@@ -0,0 +1,35 @@
...
+	DefaultPVC                string        = ""
+	DefaultS3AccessKeyID      string        = ""
+	DefaultS3Bucket           string        = ""
+	DefaultS3Endpoint         string        = ""
+	DefaultS3Path             string        = "retina/captures"
+	DefaultS3Region           string        = ""
+	DefaultS3SecretAccessKey  string        = ""
+	DefaultTcpdumpFilter      string        = ""
+	DefaultWaitPeriod         time.Duration = 1 * time.Minute
+	DefaultWaitTimeout        time.Duration = 5 * time.Minute
```
timraymond, line 34: I don't think these do much to improve readability... they just add a layer of indirection when navigating the flags. If a linter demands this, I'd say just silence it.

pkg/capture/utils/capture_image.go
```diff
@@ -40,7 +40,8 @@ func CaptureWorkloadImage(logger *log.ZapLogger, imageVersion string, debug bool
 
 	// For testing.
 	if debug {
-		if captureWorkloadImageFromEnv := os.Getenv(captureWorkloadImageEnvKey); len(captureWorkloadImageFromEnv) != 0 {
+		captureWorkloadImageFromEnv := os.Getenv(captureWorkloadImageEnvKey)
+		if len(captureWorkloadImageFromEnv) != 0 {
```
timraymond, line 44: I realize this is pre-existing, but if you do `captureWorkloadImageFromEnv == ""`, the compiler understands what you want and will intrisify this to exactly what you have here or something better.

pkg/capture/constants/create_capture_defaults.go
rbtr: I don't mind it IFF they're in the commands' package instead of centralized like this where you have to hunt for them

cli/cmd/capture/create.go
```diff
@@ -443,32 +436,32 @@ func deleteJobs(kubeClient kubernetes.Interface, jobs []batchv1.Job) []string {
...
-	createCapture.Flags().IntVar(&jobNumLimit, "job-num-limit", 0, "The maximum number of jobs can be created for each capture. 0 means no limit")
-	createCapture.Flags().BoolVar(&nowait, "no-wait", true, "Do not wait for the long-running capture job to finish")
-	createCapture.Flags().BoolVar(&debug, "debug", false, "When debug is true, a customized retina-agent image, determined by the environment variable RETINA_AGENT_IMAGE, is set")
+	createCapture.Flags().BoolVar(&includeMetadata, "include-metadata", captureConstants.DefaultIncludeMetadata, "If true, collect static network metadata into capture file")
+	createCapture.Flags().IntVar(&jobNumLimit, "job-num-limit", captureConstants.DefaultJobNumLimit, "The maximum number of jobs can be created for each capture. 0 means no limit")
+	createCapture.Flags().BoolVar(&nowait, "no-wait", captureConstants.DefaultNowait, "Do not wait for the long-running capture job to finish")
+	createCapture.Flags().BoolVar(&debug, "debug", captureConstants.DefaultDebug, "When debug is true, a customized retina-agent image, determined by the environment variable RETINA_AGENT_IMAGE, is set")
```
ritwikranjan, line 466: It's great to define constants for the default values, however for the ones that don't have any defaults, let's either leave it as is or just create single const for default string, bool and int and use them.

pkg/capture/constants/create_capture_defaults.go
ritwikranjan: Following from the comment in command package, we can possible have have a single const for go defaults i.e. empty string, false bool and 0 int

pkg/capture/crd_to_job.go
```diff
@@ -519,9 +520,30 @@ func (translator *CaptureToPodTranslator) validateTargetSelector(captureTarget r
...
+		if captureTarget.NodeSelector != nil {
+			// if node selector is using the default value (aka hasn't been set by user), set it to nil to prevent clash with namespace and pod selector
+			if reflect.DeepEqual(captureTarget.NodeSelector.MatchLabels, defaultNodeSelectors) {
+				translator.l.Info("Overriding default node selector value and setting it to nil. Using namespace and pod selectors. To use node selector, please remove namespace and pod selectors.")
+				captureTarget.NodeSelector = nil
+				return nil
+			} else {
+				return fmt.Errorf("NodeSelector is not compatible with NamespaceSelector&PodSelector. Please use one or the other")
+			}
+		}
```
ritwikranjan, line 544: I am personally not a fan of reflect so a suggestion to do it the old way, feel free ignore it as well.

```suggestion
if captureTarget.NodeSelector != nil && (captureTarget.NamespaceSelector != nil || captureTarget.PodSelector != nil) {
    defaultNodeSelectors := strings.Split(captureConstants.DefaultNodeSelectors, ",")
    isDefault := true
    for _, selector := range defaultNodeSelectors {
        kv := strings.Split(selector, "=")
        if len(kv) != 2 || captureTarget.NodeSelector.MatchLabels[kv[0]] != kv[1] {
            isDefault = false
            break
        }
    }
    if isDefault {
        translator.l.Info("Overriding default node selector value and setting it to nil. Using namespace and pod selectors. To use node selector, please remove namespace and pod selectors.")
        captureTarget.NodeSelector = nil
        return nil
    }
    return fmt.Errorf("NodeSelector is not compatible with NamespaceSelector&PodSelector. Please use one or the other")
}
```

pkg/capture/crd_to_job.go
```diff
@@ -519,9 +520,30 @@ func (translator *CaptureToPodTranslator) validateTargetSelector(captureTarget r
 	if captureTarget.NodeSelector == nil && captureTarget.PodSelector == nil {
 		return fmt.Errorf("Neither NodeSelector nor NamespaceSelector&PodSelector is set.")
 	}
```
ritwikranjan, line 522: Out of curousity, What if someone goes ahead and sets namespaceSelector, but not set the podSelector, do we just ignore it in that case?

ritwikranjan (commented):
Some minor suggestions!

pkg/capture/utils/capture_image.go
kamilprz: Changed to use equality operator rather than checking length, thanks!

pkg/capture/constants/create_capture_defaults.go
kamilprz: Thanks for feedback, got rid of the `create_capture_defaults` file and moved the defaults to the create commands package directly. Also got rid of all the empty defaults as suggested. Hope this is a happy compromise.

cli/cmd/capture/create.go
kamilprz: Got rid of all the empty defaults

pkg/capture/crd_to_job.go
kamilprz: I moved the logic away from this file into the create package itself. It simplifies the check a lot and doesn't require exporting the default values. Think that should be better.

pkg/capture/crd_to_job.go
kamilprz: If namespace selector is defined, but pod selector one of two things will happen.

if NodeSelector is nil
The if statement you highlighted will be hit and an error is thrown

if NodeSelector is not nil
The next if statement will be hit and an error is thrown (that NodeSelector and Namespace&PodSelectors are incompatible)

kamilprz: > the logic I decided on will check whether NodeSelector is the default value, if yes then it will override it to use Pod&Namespace, if its an explicit value (aka not default) an error is thrown.

I am open to feedback though if folks think that NodeSelector shouldn't be the default if no parameters are provided.

ritwikranjan (approved):

ritwikranjan (approved):

PR #860, opened 2024-10-15
fix(docs): update doc, codespaces config and GHA on commit GPG signed

SRodi:
# Description

* update contributing doc to add a note about signing commits by default
* add instructions for  codespaces users to use GPG to sign commits

## Related Issue

## Checklist

image

## Additional Notes

---

### DIFF
--- a/docs/07-Contributing/developing.md
+++ b/docs/07-Contributing/developing.md
@@ -190,8 +190,25 @@ In order to certify the provenance of commits and defend against impersonation, 
 Documentation for setting up Git and Github to sign your commits can be found [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
 Additional information about Git's use of GPG can be found [here](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work)
 
+This can be accomplished by providing a `-S` flag to `git commit` as documented [here](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--Sltkeyidgt)
+
 ### Developers Certificate of Origin (DCO)
 
 Contributions to Retina must contain a Developers Certificate of Origin within their constituent commits.
 This can be accomplished by providing a `-s` flag to `git commit` as documented [here](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s).
 This will add a `Signed-off-by` trailer to your Git commit, affirming your acceptance of the Contributor License Agreement.
+
+### Example commit
+
+Here is an example development flow to add a change made to file `docs/07-Contributing/developing.md`
+
+```sh
+git checkout -b feat-branch-1
+git add docs/07-Contributing/developing.md
+git commit -m "fix(doc): update contributing docs" -sS
+git push origin feat-branch-1 -u
+```
+
+After committing your change, when accessing [retina project's repo](https://github.com/microsoft/retina) you will get a prompt to create a PR from your fork.
+
+> Note: the DCO is not mandatory (i.e. the `-s` flag) - however, the cryptographically signing of commit is mandatory for all contributors. Hence, please make sure to always include the `-S` flag to your `git commit` instructions. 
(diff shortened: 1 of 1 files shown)


kamilprz (dismissed):

docs/07-Contributing/developing.md
```diff
@@ -190,8 +190,25 @@ In order to certify the provenance of commits and defend against impersonation,
 Documentation for setting up Git and Github to sign your commits can be found [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
 Additional information about Git's use of GPG can be found [here](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work)
 
+This can be accomplished by providing a `-S` flag to `git commit` as documented [here](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--Sltkeyidgt)
```
timraymond, line 193: You can also do `git config commit.gpgSign true` and forget about the flag. You probably don't want `--global` in this instance (unless you really want to sign everything, I do not).

docs/07-Contributing/developing.md
```diff
@@ -190,8 +190,25 @@ In order to certify the provenance of commits and defend against impersonation,
...
+```sh
+git checkout -b feat-branch-1
+git add docs/07-Contributing/developing.md
+git commit -m "fix(doc): update contributing docs" -sS
+git push origin feat-branch-1 -u
+```
+
+After committing your change, when accessing [retina project's repo](https://github.com/microsoft/retina) you will get a prompt to create a PR from your fork.
+
+> Note: the DCO is not mandatory (i.e. the `-s` flag) - however, the cryptographically signing of commit is mandatory for all contributors. Hence, please make sure to always include the `-S` flag to your `git commit` instructions.
```
timraymond, line 214: This contradicts the DCO paragraph. We should be consistent.

SRodi: agree, removed.

docs/07-Contributing/developing.md
SRodi: @timraymond I agree and I have that set too in my local gitconfig. However, this is literally the first tip of the [Signing Commits doc](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) which is already shared in this section.

The reason why I added that sentence is because we constantly have commits that do not have signature verification and this is causing unnecessary engagement from contributors/maintainers.

I rather incentivize contributors to read the GitHub docs from the provided link (and use `git config commit.gpgSign true`), or at the very least read this sentence I just added and use the `-S` flag.

timraymond: @SRodi The docs are not perfect. Providing `-S` is an administrative control that will eventually be forgotten by someone, requiring continued muda circling back with contributors. Telling contributors to modify their git config for this repo is an engineering control that forever solves this problem for a particular contributor. Honestly, I wouldn't be opposed to a `Makefile` target that just does it for you.

I'm not suggesting we change your existing sentence, but I'm saying we should add a recommendation to set `commit.gpgSign` for this repo.

rbtr: maybe in here, as a default for the codespaces users
https://github.com/microsoft/retina/blob/main/.devcontainer/installMoreTools.sh

SRodi: thanks for the comments! I have updated this PR to include a check in GH workflow to verify GPG signature exist, updated the doc to explicitly suggest using `git config commit.gpgSign true` and added the same instruction to codespaces setup (installMOreTools.sh) as mentioned above.

.github/workflows/commit-verifier.yaml
```diff
@@ -36,3 +39,12 @@ jobs:
             echo "For more details, see .github/workflows/commit-message.yaml"
             exit 1
           }
+
+      - name: Verify GPG Signature
+        run: |
+            LAST_COMMIT=$(git rev-parse HEAD)
```
kamilprz, line 45: Since this only looks at the last commit, what if I make many commits locally and only sign the most recent one, then push. In that case this check will pass from my understanding? Is that a problem?

SRodi: Good catch. I'll remove this altogether. I think for now we can just update the docs which was the original intention of this PR.

kamilprz (approved):

Issue #286, opened 2024-04-18
Feature Request: Outputting the .tar FileName as part of Retina Output please. ☕️


Tatsinnit:
**Is your feature request related to a problem? Please describe.**

Can we please get the actual filename printed once Retina CLI distributed Capture command ran successfully. cc: @rbtr and @vakalapa ☕️💡

Issue this will resolve is to the consuming tools which can then use the output and use Kubectl copy to specifically only copy that file to local machine.

So say for example if I have retina run if I get presented with the specific filename I can parse the output and only copy that tar from the node to local host. 

<img width="658" alt="Screenshot 2024-04-17 at 9 31 05 PM" src="https://github.com/microsoft/retina/assets/6233295/975784ea-2c91-4f6d-8399-83c17bdf0813">

**Describe the solution you'd like**

Can we please get the actual filename printed once Retina CLI distributed Capture command ran successfully.

**Describe alternatives you've considered**

Alternatively for now I just copy the whole of Capture folder from node to locally, as I presume this tool will not be used beyond diagnostics reasons.

**Additional context**

rbtr:
The challenge here is that the timestamp that is appended to the capture name is generated by the capture Job while it is running on the Node itself. So, the CLI doesn't know what that suffix is going to be, especially if you execute it with `--no-wait`.
The rest of the name is discoverable, it's `Capture Name - Node Name - (?Timestamp?) .tar.gz`.

I think sticking the timestamp on the end of a network capture is pretty standard for this kind of tool where you might be running it over and over and collecting data over time, and I think we should keep that. But maybe the timestamp could be generated by the CLI and passed in to the Job? This means the CLI knows what the tgz filenames will be, and also that they are predictable/consistent across all the Nodes? 

There is a minor semantic shift in that the timestamp is the time of request of the capture, and not the true start time of the capture, but the slip should be small...as long as we document those semantics clearly I think it's okay.

Tatsinnit:
> The challenge here is that the timestamp that is appended to the capture name is [generated by the capture Job](https://github.com/microsoft/retina/blob/96d42f9e6439ed6d12821ef416983df5165ccbb1/pkg/capture/provider/network_capture_common.go#L23) while it is running on the Node itself. So, the CLI doesn't know what that suffix is going to be, especially if you execute it with `--no-wait`. The rest of the name is discoverable, it's `Capture Name - Node Name - (?Timestamp?) .tar.gz`.
> 
[...]

PR #894, opened 2024-10-25
feat: Print capture output name when running retina capture create

kamilprz:
# Description

This PR adds functionality to the `retina capture create` CLI to print out the expected file output name of the capture. 

This is done by shifting the logic of the timestamp to when the CLI command is ran (rather than done after the capture is finished by the job). This means that the timestamp must be passed around for the jobs to actually generate the same filename as was presented to the user.

Screenshot below is an example of a capture job and the expected file name, compared with the actual file name generated on the node.

## Related Issue

https://github.com/microsoft/retina/issues/286

## Checklist

image

---


pkg/capture/provider/network_capture_unix.go
```diff
@@ -38,8 +40,8 @@ func NewNetworkCaptureProvider(logger *log.ZapLogger) NetworkCaptureProviderInte
 	}
 }
 
-func (ncp *NetworkCaptureProvider) Setup(captureName, nodeHostname string) (string, error) {
-	captureFolderDir, err := ncp.NetworkCaptureProviderCommon.Setup(captureName, nodeHostname)
+func (ncp *NetworkCaptureProvider) Setup(captureName, nodeHostname string, startTimestamp time.Time) (string, error) {
+	captureFolderDir, err := ncp.NetworkCaptureProviderCommon.Setup(captureName, nodeHostname, startTimestamp)
```
ritwikranjan, line 44: Following from the comment on the prev file: Instead of passing timestamp down the `Setup` call chain, we can use ncp.StartTimestamp, and similarly for other objects implementing this interface? I am assuming it is going to be constant for the lifecycle of capture

ritwikranjan (commented):
Some comments related to structure mostly, overall looks good to me.

pkg/capture/utils/filename.go
```diff
@@ -0,0 +1,27 @@
+package utils
```
timraymond, line 1: Having a `utils` package at all is kind of an antipattern. I think what you have here can be done with a type:

```go
package capture

// snip

type Time struct {
       time.Time
}

func (t Time) String() {
     // inline all the functions you have below in here
}

func NewTime(t string) (Time, error) {
     // everything from ConvertStringToTimestamp
}
```

Then we have our idea of what a `Time` is in the context of the capture package. This also makes it more type-oriented than procedural.

pkg/capture/capture_manager.go
```diff
@@ -96,6 +102,17 @@ func (cm *CaptureManager) captureNodeHostName() string {
 	return os.Getenv(captureConstants.NodeHostNameEnvKey)
 }
 
+func (cm *CaptureManager) captureStartTimestamp() (time.Time, error) {
+	tmp := os.Getenv(captureConstants.CaptureStartTimestampEnvKey)
+
+	timestamp, err := captureUtils.ConvertStringToTimestamp(tmp)
+	if err != nil {
+		cm.l.Error("Failed to parse the timestamp into the right format", zap.Error(err))
+		return time.Time{}, err
```
timraymond, line 111: It's best to either log or return errors wrapped with context to someone that will log them for you. The latter is generally preferred in the Go community. `(github.com/pkg/errors).Wrap` is what this team has decided to use for that.

pkg/capture/capture_manager.go
```diff
@@ -96,6 +102,17 @@ func (cm *CaptureManager) captureNodeHostName() string {
 	return os.Getenv(captureConstants.NodeHostNameEnvKey)
 }
 
+func (cm *CaptureManager) captureStartTimestamp() (time.Time, error) {
+	tmp := os.Getenv(captureConstants.CaptureStartTimestampEnvKey)
```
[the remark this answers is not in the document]
kamilprz: I always get called out on this habit :D

pkg/capture/utils/filename.go
kamilprz: Thanks, changed the code to take this approach. Let me know what you think :)

pkg/capture/capture_manager.go
kamilprz: Added wrapper

pkg/capture/file/timestamp.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/pkg/errors"
+)
+
+type Timestamp struct {
+	time.Time
+}
+
+const captureFileNameTimestampFormat string = "2006#01#02#15#04#05UTC"
+
+func (timestamp *Timestamp) TimestampToString() string {
```
timraymond, line 15: Sorry, I missed a detail in my earlier explanation. In Go, if you want to say "this is how this type should be represented as a string," you should implement [`fmt.Stringer`](https://pkg.go.dev/fmt#Stringer). That allows you to just pass this to basically anything in `fmt` like this:

```go
now := &Timestamp{time.Now().UTC()}
fmt.Println("The time is", now)
```

Instead of having to do this (and discover that the method exists):

```go
now := &Timestamp{time.Now().UTC()}
fmt.Println("the time is", now.TimestampToString())
```

So all you should need to do is this:
```suggestion
func (timestamp *Timestamp) String() string {
```

pkg/capture/file/capture_filename.go
```diff
@@ -0,0 +1,18 @@
...
+	"strings"
+)
+
+type CaptureFilename struct {
+	CaptureName    string
+	NodeHostname   string
+	StartTimestamp *Timestamp
+}
+
+func (cf *CaptureFilename) GenerateCaptureFileName() string {
```
timraymond, line 14: This can also likely be a `fmt.Stringer` (see other comment)
```suggestion
func (cf *CaptureFilename) String() string {
```

pkg/capture/file/timestamp.go
kamilprz: Very cool, thanks! Didn't know about this.

timraymond (approved):
This LGTM, but I did notice the request for @Tatsinnit to leave a review. Might be best to wait for their feedback as well.

Tatsinnit (approved):
❤️ Thank you so much for the kind ping, Looks Good and I think from the output form the consuming tool we can now pattern match the exact file name and then fetch it from that specific run, looks great as v1 until someone comes out with any other suggestions.

PR #903, opened 2024-10-29
fix(capture): ignore known copy failure and fix iptables issue

mainred:
# Description

- correct the logic to pick iptables mode
- ignore copy command failure

## Related Issue

## Checklist

image

## Additional Notes

None

---


mainred:
We install iptables legacy on Mariner image when building the image, but when it runs on ubuntu host, the command returns empty result, which works fine on iptables nft mode.

pkg/capture/provider/network_capture_unix.go
```diff
@@ -361,16 +364,44 @@ func (ncp *NetworkCaptureProvider) Cleanup() error {
...
+	nftIptablesMode    iptablesMode = "nft"
+)
+
+func obtainIptablesMode(l *log.ZapLogger) iptablesMode {
 	// Since iptables v1.8, nf_tables are introduced as an improvement of legacy iptables, but provides the same user
 	// interface as legacy iptables through iptables-nft command.
 	// based on: https://github.com/kubernetes-sigs/iptables-wrappers/blob/97b01f43a8e8db07840fc4b95e833a37c0d36b12/iptables-wrapper-installer.sh
-	legacySaveOut, _ := exec.Command("iptables-legacy-save").CombinedOutput()
+
+	// when both iptables modes available, we choose the one with more rules.
```
anubhabMajumdar, line 379: What's the thinking behind this? Can you not list all the rules using either one of the modes?  Given one is named legacy, shouldn't `nft` be default?

pkg/capture/provider/network_capture_unix.go
```diff
@@ -361,16 +364,44 @@ func (ncp *NetworkCaptureProvider) Cleanup() error {
...
 
-func obtainIptablesMode() string {
+type iptablesMode string
+
+const (
+	legacyIptablesMode iptablesMode = "legacy"
+	nftIptablesMode    iptablesMode = "nft"
+)
+
+func obtainIptablesMode(l *log.ZapLogger) iptablesMode {
```
anubhabMajumdar, line 374: I think you should return unhandled errors along with mode.

pkg/capture/provider/network_capture_unix.go
mainred: When both modes available, we should use the one list more rules, and the one list less rules normally just return empty rules.

I am referring to https://github.com/kubernetes-sigs/iptables-wrappers/blob/97b01f43a8e8db07840fc4b95e833a37c0d36b12/iptables-wrapper-installer.sh

pkg/capture/provider/network_capture_unix.go
mainred: The ideas in my mind is the failed iptables rules does not break the whole capture.

pkg/capture/provider/network_capture_unix.go
```diff
@@ -361,16 +364,44 @@ func (ncp *NetworkCaptureProvider) Cleanup() error {
...
 	// Since iptables v1.8, nf_tables are introduced as an improvement of legacy iptables, but provides the same user
 	// interface as legacy iptables through iptables-nft command.
 	// based on: https://github.com/kubernetes-sigs/iptables-wrappers/blob/97b01f43a8e8db07840fc4b95e833a37c0d36b12/iptables-wrapper-installer.sh
-	legacySaveOut, _ := exec.Command("iptables-legacy-save").CombinedOutput()
+
+	// when both iptables modes available, we choose the one with more rules.
+	nftIptablesModeAvaiable := true
+	legacyIptablesModeAvaiable := true
+	legacySaveOut, err := exec.Command("iptables-legacy-save").CombinedOutput()
+	if err != nil && strings.Contains(err.Error(), "command not found") {
```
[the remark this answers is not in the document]
mainred: I treat all command errors as unavailability of that mode and log the error for us to investigate.

timraymond: Grepping the error string is an anti-pattern. To accomplish this, you're better off using `exec.LookPath("iptables-legacy-save")`. This looks up the executable in the same way and expands it to an absolute path. It returns an error if the executable was not found. If you still need to run `iptables-legacy-save`, you can pass that absolute path to `exec.Command` and just handle any unexpected errors without trying to interpret them.

mainred: Hey @timraymond, I realized using grep is not good also considering the "not found" error message varies in multiple platforms. So the latest change is just to decide the right mode from the output, normally the right mode should gives more results/rules than the unexpected one.

mainred:
ping @timraymond and @anubhabMajumdar for another look.
cc @kamilprz

timraymond:
LGTM on @anubhabMajumdar 's approval

kamilprz:
Reopening and bringing it back to light @anubhabMajumdar @mainred

pkg/capture/provider/network_capture_unix.go
```diff
@@ -364,16 +367,46 @@ func (ncp *NetworkCaptureProvider) Cleanup() error {
...
 	// based on: https://github.com/kubernetes-sigs/iptables-wrappers/blob/97b01f43a8e8db07840fc4b95e833a37c0d36b12/iptables-wrapper-installer.sh
-	legacySaveOut, _ := exec.Command("iptables-legacy-save").CombinedOutput()
+
+	// When both iptables modes available, we choose the one with more rules, because the other one normally outputs empty rules.
+	nftIptablesModeAvaiable := true
+	legacyIptablesModeAvaiable := true
+	legacySaveOut, err := exec.Command("iptables-legacy-save").CombinedOutput()
+	if err != nil {
+		nftIptablesModeAvaiable = false
+		logger.Error("Failed to run iptables-legacy-save", zap.Error(err))
```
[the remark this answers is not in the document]
timraymond: @anubhabMajumdar It doesn't even really indicate that since weird things can go wrong when exec'ing under a stressed-out system. If we want to definitively test whether the command exists or not, the correct way to do that is with `exec.LookPath`. We should preflight each of these invocations to rule out the command not existing first (which can be done without exec()ing too). Then if we hit this arm of the error, we know definitively that it was something unusual in the actual execution of the command (probably OOM or something).

Also, in general, if it's happy-path control flow that the command does not exist (which we could be certain of with `exec.LookPath`, we don't need to log the error (since it's not an error). Once we are sure that it's an unusual error, we should _return_ the error from this func, wrapped with context about what we were trying to do when the error was produced.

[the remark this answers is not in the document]
kamilprz: Latest commit uses `exec.LookPath` now - this can probably be resolved?

pkg/capture/provider/network_capture_unix.go
```diff
@@ -364,16 +367,46 @@ func (ncp *NetworkCaptureProvider) Cleanup() error {
...
+		return nftIptablesMode
+	}
+
+	// when only one iptables mode available, we choose the available one when the other one is not available.
+	// when both iptables modes are not available, we choose the nft mode and let the iptables command fail.
+	if !nftIptablesModeAvaiable {
+		return legacyIptablesMode
 	}
-	return "nft"
+	return nftIptablesMode
```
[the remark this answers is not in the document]
kamilprz: Latest commit returns error if no mode is available - resolving.

pkg/capture/provider/network_capture_unix.go
kamilprz: Latest commit now returns errors - we can probably resolve this?

pkg/capture/provider/network_capture_unix.go
```diff
@@ -285,9 +286,10 @@ func (ncp *NetworkCaptureProvider) CollectMetadata() error {
 					description: "networking stats",
 				},
 				{
-					name:        "cp",
-					args:        []string{"-r", "/proc/sys/net", filepath.Join(ncp.TmpCaptureDir, "proc-sys-net")},
-					description: "kernel networking configuration",
+					name:          "cp",
+					args:          []string{"-r", "/proc/sys/net", filepath.Join(ncp.TmpCaptureDir, "proc-sys-net")},
+					description:   "kernel networking configuration",
+					ignoreFailure: true,
```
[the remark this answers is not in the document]
kamilprz: The latest commit has a comment with an explanation

kamilprz:
Closing the PR and reopening it under my ownership at - https://github.com/microsoft/retina/pull/1798

Issue #910, opened 2024-10-29
Add Retina CLI command to create ephemeral container with network debugging tools


wedaly:
**Is your feature request related to a problem? Please describe.**

I want to be able to do adhoc, exploratory debugging of node and pod networking.
 
**Describe the solution you'd like**

```
kubectl retina sh pods/<pod>
kubectl retina sh nodes/<node>
```

* Start an ephemeral container in the specified node or pod. 
* Use an image based on Azure Linux with built-in networking tools (ping, curl, nslookup, tcpdump,  conntrack, iproute2, iptables, etc.). If the cluster is using Retina, then it should already have firewall configuration to download this new image from the same source as the other Retina images.
* Run an interactive shell (bash). 
* Assign the appropriate permissions to ensure the tools work (NET_ADMIN and NET_RAW). 
* (Optionally) mount host filesystem for inspecting CNI logs, which iptables executable is installed on the host, etc. Make this opt-in (`--mount-host-filesystem` flag) since most network troubleshooting doesn't need it. 

Future work could expand this to support Windows (image with PowerShell, use host process containers for node access), but to start I care most about Linux support.

**Describe alternatives you've considered**

* [Node-shell](https://github.com/kvaps/kubectl-node-shell/blob/master/kubectl-node_shell) gives full access to the node using "nsenter" on PID 1. 
   - **Overprivileged**: Starts privileged pods on customer nodes (CAP_SYS_ADMIN and access to /proc filesystem). 
   - **Alpine Image**: Defaults to an alpine image from DockerHub that is often blocked by customer firewalls or subject to DockerHub rate limits. Can override this using an env var, but still requires an image with `nsenter` installed (and Azure Linux 2.0+ do not have it installed).
    - **No pod debugging**: Cannot easily debug connectivity from within a pod network namespace. Users can find the netns for a pod, then use `ip netns exec`, but this is painful and error-prone. 
* [Kubectl exec](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_exec/) allows us to execute a process within a pod's namespace. 
    - **Pod image missing tools**: Pod images may lack networking tools or even a shell (more common now with distroless images).
    - **Host networking**: If there's no host networking pod, we can't test connectivity from node IP. 
    - **Pod crashing**: If the target pod is crashing, we cannot exec into it. 
[...]

PR #962, opened 2024-11-06
feat: implement retina shell CLI command

wedaly:
# Description

Implement `retina shell` CLI command for adhoc network debugging of nodes and pods.

## Related Issue

#910 

## Checklist

image

## Additional Notes

For testing, need to set `RETINA_SHELL_IMAGE_REPO` and `RETINA_SHELL_IMAGE_VERSION` until the retina-shell image is published.

---

### DIFF
--- a/cli/cmd/shell.go
+++ b/cli/cmd/shell.go
@@ -1,0 +1,173 @@
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/microsoft/retina/internal/buildinfo"
+	"github.com/microsoft/retina/shell"
+	"github.com/spf13/cobra"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/resource"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"k8s.io/kubectl/pkg/scheme"
+	"k8s.io/kubectl/pkg/util/templates"
+)
+
+var (
+	configFlags              *genericclioptions.ConfigFlags
+	matchVersionFlags        *cmdutil.MatchVersionFlags
+	retinaShellImageRepo     string
+	retinaShellImageVersion  string
+	mountHostFilesystem      bool
+	allowHostFilesystemWrite bool
+	hostPID                  bool
+	capabilities             []string
+	timeout                  time.Duration
+)
+
+var (
+	// AKS requires clusters to allow access to MCR, so use this repository by default.
+	defaultRetinaShellImageRepo = "mcr.microsoft.com/containernetworking/retina-shell"
+
+	// Default version is the same as CLI version, set at link time.
+	defaultRetinaShellImageVersion = buildinfo.Version
+
+	defaultTimeout = 30 * time.Second
+
+	errMissingRequiredRetinaShellImageVersionArg = errors.New("missing required --retina-shell-image-version")
+	errUnsupportedResourceType                   = errors.New("unsupported resource type")
+)
+
+var shellCmd = &cobra.Command{
+	Use:   "shell (NODE | TYPE[[.VERSION].GROUP]/NAME)",
+	Short: "[EXPERIMENTAL] Interactively debug a node or pod",
+	Long: templates.LongDesc(`
+	[EXPERIMENTAL] This is an experimental command. The flags and behavior may change in the future.
+
+	Start a shell with networking tools in a node or pod for adhoc debugging.
+
+	* For nodes, this creates a pod on the node in the root network namespace.
+	* For pods, this creates an ephemeral container inside the pod's network namespace.
+
+	You can override the default image used for the shell container with either
+	CLI flags (--retina-shell-image-repo and --retina-shell-image-version) or
+	environment variables (RETINA_SHELL_IMAGE_REPO and RETINA_SHELL_IMAGE_VERSION).
+	CLI flags take precedence over env vars.
+`),
+
+	Example: templates.Examples(`
+		# start a shell in a node
+		kubectl retina shell node0001
+
+		# start a shell in a node, with debug pod in kube-system namespace
+		kubectl retina shell -n kube-system node0001
+
+		# start a shell as an ephemeral container inside an existing pod
+		kubectl retina shell -n kube-system pod/coredns-d459997b4-7cpzx
+
+		# start a shell in a node, mounting the host filesystem to /host with ability to chroot
+		kubectl retina shell node001 --mount-host-filesystem --capabilities SYS_CHROOT
+
+		# start a shell in a node, with NET_RAW and NET_ADMIN capabilities
+		# (required for iptables and tcpdump)
+		kubectl retina shell node001 --capabilities NET_RAW,NET_ADMIN
+`),
+	Args: cobra.ExactArgs(1),
+	RunE: func(_ *cobra.Command, args []string) error {
+		// retinaShellImageVersion defaults to the CLI version, but that might not be set if the CLI is built without -ldflags.
+		if retinaShellImageVersion == "" {
+			return errMissingRequiredRetinaShellImageVersionArg
+		}
+
+		namespace, explicitNamespace, err := matchVersionFlags.ToRawKubeConfigLoader().Namespace()
+		if err != nil {
+			return fmt.Errorf("error retrieving namespace arg: %w", err)
+		}
+
+		// This interprets the first arg as either a node or pod (same as kubectl):
+		//   "node001"           -> node
+		//   "node/node001"      -> node
+		//   "pod/example-7cpzx" -> pod
+		r := resource.NewBuilder(configFlags).
+			WithScheme(scheme.Scheme, scheme.Scheme.PrioritizedVersionsAllGroups()...).
+			FilenameParam(explicitNamespace, &resource.FilenameOptions{}).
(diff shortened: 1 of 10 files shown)


cli/cmd/shell.go
```diff
@@ -0,0 +1,150 @@
...
+	retinaShellImageRepo     string
+	retinaShellImageVersion  string
+	mountHostFilesystem      bool
+	allowHostFilesystemWrite bool
+	hostPID                  bool
+	capabilities             []string
+	timeout                  time.Duration
+)
+
+const defaultRetinaShellImageRepo = "ghcr.io/microsoft/retina/retina-shell"
```
wedaly, line 30: Few questions about this part:

* I'd prefer to use MCR instead of ghcr, since AKS clusters are required to have access to MCR. Everything else in this repo seems to be using ghcr, but is it possible to use MCR instead? (can make sure the image is replicated there before merging this).
* Defaulting to the Retina CLI version for the image tag. This works when the version is set in release builds, but will fail otherwise since we're defaulting to 0.0.5 here https://github.com/microsoft/retina/blob/2b109a35e45e5bc8e6c32a1cdb3578dacddc772e/cli/cmd/version.go#L12-L14 could we bump this later to a version that includes retina shell? (once it exists)

cli/cmd/shell.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+		}
+	}
+	shellCmd.Flags().StringVar(&retinaShellImageRepo, "retina-shell-image-repo", defaultRetinaShellImageRepo, "The container registry repository for the image to use for the shell container")
+	shellCmd.Flags().StringVar(&retinaShellImageVersion, "retina-shell-image-version", Version, "The version (tag) of the image to use for the shell container")
+	shellCmd.Flags().BoolVarP(&mountHostFilesystem, "mount-host-filesystem", "m", false, "Mount the host filesystem to /host. Applies only to nodes, not pods.")
+	shellCmd.Flags().BoolVarP(&allowHostFilesystemWrite, "allow-host-filesystem-write", "w", false, "Allow write access to the host filesystem. Implies --mount-host-filesystem. Applies only to nodes, not pods.")
+	shellCmd.Flags().BoolVar(&hostPID, "host-pid", false, "Set HostPID on the shell container. Applies only to nodes, not pods.")
+	shellCmd.Flags().StringSliceVarP(&capabilities, "capabilities", "c", []string{}, "Add capabilities to the shell container")
+	shellCmd.Flags().DurationVar(&timeout, "timeout", 30*time.Second, "The maximum time to wait for the shell container to start")
```
wedaly, line 142: looking for feedback about how to structure these flags. This was the simplest way I could think of that supported all the use cases I wanted to support (see docs/06-Troubleshooting/shell.md)

shell/manifests_test.go
```diff
@@ -0,0 +1,91 @@
+package shell
```
wedaly, line 1: Wasn't sure how to test the other parts of this, as it requires a k8s cluster and interactive shell / tty, so I added unit tests for just the manifest generation. Open to suggestions about how to test this better, maybe adding E2E in a followup PR.

shell/validation.go
```diff
@@ -0,0 +1,25 @@
...
+func validateOperatingSystemSupportedForNode(ctx context.Context, clientset *kubernetes.Clientset, nodeName string) error {
+	node, err := clientset.CoreV1().
+		Nodes().
+		Get(ctx, nodeName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	osLabel := node.Labels["kubernetes.io/os"]
+	if osLabel != "linux" { // Only Linux supported for now.
```
wedaly, line 20: I tried supporting Windows, but got confused with how host process containers work, so I'm leaving this as a followup.

cli/cmd/shell.go
```diff
@@ -0,0 +1,150 @@
...
+		if err := r.Err(); err != nil {
+			return err
+		}
+
+		restConfig, err := matchVersionFlags.ToRESTConfig()
+		if err != nil {
+			return err
+		}
+
+		config := shell.Config{
```
wedaly, line 91: golang-ci lint is complaining that `config` isn't used, but it's definitely used on lines 105 and 108 below. Maybe it's getting confused because it's used in a closure?

cli/cmd/shell.go
timraymond: You could make it a var, have it set by default to GHCR and then patch this var at link time with MCR if needed.

cli/cmd/shell.go
timraymond: It's probably upset because it can't prove that the closure gets invoked, and it's trying to call `r.Visit` a liar. I'd just silence it since it's trying to be too smart and failing at it.

cli/cmd/shell.go
wedaly: yeah, that's an option. Do we do any patching like that today when building the retina kubectl plugin? (Ideally the official plugin release would use MCR so it will "just work" in any AKS cluster)

wedaly: `Version` got moved to buildinfo and the default fallback removed. I updated this PR to: (1) use MCR by default, and (2) use buildinfo.Version and check that it's actually set (may not be when built locally).

cli/cmd/shell.go
wedaly: hmm, apparently it's impossible to exclude typecheck errors https://golangci-lint.run/welcome/faq/#why-is-it-not-possible-to-skipignore-typecheck-errors

wedaly: possibly a bug in golangci-lint version 1.55.2. I don't see this error on later versions

wedaly: trying to fix it here: https://github.com/microsoft/retina/pull/1096

docs/06-Troubleshooting/shell.md
```diff
@@ -0,0 +1,185 @@
...
+export RETINA_SHELL_IMAGE_VERSION=v0.0.1 # optional, if not set defaults to the Retina CLI version.
+kubectl retina shell node0001 # this will use the image "example.azurecr.io/retina/retina-shell:v0.0.1"
+```
+
+## Limitations
+
+* Windows nodes and pods are not yet supported.
+* `bpftool` and `bpftrace` are not supported.
+* The shell image link `iptables` commands to `iptables-legacy`, even if the node itself links to `iptables-nft`.
+* `nsenter` is not supported.
```
nddq, line 184: i assume starting a shell inside the pod's network namespace would circumvent this?

docs/06-Troubleshooting/shell.md
```diff
@@ -0,0 +1,185 @@
...
+export RETINA_SHELL_IMAGE_REPO="example.azurecr.io/retina/retina-shell"
+export RETINA_SHELL_IMAGE_VERSION=v0.0.1 # optional, if not set defaults to the Retina CLI version.
+kubectl retina shell node0001 # this will use the image "example.azurecr.io/retina/retina-shell:v0.0.1"
+```
+
+## Limitations
+
+* Windows nodes and pods are not yet supported.
+* `bpftool` and `bpftrace` are not supported.
+* The shell image link `iptables` commands to `iptables-legacy`, even if the node itself links to `iptables-nft`.
```
nddq, line 183: is this an azure linux limitation? along with bpftool and bpftrace, which i think would be incredibly useful for Retina

cli/cmd/shell.go
```diff
@@ -0,0 +1,173 @@
...
+		if !cmd.Flags().Changed("retina-shell-image-version") {
+			if envVersion := os.Getenv("RETINA_SHELL_IMAGE_VERSION"); envVersion != "" {
+				retinaShellImageVersion = envVersion
+			}
+		}
+	}
+	shellCmd.Flags().StringVar(&retinaShellImageRepo, "retina-shell-image-repo", defaultRetinaShellImageRepo, "The container registry repository for the image to use for the shell container")
+	shellCmd.Flags().StringVar(&retinaShellImageVersion, "retina-shell-image-version", defaultRetinaShellImageVersion, "The version (tag) of the image to use for the shell container")
+	shellCmd.Flags().BoolVarP(&mountHostFilesystem, "mount-host-filesystem", "m", false, "Mount the host filesystem to /host. Applies only to nodes, not pods.")
+	shellCmd.Flags().BoolVarP(&allowHostFilesystemWrite, "allow-host-filesystem-write", "w", false,
```
nddq, line 161: having a text editor in the image would be useful if this option is set 🙂

nddq (approved):
some nit comments but otherwise lgtm. Thanks!

docs/06-Troubleshooting/shell.md
wedaly: yes, this is an Azure Linux thing. But they're going to switch to iptables-nft soon (like in the next month or so)

docs/06-Troubleshooting/shell.md
wedaly: yeah, I didn't see the point of using `nsenter` if you can start a debug container inside the pod netns directly. IIRC making nsenter work required `privileged=true` which I think should be discouraged

cli/cmd/shell.go
wedaly: we should just install a text editor in the image so it's always available

docs/06-Troubleshooting/shell.md
wedaly: bpftool and bpftrace will probably work, but I wasn't able to find the right set of capabilities when I tried it.

Issue #806, opened 2024-10-02
[conntrack] potentially useful metrics that leverage conntrack


nddq:
These can be useful as Prometheus's metrics:
- Packets and bytes count per connection.
- Number of connections going to world.

PR #1057, opened 2024-11-22
feat(ct-metrics): add initial conntrack metrics for Prometheus

SRodi:
# Description

Create initial metrics for conntrack:
- packets_count_per_connection [gauge] see example below
- bytes_count_per_connection [gauge]
- connections_count [count] **WIP**

## Related Issue

#806 

## Checklist

This example shows a long-lived TCP connection between client `src_ip="10.244.0..230` and server `10.244.1.85` 

image

This example shows the `bytes_count_per_connection` where `dst_ip=10.244.0.113` which is the pod IP for `metrics-server` on one of the cluster nodes. 

image

(WIP) Here is some initial work on Hubble metrics

image

## Considerations

Metrics have been added as `pod-level metrics` since the number of time series increases overtime based on number of new connections in the cluster - the count of time series is the same for both `packets_count_per_connection` and `bytes_count_per_connection` metrics since both use the same labels.

image

---


MikeZappa87:
Did we think though around scalability, performance and stability when adding these metrics? Do we have any concerns? Do we have numbers to back this up?

pkg/metrics/metrics.go
```diff
@@ -140,6 +140,26 @@ func InitializeMetrics() {
 		dnsResponseCounterDescription,
 	)
 
+	// Conntrack Metrics
+	ConntrackPacketsCounter = exporter.CreatePrometheusGaugeVecForMetric(
```
vakalapa, line 144: Can you find a way to integrate these into Hubble metrics as well ?

SRodi: still WIP, but some initial work is done, see this screenshot
image

SRodi:
after an offline discussion with @vakalapa and @nddq it was agreed to abandon this PR in favor of multiple PRs to break down the solution into:
* data plane implementation
* control plane Hubble
* control plane legacy
* advance metrics

It was also agreed to disable those metrics by default, with the use of a feature flag similar to `enableConntrackMetrics` 

cc @MikeZappa87

Issue #1040, opened 2024-11-20
Measure rate of growth of metrics


anubhabMajumdar:
Metrics generated by Retina is supposed to of low cardinality over time. If there's a label that is unbound, the metrics volume would constantly increase, causing OOMKIll issues for agent, and may overflow storage of these metrics on an external entity.

Add a test that
- generates lots of L4 events
- generate lots of DNS queries

Measure rate of growth of metrics volume to make sure we don't have unbound growth.

alexcastilio:
Implemented by #1235 and #1345

PR #1061, opened 2024-11-22
tool: update kapinger to make random dns queries

alexcastilio:
# Description

Update kapinger to make random dns queries

## Related Issue

## Checklist

## Additional Notes

---

### DIFF
--- a/hack/tools/kapinger/clients/dns.go
+++ b/hack/tools/kapinger/clients/dns.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"math/rand"
 	"net"
 	"time"
 )
@@ -30,7 +31,7 @@ func (k *KapingerDNSClient) MakeRequests(ctx context.Context) error {
 		case <-ticker.C:
 			go func() {
 				for i := 0; i < k.volume; i++ {
-					domain := "retina.sh"
+					domain := randomString(20) + ".test"
 
 					ips, err := net.LookupIP(domain)
 					if err != nil {
@@ -42,4 +43,17 @@ func (k *KapingerDNSClient) MakeRequests(ctx context.Context) error {
 			}()
 		}
 	}
+}
+
+func randomString(n int) string {
+	var letters = []rune("abcdefghijklmnopqrstuvwxyz")
+	b := make([]rune, n)
+
+	rand.New(rand.NewSource(time.Now().UnixNano()))
+
+	for i := range b {
+		b[i] = letters[rand.Intn(len(letters))]
+	}
+
+	return string(b)
 }


MikeZappa87:
What is the intention of this PR? This looks like it issues DNS queries for 'random' domain names? What is the scale? 1, 10, 100, 1000, 10000? After these queries are executed what is the consumer? Are we measuring something? This sounds like the consumer doesn't have a great means of testing.

hack/tools/kapinger/clients/dns.go
```diff
@@ -43,3 +44,16 @@ func (k *KapingerDNSClient) MakeRequests(ctx context.Context) error {
 		}
 	}
 }
+
+func randomString(n int) string {
+	var letters = []rune("abcdefghijklmnopqrstuvwxyz")
+	b := make([]rune, n)
+
+	rand.New(rand.NewSource(time.Now().UnixNano()))
```
MikeZappa87, line 52: Does the New function return a type? On Line 55 I would expect that to use the instance. 

r := rand.New(.....)

on line 55 ...

r.IntN(......)

I don't think this is doing what you think it is. Perhaps create a unit test for this function and verify.

MikeZappa87 (changes requested):

hack/tools/kapinger/clients/dns.go
timraymond: It will kind of work :) `rand.Intn` uses a package global logger, which I believe is initialized to the same seed. The new random source generated on this line is discarded, which is likely not intended.

hack/tools/kapinger/clients/dns.go
```diff
@@ -30,7 +31,7 @@ func (k *KapingerDNSClient) MakeRequests(ctx context.Context) error {
 		case <-ticker.C:
 			go func() {
 				for i := 0; i < k.volume; i++ {
-					domain := "retina.sh"
+					domain := randomString(20) + ".com"
```
timraymond, line 34: Please use `.test` instead of `.com` for tests involving DNS per [RFC 2606](https://www.rfc-editor.org/rfc/rfc2606.html)

timraymond, line 34: Please use `.test` instead of `.com` for tests involving DNS per [RFC 2606](https://www.rfc-editor.org/rfc/rfc2606.html)

timraymond (changes requested):

alexcastilio:
> What is the intention of this PR? This looks like it issues DNS queries for 'random' domain names? What is the scale? 1, 10, 100, 1000, 10000? After these queries are executed what is the consumer? Are we measuring something? This sounds like the consumer doesn't have a great means of testing.

The idea is to generate random DNS queries. This will be used to scale test retina which will be generating metrics for traffic on the cluster (both these DNS queries and other traffic generated). This was requested as part of https://github.com/microsoft/retina/issues/1040

hack/tools/kapinger/clients/dns.go
alexcastilio: Done

timraymond:
@MikeZappa87 pursuant to our earlier concerns about the external DNS traffic this might generate, it looks like `.test` is configured to black hole requests. The SOA record lists `localhost` as the resolver for `.test`:

```
; <<>> DiG 9.18.28 <<>> @4.2.2.2 test SOA
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 52782
;; flags: qr aa rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;test.                          IN      SOA

;; ANSWER SECTION:
test.                   10800   IN      SOA     localhost. nobody.invalid. 1 3600 1200 604800 10800

;; Query time: 10 msec
;; SERVER: 4.2.2.2#53(4.2.2.2) (UDP)
;; WHEN: Mon Nov 25 15:17:24 EST 2024
;; MSG SIZE  rcvd: 92
```

MikeZappa87:
> > What is the intention of this PR? This looks like it issues DNS queries for 'random' domain names? What is the scale? 1, 10, 100, 1000, 10000? After these queries are executed what is the consumer? Are we measuring something? This sounds like the consumer doesn't have a great means of testing.
> 
> The idea is to generate random DNS queries. This will be used to scale test retina which will be generating metrics for traffic on the cluster (both these DNS queries and other traffic generated). This was requested as part of #1040

The concern is that this could be hitting DNS servers that we don't own nor do we have an agreement to do this. It probably makes more sense to stand up a DNS server we control to do this or come up with a design change to test retina without doing this.

timraymond:
@MikeZappa87 I did some more research, and it depends very much on the resolver you ask. Cloudflare (1.1.1.1) and Google (8.8.8.8) both forward the request to the root servers, and *they* are the ones that respond with NXDOMAIN. Needless to say, flooding the root servers is A Very Bad Thing for the reasons you outline.

On the other hand, Level 3 (4.2.2.2) and Quad 9 (9.9.9.9) black hole the .test TLD in the way that I showed earlier (that request was to Level 3).

I agree though... even though this is the TLD to use, it's clear that resolvers handle it inconsistently (and probably shouldn't be trusted to consistently handle it). We should just stick BIND or CoreDNS in there to field all this traffic.

PR #1092, opened 2024-12-02
feat(healthchecks): Adding Example Health Checks

mereta:
# Description

## Related Issue

## Checklist

## Additional Notes

---

### DIFF
--- a/cmd/hubble/daemon_linux.go
+++ b/cmd/hubble/daemon_linux.go
@@ -7,6 +7,7 @@ package hubble
 import (
 	"context"
 	"fmt"
+	"os"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -32,6 +33,7 @@ import (
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	zapf "sigs.k8s.io/controller-runtime/pkg/log/zap"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
@@ -64,6 +66,16 @@ var (
 			if err != nil {
 				logger.Error("failed to create manager")
 				return nil, nil, fmt.Errorf("creating new controller-runtime manager: %w", err)
+			}
+
+			if err := ctrlManager.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+				logger.Error("unable to set up healthz check", err)
+				os.Exit(1)
+			}
+
+			if err := ctrlManager.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+				logger.Error("unable to set up readyz check", err)
+				os.Exit(1)
 			}
 
 			return ctrlManager, ctrlManager.GetClient(), nil
--- a/cmd/legacy/daemon.go
+++ b/cmd/legacy/daemon.go
@@ -4,6 +4,7 @@ package legacy
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"strings"
 	"time"
@@ -53,6 +54,10 @@ const (
 
 	nodeNameEnvKey = "NODE_NAME"
 	nodeIPEnvKey   = "NODE_IP"
+)
+
+var (
+	healthzChecker healthz.Checker
 )
 
 var scheme = k8sruntime.NewScheme()
@@ -215,14 +220,6 @@ func (d *Daemon) Start() error {
 	}
 
 	//+kubebuilder:scaffold:builder
-
-	if healthCheckErr := mgr.AddHealthzCheck("healthz", healthz.Ping); healthCheckErr != nil {
-		mainLogger.Fatal("Unable to set up health check", zap.Error(healthCheckErr))
-	}
-	if addReadyCheckErr := mgr.AddReadyzCheck("readyz", healthz.Ping); addReadyCheckErr != nil {
-		mainLogger.Fatal("Unable to set up ready check", zap.Error(addReadyCheckErr))
-	}
-
 	// k8s Client used for informers
 	cl := kubernetes.NewForConfigOrDie(mgr.GetConfig())
 
@@ -294,6 +291,21 @@ func (d *Daemon) Start() error {
 				mainLogger.Fatal("unable to create metricsConfigController", zap.Error(err))
 			}
 		}
+
+		// Define a custom health check for advanced metrics
+		healthzChecker = healthz.CheckHandler{
+			Checker: healthz.Checker(func(req *http.Request) error {
+				_, err := metricsModule.Status()
+				if err != nil {
+					mainLogger.Error("failed to get metrics module status fr advanced metrics", zap.Error(err))
+					return err
+				}
+				return nil
+			}),
+		}.Checker
+	} else {
+		// Advanced Metric not enabled, Ping healthcheck
+		healthzChecker = healthz.Ping
 	}
 
 	controllerMgr, err := cm.NewControllerManager(daemonConfig, cl, tel)
@@ -314,6 +326,17 @@ func (d *Daemon) Start() error {
 	// Start controller manager, which will start http server and plugin manager.
 	go controllerMgr.Start(ctx)
 	mainLogger.Info("Started controller manager")
+
+	//Set health checks according to retina confiuration
+	if err := mgr.AddHealthzCheck("healthz", healthzChecker); err != nil {
+		mainLogger.Error("unable to set up custom health check", zap.Error(err))
+		os.Exit(1)
+	}
+
+	if err := mgr.AddReadyzCheck("readyz", healthzChecker); err != nil {
+		mainLogger.Error("unable to set up custom ready check", zap.Error(err))
+		os.Exit(1)
+	}
 
 	// Start all registered controllers. This will block until container receives SIGTERM.
 	if err := mgr.Start(ctx); err != nil {
--- a/deploy/hubble/manifests/controller/helm/retina/templates/agent/daemonset.yaml
+++ b/deploy/hubble/manifests/controller/helm/retina/templates/agent/daemonset.yaml
@@ -77,10 +77,22 @@ spec:
               cpu: {{ .Values.resources.limits.cpu | quote }}
           readinessProbe:
             httpGet:
-              path: /metrics
-              port: {{ .Values.agent.container.retina.ports.containerPort }}
-            initialDelaySeconds: 10
-            periodSeconds: 30
+              path: /readyz
+              port: {{ .Values.agent.container.retina.ports.healthPort }}
+            initialDelaySeconds: {{ .Values.readinessProbe.initialDelaySeconds | default "30" }}
+            periodSeconds: {{ .Values.readinessProbe.periodSeconds | default "30" }}
+            timeoutSeconds: {{ .Values.readinessProbe.timeoutSeconds | default "1" }}
+            failureThreshold: {{ .Values.readinessProbe.failureThreshold | default "3" }}
+            successThreshold: {{ .Values.readinessProbe.successThreshold | default "1" }}
+          livenessProbe:
+            httpGet:
+              path: /healthz
+              port: {{ .Values.agent.container.retina.ports.healthPort }}
+            initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds | default "30" }}
+            periodSeconds: {{ .Values.livenessProbe.periodSeconds | default "30" }}
+            timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds | default "1" }}
+            failureThreshold: {{ .Values.livenessProbe.failureThreshold | default "3" }}
+            successThreshold: {{ .Values.livenessProbe.successThreshold | default "1" }}
           env:
           - name: POD_NAME
             valueFrom:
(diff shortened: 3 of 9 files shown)


cmd/hubble/daemon_linux.go
```diff
@@ -66,6 +68,16 @@ var (
...
+			if err := ctrlManager.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+				logger.Error("unable to set up healthz check", err)
+				os.Exit(1)
+			}
+
+			if err := ctrlManager.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+				logger.Error("unable to set up readyz check", err)
+				os.Exit(1)
+			}
+
```
mereta, line 80: Hubble Checks

cmd/legacy/daemon.go
```diff
@@ -294,6 +291,21 @@ func (d *Daemon) Start() error {
...
+				if err != nil {
+					mainLogger.Error("failed to get metrics module status fr advanced metrics", zap.Error(err))
+					return err
+				}
+				return nil
+			}),
+		}.Checker
+	} else {
+		// Advanced Metric not enabled, Ping healthcheck
+		healthzChecker = healthz.Ping
```
mereta, line 308: Example custom handler for advanced metrics or simple healthz.Ping for basic

pkg/enricher/enricher.go
```diff
@@ -189,3 +193,54 @@ func (e *Enricher) Write(ev *v1.Event) {
...
+		if ts.Before(since) {
+			// scanned the last minute, exit loop
+			break
+		}
+		lastSeenEvent = e
+		count++
+	}
+	fl := float64(count) / at.Sub(since).Seconds()
+	return fl, nil
+}
```
mereta, line 246: This is example check based on Hubble check

cmd/hubble/daemon_linux.go
```diff
@@ -66,6 +68,16 @@ var (
 				return nil, nil, fmt.Errorf("creating new controller-runtime manager: %w", err)
 			}
 
+			if err := ctrlManager.AddHealthzCheck("healthz", healthz.Ping); err != nil {
```
MikeZappa87, line 71: Is healthz.Ping the HTTP Handler for the health check?

mereta: Ping is like a placeholder function. It will always return 'true'

`// Ping returns true automatically when checked.
var Ping Checker = func(_ *http.Request) error { return nil }
`

https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/healthz/healthz.go

cmd/hubble/daemon_linux.go
```diff
@@ -66,6 +68,16 @@ var (
...
+				logger.Error("unable to set up healthz check", err)
+				os.Exit(1)
```
timraymond, line 73: `os.Exit` is a heavy hammer, to be used with extreme discretion. It bypasses all deferred functions in Go, which is the primary way that resource cleanup happens. It's actually slightly better to panic (but also do that as rarely as possible) since there's at least a chance of it getting `recover()`ed by something further up the call stack. **Always prefer to return errors** unless there's just really no way to do it. Save panics for "developer errors" where there's no way for the compiler to check something at build time and you want it to blow up loudly (think compiling a regexp at runtime that happens to be invalid).

For reference, I usually start most Go applications like this:

```go
package main

func main() {
       if err := run(); err != nil {
              fmt.Println(err)
              os.Exit(1) // this is the only place in the entire application this appears, and really only because I want the exit code to be 1 instead of 0.
       }
}

func run() error {
       if err := somethingBadCouldHappen(); err != nil {
              return fmt.Errorf("invoking something potentially bad: %w", err) 
       }
       return nil
}
```

cmd/legacy/daemon.go
```diff
@@ -315,6 +307,17 @@ func (d *Daemon) Start() error {
 	go controllerMgr.Start(ctx)
 	mainLogger.Info("Started controller manager")
 
+	//Set health checks
```
timraymond, line 310: Small style nit. Leading spaces after the `//` are preferred, unless there's some kind of pragma that the comment is trying to invoke (like `//go:build`, for example)
```suggestion
	// Set health checks
```

cmd/hubble/daemon_linux.go
mereta: @timraymond  That makes sense, though in this particular case, I'm wondering if it can recover? If, for example the health check creation fails and we just return an error, the /healthz endpoint is never set up, we still have the probe that will try and GET from this endpoint. If that fails the pods will keep restarting and won't become available.

cmd/hubble/daemon_linux.go
mereta: Updated

deploy/hubble/manifests/controller/helm/retina/templates/agent/daemonset.yaml
```diff
@@ -196,8 +208,8 @@ spec:
             - .\setkubeconfigpath.ps1; ./controller.exe --config ./retina/config.yaml --kubeconfig ./kubeconfig
           readinessProbe:
             httpGet:
-              path: /metrics
-              port: {{ .Values.agent.container.retina.ports.containerPort }}
+              path: /healthz
+              port: {{ .Values.agent.container.retina.ports.healthPort }}
```
mereta, line 212: @matmerr Question on the windows configurations in the hubble control plane manifests. Is there a reason we have this or should we do some housekeeping?

cmd/hubble/daemon_linux.go
timraymond: @mereta Sounds good. It might be okay in this instance, but in my experience it's not even worth reasoning about. Logic can be sound at this one moment in time, but as changes happen in the codebase someone still needs to remember that panic / recover is broken here and their defers will not run (in any goroutine). In general, it creates surprising behavior in the language and suppresses vital debugging information if something goes wrong. Being extremely judicious with `os.Exit` as a general principle will keep an entire class of spooky bugs out of the Go you write (honestly, we should have a linter for this).

rbtr: > honestly, we should have a linter for this.

kinda https://github.com/microsoft/retina/pull/1092/files/df90b99316baf690bd09504517875b8a31bd2328..a200860d63a1bb65c815f0707e9b3e8ba65ecdfc

timraymond: Oooh, that one is fancy. I'm impressed the static analysis is good enough to tell what defers won't run.

cmd/hubble/daemon_linux.go
```diff
@@ -66,6 +67,14 @@ var (
 				return nil, nil, fmt.Errorf("creating new controller-runtime manager: %w", err)
 			}
 
+			if err := ctrlManager.AddHealthzCheck("healthz", func(_ *http.Request) error { return nil }); err != nil {
+				logger.Error("unable to set up agent healthz check", err)
```
vakalapa, line 71: We have to look at upgrade scenario, what happens when an existing retina agent gets upgraded with this change, will it work fine ? Will downgrade be fine as well ?

mereta: Legacy:
0.0.17 manifest with 0.0.19-health-feature image - OK (/metrics functionality still there)
 
Hubble:
0.0.17 manifest with 0.0.19-health-feature image - OK (/metrics functionality still there)
 
Legacy:
0.0.19-health-feature manifest with 0.0.17 image - OK (/healthz & /readyz functionality not new)
 
Hubble:
0.0.19-health-feature manifest with 0.0.17 image - NOT OK  (/healthz & /readyz functionality NEW to retina Hubble) this would be a breaking change for retina hubble in case of downgrade

deploy/legacy/manifests/controller/helm/retina/values.yaml
```diff
@@ -86,6 +86,7 @@ daemonset:
       metricsBindAddress: ":18080"
       ports:
         containerPort: 10093
+        healthPort: 18081
```
vakalapa, line 89: Why create a new port? and not reuse the existing 10093 ? since we are a hostnet: true pod, any port number we use comes with restrictions as it takes away from nodes usable port ranges. In AKS, we will have to pre register the ports we will be using to make sure customers are also aware of the ports they should not use to avoid any conflict.

vakalapa (changes requested):

deploy/legacy/manifests/controller/helm/retina/values.yaml
mereta: This seems to be already configured in Legacy & Hubble control planes as default HealthProbeBindAddress.
Its always provided when we set up k8s controller runtime options.
Hubble: https://github.com/microsoft/retina/blob/4b12472cc5007e0931238dd5e26f819e6be093b5/pkg/config/hubble_config_linux.go#L42

Legacy: https://github.com/microsoft/retina/blob/4b12472cc5007e0931238dd5e26f819e6be093b5/cmd/root.go#L42C10-L42C15

I think also @rbtr mentioned having a configurable port.
Do you think we should change to 10093?

mereta: Ok I did some testing and got errors when using port 10093. 
I don't think we can reuse the Retina port, some findings below:

We start a http api server with that port in controller manager: https://github.com/microsoft/retina/blob/b27958f26d0d61fc82568b0174b667cdd89522ce/pkg/managers/controllermanager/controllermanager.go#L56
 
And the trail for setting up health check is:
The HealthProbeBindAddress we pass as part of options in the ctrl manager ends up here where a listener is created.
comment:
This will throw an error if the bind address is invalid or already in use.
(same logic as providing pprof address below)
https://github.com/kubernetes-sigs/controller-runtime/blob/aea2e32a936584b06ae6f7992f856fe7292b0297/pkg/manager/manager.go#L407
 
If listener exists k8s tries to add HealthProbeServer:
https://github.com/kubernetes-sigs/controller-runtime/blob/aea2e32a936584b06ae6f7992f856fe7292b0297/pkg/manager/internal.go#L400
 
addHealthProverServer() definition - creates a new server:
https://github.com/kubernetes-sigs/controller-runtime/blob/aea2e32a936584b06ae6f7992f856fe7292b0297/pkg/manager/internal.go#L290

timraymond: Yeah, I agree with Vamsi. Controller Manager's implementation isn't fit for purpose here, so we should just use the standard library and write the two HTTP handlers on whatever server is bound to 10093. It will probably provide a better healthcheck overall, since we're testing whether traffic can be served on 10093.

mereta: @vakalapa @timraymond Updated the PR.

ritwikranjan (commented):
Mostly opinions and curious question rather than any explicit suggestion.

deploy/hubble/manifests/controller/helm/retina/templates/agent/daemonset.yaml
```diff
@@ -80,10 +80,22 @@ spec:
               cpu: {{ .Values.resources.limits.cpu | quote }}
           readinessProbe:
             httpGet:
-              path: /metrics
+              path: /health
               port: {{ .Values.agent.container.retina.ports.containerPort }}
```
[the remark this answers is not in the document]
mereta: For hubble we can use hubble-status, which what the original PR. I'm not sure if for this its out of scope.
But something to consider for sure.

deploy/legacy/manifests/controller/helm/retina/templates/daemonset.yaml
```diff
@@ -49,17 +49,17 @@ spec:
...
+              port: {{ .Values.daemonset.container.retina.ports.containerPort }}
             initialDelaySeconds: {{ .Values.readinessProbe.initialDelaySeconds | default "30" }}
             periodSeconds: {{ .Values.readinessProbe.periodSeconds | default "30" }}
             timeoutSeconds: {{ .Values.readinessProbe.timeoutSeconds | default "1" }}
             failureThreshold: {{ .Values.readinessProbe.failureThreshold | default "3" }}
             successThreshold: {{ .Values.readinessProbe.successThreshold | default "1" }}
           livenessProbe:
             httpGet:
-              path: /metrics
-              port: {{ .Values.retinaPort }}
```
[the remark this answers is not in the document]
mereta: Not sure both are defined in values already. Maybe a good idea to just use one of them. I could switch to use values.retinaPort

deploy/hubble/manifests/controller/helm/retina/templates/agent/daemonset.yaml
```diff
@@ -80,10 +80,22 @@ spec:
...
-            initialDelaySeconds: 10
-            periodSeconds: 30
+            initialDelaySeconds: {{ .Values.readinessProbe.initialDelaySeconds | default "30" }}
+            periodSeconds: {{ .Values.readinessProbe.periodSeconds | default "30" }}
+            timeoutSeconds: {{ .Values.readinessProbe.timeoutSeconds | default "1" }}
+            failureThreshold: {{ .Values.readinessProbe.failureThreshold | default "3" }}
+            successThreshold: {{ .Values.readinessProbe.successThreshold | default "1" }}
```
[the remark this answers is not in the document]
mereta: Done.

deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -939,3 +939,29 @@ tls:
...
+## @param readinessProbe.failureThreshold [array] Failure threshold for readinessProbe
+## @param readinessProbe.successThreshold [array] Success threshold for readinessProbe
+## Ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes
+##
+readinessProbe:
+  initialDelaySeconds: 30
+  periodSeconds: 30
+  timeoutSeconds: 15
+  failureThreshold: 3
+  successThreshold: 1
```
mereta, line 954: @anubhabMajumdar Any input on the values? Do they look ok?

Issue #1071, opened 2024-11-26
add kernel version to telemetry traces.


jimassa:
heartbeat already comes with kernel version attribute; https://github.com/microsoft/retina/blob/5ff1567e8a988209494b0cddd34f3939ac51e53c/pkg/telemetry/telemetry.go#L170

We would like to add it to traces as well in order to directly pinpoint what kernel version is producing a specific log.

Anirudh2112:
Hi @jimassa, I would like to contribute to this

Anirudh2112:
Hello @jimassa I have made the necessary changes and committed a PR. Kindly review it. Thank you.

matmerr:
isn't it possible to correlate pod name with kernel version in heartbeat?

Anirudh2112:
Thank you for the suggestion! I will add the pod name to the heartbeat properties to enable correlation with the kernel version. I'll update the PR shortly. Please let me know if there are any other changes/features to be implemented.

matmerr:
podname should already be a part of telemetry properties, so it should be all good

https://github.com/microsoft/retina/blob/8963dd561d4d5b52c1f711dcbc13fe9f616bd360/pkg/telemetry/telemetry.go#L155

Anirudh2112:
Yes, In the TrackEvent method, t.properties is merged with the provided event properties, so podname will automatically be included in the heartbeat telemetry.

PR #1094, opened 2024-12-02
feat: add kernelVersion to telemetry traces

Anirudh2112:
# Description

This pull request adds the kernelVersion property to telemetry traces in the TrackTrace method. This enhancement allows pinpointing the specific kernel version responsible for producing logs.

# To achieve this:

Introduced a kernelVersionFunc variable to enable easier mocking during tests.
Updated the TrackTrace method to include kernelVersion in telemetry traces.
Added a test (TestTrackTraceIncludesKernelVersion) to verify the inclusion of kernelVersion in trace properties.
Related Issue
This pull request addresses the issue:
[#1071](https://github.com/microsoft/retina/issues/1071)

# Checklist

 I have read the contributing documentation
 I signed and signed-off the commits (git commit -S -s ...). 
 I have tested the changes locally.
 I have followed the project's style guidelines.
 I have added tests.

# Screenshots (if applicable) or Testing Completed

# Test Output

The following test passes successfully:
* go test -run TestTrackTraceIncludesKernelVersion


Anirudh2112:

pkg/telemetry/telemetry.go
```diff
@@ -22,7 +22,7 @@ var (
 	client  appinsights.TelemetryClient
 	version string
 	mbShift uint64 = 20
-
+    kernelVersionFunc   = KernelVersion
```
timraymond, line 25: I think this needs to be run through gofmt.

pkg/telemetry/telemetry.go
```diff
@@ -224,6 +224,14 @@ func (t *TelemetryClient) TrackMetric(metricname string, value float64, properti
 
 func (t *TelemetryClient) TrackTrace(name string, severity contracts.SeverityLevel, properties map[string]string) {
 	trace := appinsights.NewTraceTelemetry(name, severity)
+
+	kernelVersion, err := kernelVersionFunc(context.Background()) 
+	if err == nil {
+		properties[kernelversion] = kernelVersion 
+	} else {
+		t.trackWarning(err, "failed to get kernel version for trace")
+	}
```
timraymond, line 233: Usually preference is to test `err != nil`, can you swap this conditional?

Anirudh2112: Hi, @timraymond I've made the necessary changes and pushed the code. Kindly let me know if anything else needs to be changed/modified. Thank you.

rbtr: @Anirudh2112 I think you need to take another look at that conditional and make sure you updated it fully

pkg/telemetry/telemetry.go
```diff
@@ -224,6 +224,14 @@ func (t *TelemetryClient) TrackMetric(metricname string, value float64, properti
 
 func (t *TelemetryClient) TrackTrace(name string, severity contracts.SeverityLevel, properties map[string]string) {
 	trace := appinsights.NewTraceTelemetry(name, severity)
+
+	kernelVersion, err := kernelVersionFunc(context.Background())
+	if err == nil {
+		t.trackWarning(err, "failed to get kernel version for trace")
```
matmerr, line 230: won't this cause infinite recursion in the error case?

`t.TrackTrace -> t.trackWarning -> t.TrackTrace`

matmerr (changes requested):

Anirudh2112:
Yes, @matmerr that is true, and thank you @rbtr for catching that.  I’ve updated the conditional to check for err != nil to avoid infinite recursion. Please let me know if there are any other concerns.

rbtr:
@Anirudh2112 the infinite recursion is still present, `trackWarning` calls `TrackTrace`

Anirudh2112:
Apologies for the oversight regarding the infinite recursion. In the recent changes, I’ve addressed this by:

1. Replacing the call to t.trackWarning with direct error logging to avoid recursion.
2. Ensuring that errors are logged as separate telemetry traces using appinsights.NewTraceTelemetry.
3. Preserving the original functionality by adding kernelVersion to the properties map when available and maintaining the flow for the main trace.

timraymond:
@Anirudh2112 I had a few suggestions, mostly around the use of the global var to mock the `KernelVersion` func. Two issues with it: 1) it makes the tests racy, which limits the amount of concurrency we can use when running the test suite. Fast tests are nice, and you can only really have fast tests by staying vigilant for things that slow them down :) 2) Depending on the `KernelVersion` func is really the wrong dependency to take--what we want to depend on the behavior of "give me the kernel version from this host." The way to depend on behavior in Go is to depend on an interface.

Looking at the patch closer though, I can see what you had to work with, and a lot of it was pre-existing. I went ahead and made the changes I suggested to save some back-and-forth. The commit is on [traymond/kernelversion-traces](https://github.com/microsoft/retina/tree/traymond/kernelversion-traces). Assuming the typical "triangular workflow," you should be able to fetch it from your `upstream` remote, then `git merge --ff-only` onto your branch to pull the change in.

Anirudh2112:
@timraymond I've merged your changes from the traymond/kernelversion-traces branch. I appreciate the improvements made to handle platform detection at runtime and the cleaner approach using the HostInfoClient. I've reviewed the changes and understand this removes the need for compile-time platform detection while making the code more testable. Please let me know if there's anything else needed.

PR #1098, opened 2024-12-04
feat: let plugins self-register

rbtr:
# Description

Refactors plugin registration such that Plugins self-register into the plugin Registry simply by their package being imported instead of being explicitly registered by external logic. This should look familiar as a Go plugin pattern as it's fundamentally the same as what's done in [Telegraf](https://github.com/influxdata/telegraf/blob/543b907cdf46d555afe083de846dc4ed055a4339/plugins/outputs/syslog/syslog.go#L191), [CoreDNS](https://github.com/coredns/coredns/blob/7429380b1cacebde73084ad3cb6fb676c69bd6b8/core/plugin/zplugin.go) (Caddy), and others. 

- Abuses `init()`s and aliasing to hide anything deeper than `pkg/plugin` from the rest of Retina, so that Plugins are more encapsulated. 
- Privatize the `Name` const in every Plugin since it no longer needs to be accessible from outside of their package. 
- Remove the `api.Name` type from the Plugin spec since it is redundant.
    Creating a string type like that is useful for enums, when you want to ensure at a data boundary that the string you're ingesting is one of your allowed values. Plugin names are self-provided consts and not used like this anyway, so it could only cause confusion. Removing it means we need to do many less type coercions. This also means we can pass around the plugin list from the conf (a []string) directly.

## Related Issue

I was investigating making it possible for a Plugin to auto-detect that it should be enabled (via some auto-enable registry of detectFuncs, or similar) and wanted to clean this up first so to simplify that implementation.


rbtr:
> @rbtr it looks like we need to delete https://github.com/microsoft/retina/tree/main/pkg/plugin/api if we apply this change?

there was a dangling `//go:generate` comment which referenced the old path

pkg/plugin/dns/dns_linux.go
```diff
@@ -17,21 +17,25 @@ import (
...
-	"github.com/microsoft/retina/pkg/plugin/api"
 	"github.com/microsoft/retina/pkg/plugin/common"
+	"github.com/microsoft/retina/pkg/plugin/registry"
 	"github.com/microsoft/retina/pkg/utils"
 	"go.uber.org/zap"
 )
 
-func New(cfg *kcfg.Config) api.Plugin {
+func init() {
+	registry.Plugins[name] = New
```
timraymond, line 27: I might make this a method of the registry, just so that you can loudly panic if there's a name collision. Unlikely to be an issue in the near term, but you never know.

pkg/plugin/include_linux.go
```diff
@@ -0,0 +1,15 @@
+// nolint // don't complain about this file
```
timraymond, line 1: lol

pkg/plugin/include_windows.go
```diff
@@ -0,0 +1,8 @@
+// nolint // don't complain about this file
+package plugin
+
+// Plugins self-register via their init() funcs as long as they are imported.
+import (
+	_ "github.com/microsoft/retina/pkg/plugin/hnsstats"
+	_ "github.com/microsoft/retina/pkg/plugin/pktmon"
```
timraymond, line 7: Nicely done--it's obvious which are windows plugins only :)

timraymond (approved):

pkg/plugin/dns/dns_linux.go
rbtr: Good idea, don't want to let a new plugin clobber an existing registration.
I had started to do it [here](https://github.com/microsoft/retina/blob/7e2d7130c03779a3334f6b693a37e0d2205b3a3e/pkg/plugin/registry/registry.go#L42-L44), let me finish hiding the Registry from direct txs

rbtr: updated

timraymond (approved):

SRodi (approved):

PR #1119, opened 2024-12-10
fix: Add error checking for HNS operations

timraymond:
The error checking introduced in adcba81f did not adequately check errors returned from underlying HNS operations before wrapping the errors. This results in empty, yet still non-nil errors being returned from these functions in all cases. Consequently, the callers' tests of success (if err != nil) become tautologies, and fail in all cases, thus causing the HNS plugin to continually crash.

### DIFF
--- a/pkg/plugin/hnsstats/vfp_counters_windows.go
+++ b/pkg/plugin/hnsstats/vfp_counters_windows.go
@@ -152,15 +152,21 @@ func getVfpPortCountersRaw(portGUID string) (string, error) {
 
 	cmd := exec.Command("cmd", "/c", vfpCmd)
 	out, err := cmd.Output()
+	if err != nil {
+		return "", errors.Wrap(err, "errored while running vfpctrl /get-port-counter")
+	}
+	return string(out), nil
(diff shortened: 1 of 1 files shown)


huntergregory (approved):

pkg/plugin/hnsstats/vfp_counters_windows.go
```diff
@@ -152,15 +152,21 @@ func getVfpPortCountersRaw(portGUID string) (string, error) {
 
 	cmd := exec.Command("cmd", "/c", vfpCmd)
 	out, err := cmd.Output()
+	if err != nil {
+		return "", errors.Wrap(err, "errored while running vfpctrl /get-port-counter")
```
MikeZappa87, line 156: an error occurred, produced an error, ..... errored comes off as technical slang. I treat this like "til"

timraymond:
This is likely a red herring for a root cause, as it was pointed out by @rbtr that `errors.Wrap` performs a nil err check and re-returns a nil err in that case. Closing.

Issue #1045, opened 2024-11-20
Some of the Retina capture unit tests are flaky


nddq:
![Image](https://github.com/user-attachments/assets/0408eb04-a4e9-4c21-ae72-8b89acf005fc)
Need to investigate why

kamilprz:
Hey @nddq , did you run the `make test` under any specific conditions? I have ran it around 10 times from my end and each time I get all tests passing. Unable to reproduce the failures.

![Image](https://github.com/user-attachments/assets/ecdd1715-52cc-42bd-bda9-fb253054b58c)

nddq:
hey @kamilprz, you can view pass test runs on GitHub Actions. I have not tried to reproduce this locally yet since this seems to only occured in our pipelines

kamilprz:
Example of failing tests - https://github.com/microsoft/retina/actions/runs/12469454230/job/34802618326

PR #1173, opened 2024-12-31
fix(test): Add custom comparer to ignore timestamps in capture tests

kamilprz:
# Description

Previously, in [PR 894](https://github.com/microsoft/retina/pull/894), timestamps were introduced to be able to print capture names as their jobs were beginning. 

This PR addresses an oversight in the testing which compared whether the timestamp of the test and the timestamp of the test capture were the same. This was not always the case - particularly if there was some delay between the beginning of the test and the beginning of the capture. The change in this PR adds a custom comparer to the `cmp.Diff` which ignores the timestamp environment variables.

## Related Issue

[1045](https://github.com/microsoft/retina/issues/1045)

## Checklist

## Testing Completed

The tests now succeed regardless of whether a delay occurs or not - causing the timestamps to differ. This was simulated with an artificial delay with `time.Sleep`.

---

### DIFF
--- a/pkg/capture/crd_to_job_test.go
+++ b/pkg/capture/crd_to_job_test.go
@@ -820,6 +820,10 @@ func Test_CaptureToPodTranslator_ValidateCapture(t *testing.T) {
 	}
 }
 
+func isIgnorableEnvVar(envVar corev1.EnvVar) bool {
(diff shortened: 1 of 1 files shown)


ibezrukavyi (changes requested):
Could you please add separate check that timestamp is not empty and valid.

ibezrukavyi (approved):

PR #1210, opened 2025-01-10
fix: Remove gomnd from golangci linters

timraymond:
This linter causes more harm than good, encouraging practices that actively obfuscate simple values whose meaning could be derived from context. Since it has no context-awareness, it flags everything from file modes, to obvious iteration counters as "magic numbers." This determination is fairly straightforward for a reviewer to make by hand, and it comes up relatively infrequently. The best course of action, until a better (probably LLM-based) magic number detector can be found.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -7,12 +7,10 @@ issues:
     - path: pkg/metrics/types_windows.go
       linters:
         - revive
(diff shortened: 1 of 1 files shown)


rbtr (approved):
llmnd? 😅

PR #1235, opened 2025-01-16
feat: add metricscardinality to heartbeat

alexcastilio:
# Description

Add `metricscardinality` to heartbeat. This will give visibility on the number of time series being exposed by retina.

## Related Issue

#1040 

## Checklist

Metrics exported with heartbeat:
image

## Additional Notes

Metrics of types `histogram` and `summary` expose multiple time series during a scrape. Code is counting according to number of time series exposed at /metrics endpoint.

Ref: 
https://prometheus.io/docs/concepts/metric_types/#histogram
https://prometheus.io/docs/concepts/metric_types/#summary

---

### DIFF
--- a/pkg/telemetry/telemetry.go
+++ b/pkg/telemetry/telemetry.go
@@ -10,12 +10,15 @@ import (
 	"os"
 	"runtime"
 	"runtime/debug"
+	"strconv"
 	"sync"
 	"time"
 
 	"github.com/microsoft/ApplicationInsights-Go/appinsights"
 	"github.com/microsoft/ApplicationInsights-Go/appinsights/contracts"
+	"github.com/microsoft/retina/pkg/exporter"
 	"github.com/microsoft/retina/pkg/log"
+	io_prometheus_client "github.com/prometheus/client_model/go"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


matmerr:
I think a simpler solution may just be to enumerate over the metrics and take the length of the metrics slice

https://gist.github.com/matmerr/f2e31d9be0b5af0eec750fa944edc617#file-counter_test-go-L58

---

```go
// count all metrics and combination of labels currently exposed on
// prometheus endpoint
metrics, err := registry.Gather()
if err != nil {
	fmt.Println("Error gathering metrics: ", err)
}

// count all metrics exposed for each metric
var metriccounter int
for _, metric := range metrics {
	metriccounter += len(metric.Metric)
}
```

---

where the metrics endpoint looks like this:

image

and in stdout we see the same, total metrics is 4
image

alexcastilio:
@matmerr if I only loop through the metrics and add the length of each, this will cause a difference between the count seen through scraping and by this code. Metrics of type histogram and summary expose more time series for quantile/bucket, count and sum. Is it ok to exist this difference in the counting between what's seen in the `/metrics` endpoint and by summing the length of metrics in code?

anubhabMajumdar:
@alexcastilio How much is the difference? Is it 2x or a constant number?

matmerr:
The issue like @alexcastilio showed me is that histogram's create more metrics that can't be summed by just length, and our apiserver metrics latency uses histogram type.

https://github.com/microsoft/retina/blob/23c34b8068947b38af6d34b4f737945785434d2b/pkg/exporter/prometheusexporter.go#L79

Confirmed with him offline that this diff outputs the correct results. Sample endpoint, including metric of histogram type:
image

Results of his diff:
image

matmerr (approved):

matmerr (approved):

pkg/telemetry/telemetry.go
```diff
@@ -174,10 +177,55 @@ func (t *TelemetryClient) heartbeat(ctx context.Context) {
...
+	}
+
+	props["metricscardinality"] = strconv.Itoa(metricscardinality)
+
 	maps.Copy(props, cpuProps)
 	maps.Copy(props, t.profile.GetMemoryUsage())
 	t.TrackEvent("heartbeat", props)
 }
+func metricsCardinality() (int, error) {
+	metricFamilies, err := exporter.CombinedGatherer.Gather()
```
anubhabMajumdar, line 193: `CombinedGatherer` can be nil.

pkg/telemetry/telemetry.go
```diff
@@ -174,10 +177,55 @@ func (t *TelemetryClient) heartbeat(ctx context.Context) {
...
+
+	metricscardinality := 0
+
+	for _, mf := range metricFamilies {
+		switch mf.GetType() { //nolint:exhaustive // 'default' satisfies exhaustiveness
+
+		case io_prometheus_client.MetricType_HISTOGRAM:
+			metrics := mf.GetMetric()
+			for _, m := range metrics {
+				metricscardinality += len(m.GetHistogram().GetBucket()) + 3 // +3 for le="+Inf", _sum and _count
```
anubhabMajumdar, line 206: `GetHistogram` can return nil.

pkg/telemetry/telemetry.go
alexcastilio: @anubhabMajumdar addressed in #1293

pkg/telemetry/telemetry.go
alexcastilio: @anubhabMajumdar addressed in #1293

PR #1248, opened 2025-01-20
test: automate scale test execution

alexcastilio:
# Description

Create automation workflow to run scale test automatically. Initially to keep it simpler, it will be executed daily with latest commit from main. In future it can be changed to run at merge if desired.
Current Scale test configuration is 1000 deployments of 20 replicas of kapinger pods running on a cluster of 1000 nodes. The cluster will be created and deleted at each automatic execution.

Changes in this PR:
* Creation of workflow for daily execution
* Add code to create Scale Test cluster
* Simplify environment variables handling by doing it in `params` package 
* Update kapinger version
* Update kapinger memory limit
* Update context handling in scale test steps (improve flakiness of test)
* Notify test job's root ctx of os signals (SIGINT and SIGTERM) to allow proper cleanup in case of interruption

## Related Issue

## Checklist

Workflow execution was tested by triggering it on push in PR branch. Workflow: https://github.com/microsoft/retina/actions/workflows/daily-scale-test.yaml
image

## Additional Notes

---

### DIFF
--- a/test/e2e/framework/kubernetes/check-pod-status.go
+++ b/test/e2e/framework/kubernetes/check-pod-status.go
@@ -14,14 +14,14 @@ import (
 )
 
 const (
-	RetryTimeoutPodsReady     = 5 * time.Minute
-	RetryIntervalPodsReady    = 5 * time.Second
-	timeoutWaitForPodsSeconds = 1200
+	RetryTimeoutPodsReady  = 5 * time.Minute
+	RetryIntervalPodsReady = 5 * time.Second
 
 	printInterval = 5 // print to stdout every 5 iterations
 )
 
 type WaitPodsReady struct {
+	Ctx                context.Context
 	KubeConfigFilePath string
 	Namespace          string
 	LabelSelector      string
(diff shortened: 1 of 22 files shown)


test/e2e/framework/kubernetes/check-pod-status.go
```diff
@@ -14,14 +14,14 @@ import (
...
-	RetryIntervalPodsReady    = 5 * time.Second
-	timeoutWaitForPodsSeconds = 1200
+	RetryTimeoutPodsReady  = 5 * time.Minute
+	RetryIntervalPodsReady = 5 * time.Second
 
 	printInterval = 5 // print to stdout every 5 iterations
 )
 
 type WaitPodsReady struct {
+	Ctx                context.Context
```
matmerr, line 24: we shouldn't be storing context here, this doc is a pretty great reason why

https://go.dev/blog/context-and-structs#prefer-contexts-passed-as-arguments

alexcastilio: I agree, but the problem is that current framework expects a step to implement Run(), instead of Run(ctx). The alternative I see is to refactor the framework to pass ctx as arguments, but this is a big job that will require many changes on what is already implemented through tests. What do you think?

timraymond: I agree with @matmerr . This always seems like a good idea to avoid injecting context as function arguments, but it is a false economy. You quickly become unable to reason about the lifetime of these contexts anymore and it becomes difficult to refactor it to be correct. I would rather no contexts if we cannot inject them as function arguments.

alexcastilio: Since refactoring framework is not in scope for the current task, I have updated code to use `context.TODO()` when context is needed

matmerr (approved):

Issue #1236, opened 2025-01-17
Potential memory leak in packetparser's interfaceLockMap


nddq:
Currently, `packetparser` generates an entry in the `interfaceLockMap` for each new interface that comes up. However, it fails to remove these entries when the interfaces go down. In environments with high pod counts and frequent churn, this can cause the map to grow indefinitely, resulting in a memory leak.

```go
        ifaceKey := ifaceToKey(iface)
	lockMapVal, _ := p.interfaceLockMap.LoadOrStore(ifaceKey, &sync.Mutex{})
	mu := lockMapVal.(*sync.Mutex)
	mu.Lock()
	defer mu.Unlock()

	switch event.Type {
	case endpoint.EndpointCreated:
		p.l.Debug("Endpoint created", zap.String("name", iface.Name))
		p.createQdiscAndAttach(iface, Veth)
	case endpoint.EndpointDeleted:
		p.l.Debug("Endpoint deleted", zap.String("name", iface.Name))
		// Clean.
		if value, ok := p.tcMap.Load(ifaceKey); ok {
			v := value.(*tcValue)
			p.clean(v.tc, v.qdisc)
			// Delete from map.
			p.tcMap.Delete(ifaceKey)
		}
	default:
		// Unknown.
		p.l.Debug("Unknown event", zap.String("type", event.Type.String()))
	}
```

PR #1249, opened 2025-01-21
fix: Clean up interfaceLockMap entries on endpoint deletion

byte-msft:
# Description

The packetParser was creating entries in interfaceLockMap for each new interface 
but failing to remove them when interfaces were deleted. In environments with 
high pod counts and frequent churn, this caused a memory leak as the map grew 
indefinitely.

## Related Issue

[Potential memory leak in packetparser's interfaceLockMap #1236](https://github.com/microsoft/retina/issues/1236)

## Checklist

- [X] I have read the [contributing documentation](https://retina.sh/docs/contributing).

### Solution
- Added cleanup of interfaceLockMap entries in the EndpointDeleted case
- Improved mutex handling logic to prevent resource leaks
- Updated test cases to verify proper cleanup of both tcMap and interfaceLockMap

### Testing
- Added comprehensive test coverage for interface deletion scenario
- Verified cleanup of both maps in test cases
- Tested with high pod churn scenarios

### Impact
This fix prevents memory leaks in environments with frequent pod creation/deletion,
improving the overall stability and resource usage of the system.
---


SRodi (commented):
LGTM but I'll leave @nddq or @rbtr to review & approve

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -380,26 +380,38 @@ func (p *packetParser) endpointWatcherCallbackFn(obj interface{}) {
...
+		lockMapVal, _ := p.interfaceLockMap.LoadOrStore(ifaceKey, &sync.Mutex{})
+		mu := lockMapVal.(*sync.Mutex)
+		mu.Lock()
+		defer mu.Unlock()
+
 		p.l.Debug("Endpoint created", zap.String("name", iface.Name))
 		p.createQdiscAndAttach(iface, Veth)
 	case endpoint.EndpointDeleted:
+		// Get the mutex only if it exists
+		lockMapVal, exists := p.interfaceLockMap.Load(ifaceKey)
```
kamilprz, line 397: Bit of a nitpick / question since I'm still new to Go. Is it recommended to stick with the `ok` idiom? Or is it fine to use other variable names like `exists`?

byte-msft: Hey! Yeah, I initially used `exists` to make the code more readable, but you're right - we should stick with `ok` to follow Go conventions in the main code.

For the test case though, I deliberately kept `tcMapExists` and `lockMapExists` because test code is a bit different - clarity is super important there since we're verifying specific behaviors. The more explicit naming makes it immediately obvious what we're testing for, especially when someone's debugging failed tests.

Let me know if you'd like me to update the main code to use `ok` instead of `exists`!

nddq:
we should take this chance to examine whether or not do we even need these `sync.Mutex` for each interface, since we are processing them sequentially and not concurrently anyway

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -380,26 +380,38 @@ func (p *packetParser) endpointWatcherCallbackFn(obj interface{}) {
...
 	ifaceKey := ifaceToKey(iface)
-	lockMapVal, _ := p.interfaceLockMap.LoadOrStore(ifaceKey, &sync.Mutex{})
-	mu := lockMapVal.(*sync.Mutex)
-	mu.Lock()
-	defer mu.Unlock()
 
 	switch event.Type {
 	case endpoint.EndpointCreated:
+		// Create mutex only when needed
+		lockMapVal, _ := p.interfaceLockMap.LoadOrStore(ifaceKey, &sync.Mutex{})
```
MikeZappa87, line 388: I am curious of the need here. This seems a little bit complex and could a simpler approach be used?

byte-msft: Sure, thanks 

I've removed the Mutex mechanism since we're now using a sequential approach for adding and removing interfaces

byte-msft:
> we should take this chance to examine whether or not do we even need these `sync.Mutex` for each interface, since we are processing them sequentially and not concurrently anyway


Yeah, you make a valid point! Since we're processing interfaces sequentially, we can safely remove the mutex mechanism. Let's keep it simple and avoid unnecessary complexity. I've updated the PR to remove the mutex-related code.

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -380,24 +379,18 @@ func (p *packetParser) endpointWatcherCallbackFn(obj interface{}) {
 	}
 
 	iface := event.Obj.(netlink.LinkAttrs)
-
 	ifaceKey := ifaceToKey(iface)
-	lockMapVal, _ := p.interfaceLockMap.LoadOrStore(ifaceKey, &sync.Mutex{})
```
anubhabMajumdar, line 385: Why are you deleting this? The `interfaceLockMap` allows us to store a per interface lock and we can create/delete multiple qdisc in parallel. This is necessary (in place of a single lock) because large number of pods can come up at the same time, and we should start capturing packets as quickly as possible.

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -380,24 +379,18 @@ func (p *packetParser) endpointWatcherCallbackFn(obj interface{}) {
...
 	case endpoint.EndpointDeleted:
 		p.l.Debug("Endpoint deleted", zap.String("name", iface.Name))
-		// Clean.
+		// Clean tcMap.
 		if value, ok := p.tcMap.Load(ifaceKey); ok {
 			v := value.(*tcValue)
 			p.clean(v.tc, v.qdisc)
-			// Delete from map.
 			p.tcMap.Delete(ifaceKey)
 		}
```
anubhabMajumdar, line 395: We need to delete the `ifacekey` from `interfaceLockMap` if it's deleted.

anubhabMajumdar (changes requested):

pkg/plugin/packetparser/packetparser_linux.go
byte-msft: Reverted my latest changes, so I brought back the interfaceLockMap  and removing the ifacekey from the map

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -373,36 +373,44 @@ func (p *packetParser) clean(rtnl nltc, qdisc *tc.Object) {
...
-	mu := lockMapVal.(*sync.Mutex)
-	mu.Lock()
-	defer mu.Unlock()
 
 	switch event.Type {
 	case endpoint.EndpointCreated:
-		p.l.Debug("Endpoint created", zap.String("name", iface.Name))
-		p.createQdiscAndAttach(iface, Veth)
+		// Get or create mutex atomically
+		lockMapVal, loaded := p.interfaceLockMap.LoadOrStore(ifaceKey, &sync.Mutex{})
```
anubhabMajumdar, line 387: Why bring this inside the switch case? We are duplicating code in L399-L404.

byte-msft: Hey Anubhab,

Yeah, I get the concern about code duplication, but in this case I think it makes sense to keep the mutex logic separate in each case. The create and delete paths need different mutex handling - creation needs to make a new mutex if it doesn't exist, while deletion only works with existing ones. So, I added an extra check. 

Trying to handle it outside the switch would probably make the code more generalized and less clear, in addition we are locking the map if the other cases will be added. 

What do you think? Happy to revert the code to the previous state, if you think if it's unnecessary

pkg/plugin/packetparser/packetparser_linux.go
kamilprz: Sure yeah, let's stick with `ok` for the main code then :)

timraymond: Yeah, `ok` is still best in this circumstance. It's widely-understood what that means.

byte-msft: Sure, it's resolved

pkg/plugin/packetparser/packetparser_linux.go
nddq: I'm still not clear on this. Why is a per-interface mutex necessary for packetparser to handle concurrent create/delete qdisc operations? As far as I understand, operations on different interfaces shouldn't cause a data race.

pkg/plugin/packetparser/packetparser_linux.go
byte-msft: Hi Anubhab, 

I've reverted the lock map optimisations and just deleting the record from the map now

nddq (approved):
approved, although I'm still not thoroughly convinced that we need the lock map, I'll not die on this hill for this PR, will let @anubhabMajumdar sign this off

anubhabMajumdar (approved):

anubhabMajumdar:
> approved, although I'm still not thoroughly convinced that we need the lock map, I'll not die on this hill for this PR, will let @anubhabMajumdar sign this off

Had a conversation offline to resolve this. Change looks good to me.

Issue #1190, opened 2025-01-06
[conntrack] Legacy control plane metrics implementation


SRodi:
Implement at the user-space level conntrack Prometheus metrics for legacy control plane 

As per #806:
> These can be useful as Prometheus's metrics:
> 
> - Packets and bytes count per connection.
> - Number of connections going to world.
> - Connections with unknown direction.

PR #1253, opened 2025-01-21
feat(conntrack-metrics): legacy control plane basic mode

SRodi:
# Description

Initial implementation of conntrack metrics for the legacy control plane in basic mode. The aggregation is at the node level, which makes the metric bounded.

Metrics:
- conntrack_bytes_tx
- conntrack_bytes_rx
- conntrack_packets_tx
- conntrack_packets_rx
- conntrack_total_connections

## Related Issue

fixes https://github.com/microsoft/retina/issues/1190

## Checklist

Bytes RX

image

Total connections metric

image

## Additional Notes

---

### DIFF
--- a/pkg/metrics/types.go
+++ b/pkg/metrics/types.go
@@ -45,10 +45,10 @@ const (
 	lostEventsCounterDescription                     = "Number of events lost in control plane"
 
 	// Conntrack metrics
-	ConntrackPacketForwardDescription    = "Number of forward packets"
-	ConntrackPacketReplyDescription      = "Number of reply packets"
-	ConntrackBytesForwardDescription     = "Number of forward bytes"
-	ConntrackBytesReplyDescription       = "Number of reply bytes"
+	ConntrackPacketTxDescription         = "Number of tx packets"
+	ConntrackPacketRxDescription         = "Number of rx packets"
+	ConntrackBytesTxDescription          = "Number of tx bytes"
+	ConntrackBytesRxDescription          = "Number of rx bytes"
 	ConntrackTotalConnectionsDescription = "Total number of connections"
 )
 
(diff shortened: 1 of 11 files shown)


pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -637,6 +637,19 @@ func (p *packetParser) processRecord(ctx context.Context, id int) {
 				p.enricher.Write(ev)
 			}
 
+			// Add conntrack metrics.
```
nddq, line 640: this doesn't look right to me. If there are multiple connections of the same type and direction, when packetparser receives each packet event from each connection, would this just "reset" the packet and byte metrics for this node to whatever the counters of the connection the packet belongs to? Let's say that I have connection A and B, and packetparser is receiving packets from A for every other packet from B, the node metrics would flip flop

nddq (changes requested):

pkg/plugin/packetparser/packetparser_linux.go
SRodi: Thanks @nddq ! I'll look into implementing this inside conntrack GC loop to increment connection-level counters based on BPF conntrack metrics. Then we can set the metrics values as node-level (basic) and pod-level (advanced) metrics. I'll push another commit soon.

pkg/metrics/types.go
```diff
@@ -43,6 +43,13 @@ const (
 	// Control plane metrics
 	pluginManagerFailedToReconcileCounterDescription = "Number of times the plugin manager failed to reconcile the plugins"
 	lostEventsCounterDescription                     = "Number of events lost in control plane"
+
+	// Conntrack metrics
+	ConntrackPacketForwardDescription    = "Number of forward packets"
```
MikeZappa87, line 48: forward or forwarded?

SRodi: I think we should use ‘forward’ since in this context we refer to ‘direction’ more than ‘action’. We also want to be consistent with the actual metrics implementation in BPF, which uses 'forward' and 'reply'. https://github.com/microsoft/retina/blob/f5fb767df6a2076222f02112ea5585ae2f699613/pkg/plugin/conntrack/_cprog/conntrack.c#L27

timraymond: Might be better, then, to say 'Number of packets marked "forward"'.

VincentS:
@SRodi Any update on this PR ?

pkg/metrics/types.go
SRodi: I have updated this to `tx` and `rx` to avoid any confusion and remain consistent with the `ct_entry` naming convention.

nddq (approved):

ibezrukavyi (approved):

PR #1266, opened 2025-01-24
chore: add APP_INSIGHTS_ID to image build

alexcastilio:
# Description

* Add APP_INSIGHTS_ID to image build to allow monitoring through heartbeat during test execution.
* Add test to configuration packages (for both operator and daemonset)
* Make heartbeat interval configurable
  * New keys in `deploy/standard/manifests/controller/helm/retina/values.yaml`

## Related Issue

## Checklist

Manual test was also done to validate that the configuration was propagated from helm chart `values.yaml` file to both operator and agent. Value set was printed to log:
image
image

## Additional Notes

---

### DIFF
--- a/operator/config/config.go
+++ b/operator/config/config.go
@@ -2,9 +2,18 @@ package config
 
 import (
 	"fmt"
+	"log"
+	"time"
 
 	"github.com/microsoft/retina/pkg/config"
 	"github.com/spf13/viper"
+)
+
+const MinTelemetryInterval time.Duration = 2 * time.Minute
+
+var (
+	DefaultTelemetryInterval       = 5 * time.Minute
+	ErrorTelemetryIntervalTooSmall = fmt.Errorf("telemetryInterval smaller than %v is not allowed", MinTelemetryInterval)
 )
 
 type OperatorConfig struct {
@@ -14,8 +23,9 @@ type OperatorConfig struct {
 	EnableTelemetry bool   `yaml:"enableTelemetry"`
 	LogLevel        string `yaml:"logLevel"`
 	// EnableRetinaEndpoint indicates whether to enable RetinaEndpoint
-	EnableRetinaEndpoint bool `yaml:"enableRetinaEndpoint"`
-	RemoteContext        bool `yaml:"remoteContext"`
+	EnableRetinaEndpoint bool          `yaml:"enableRetinaEndpoint"`
+	RemoteContext        bool          `yaml:"remoteContext"`
+	TelemetryInterval    time.Duration `yaml:"telemetryInterval"`
 }
 
 func GetConfig(cfgFileName string) (*OperatorConfig, error) {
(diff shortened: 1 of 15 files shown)


operator/config/config.go
```diff
@@ -35,5 +37,10 @@ func GetConfig(cfgFileName string) (*OperatorConfig, error) {
 		return nil, fmt.Errorf("error unmarshalling config: %w", err)
 	}
 
+	// If unset, default telemetry interval to 5 minutes.
+	if cfg.TelemetryInterval == 0 {
+		cfg.TelemetryInterval = 5 * time.Minute
```
matmerr, line 42: let's keep to 15min, and since this is a configmap change we'll need to notify a few people

https://github.com/microsoft/retina/pull/1266/files#diff-754f5ed2c6297aa0f46f6627521cda46a66d540bcdc9713857a6a84f7fc7358aL52

nddq: time to add CODEOWNERS?

rbtr: I think it would be better if a zero value `TelemetryInterval` means no telemetry/the previous behavior. That keeps any changed behavior intentionally opt-in, backcompat, and obvious

alexcastilio: @matmerr currently it is 15 min for the daemonset heartbeat. This config is for operator, which is 5 min. No change in default values:
https://github.com/microsoft/retina/pull/1266/files#diff-0737b8b43b538114b7d2a8728798d2d1e5648063201b8584d6234491f6467d96R30

@rbtr currently a different key in the yaml will enable telemetry:
https://github.com/microsoft/retina/blob/743413d0bafb4b025e22d0c164e6f2637f4b9038/deploy/standard/manifests/controller/helm/retina/values.yaml#L69

Do you think `enableTelemetry` key should be removed and keep only the `telemetryInterval`? (disable telemetry if time is set to 0)

matmerr: @huntergregory is there a specific reason operator heartbeat is so frequent at 5min?

matmerr: > I think it would be better if a zero value TelemetryInterval means no telemetry/the previous behavior. That keeps any changed behavior intentionally opt-in, backcompat, and obvious

@rbtr in this specific case I think reading `telemetryInterval` as 0 may require a user to jump to docs. I could see it being interpreted as disabled and don't send telemetry at all, or send telemetry in a continuous stream with a 0 second delay. I kinda prefer keeping the whole telemetry functionality behind the explicit `enableTelemetry` bool

rbtr: if there's a separate enable flag that bypasses this entirely, on objection. let's just make sure that if this is unset it defaults to the previous behavior

pkg/config/testwith/config.yaml
```diff
@@ -9,3 +9,4 @@ metricsIntervalDuration: "10s"
 # used to export telemetry to AppInsights
 telemetryEnabled: true
 dataAggregationLevel: "low"
+telemetryInterval: "15m"
```
matmerr, line 12: we can be careful, but is it possible to add a test around the behavior if someone just uses `15` instead of `15m`?

matmerr (approved):

pkg/config/testwith/config.yaml
alexcastilio: I added check comparing with minimum value allowed (now 2m). If it's less then the minimum, an error will be triggered. If value is zero or unset, it will be set to default to keep previous behavior

ibezrukavyi (approved):

kamilprz (approved):

PR #1283, opened 2025-01-29
fix: kapinger use service IP instead of name

matmerr:
# Description

CoreDNS wasn't playing nice with massive bursts of service name fqdn lookups, this moves to using the ip of the service(s) directly

## Related Issue

## Checklist

## Additional Notes

---

### DIFF
--- a/hack/tools/kapinger/clients/http.go
+++ b/hack/tools/kapinger/clients/http.go
@@ -82,12 +82,12 @@ func (k *KapingerHTTPClient) MakeRequests(ctx context.Context) error {
 	ticker := time.NewTicker(k.interval)
 	sem := make(chan struct{}, 5) // semaphore to limit the number of active goroutines
 
-	serviceController, err := NewServiceLoggingController(k.clientset, k.labelselector)
+	serviceController, err := newServiceController(k.clientset, k.labelselector)
 	if err != nil {
 		return fmt.Errorf("error creating service controller: %w", err)
 	}
 
-	go serviceController.Run(ctx)
+	go serviceController.run(ctx)
 
 	for {
 		select {
--- a/hack/tools/kapinger/clients/http.go
+++ b/hack/tools/kapinger/clients/http.go
@@ -41,7 +41,7 @@ func NewKapingerHTTPClient(clientset *kubernetes.Clientset, labelselector string
 			Transport: &http.Transport{
 				DisableKeepAlives: true,
 			},
-			Timeout: 3 * time.Second,
+			Timeout: 15 * time.Second,
 		},
 		labelselector: labelselector,
 		clientset:     clientset,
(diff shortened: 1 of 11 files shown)


hack/tools/kapinger/clients/http_service_controller.go
```diff
@@ -0,0 +1,111 @@
...
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+)
+
+type serviceController struct {
+	sync.RWMutex
+	serviceInformer cache.SharedIndexInformer
+
+	ips []string
```
SRodi, line 24: Although a slice is more memory-efficient, if the order of IPs is not important, have you considered using a map instead? The lookup time would be much faster, especially with large number of IPs. Using a map would also simplify the code for removeIP func.

MikeZappa87:
What was the issue with CoreDNS? Have we informed the correct teams of a possible issue with CoreDNS?

hack/tools/kapinger/clients/http_service_controller.go
MikeZappa87, line 24: Thoughts on using net.IP instead of string? What stops me from putting 'wingding' as a value?

matmerr: no complaints, updated

but also if service.Spec.ClusterIP changes from string type, I think the world might stop :)

matmerr: The way we're currently using this adding service IP's happens once, and/or relatively infrequently, whereas pulling arbitrary IP's is the hot path. With a slice we can easily pull a random IP via direct index, or iterate over a subslice of IP's with very little overhead

hack/tools/kapinger/servers/http.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	_ "net/http/pprof"
```
huntergregory, line 8: what does this do? Comment for future users?

hack/tools/kapinger/go.mod
```diff
@@ -3,10 +3,12 @@ module github.com/microsoft/retina/hack/tools/kapinger
 go 1.22.5
 
 require (
+	golang.org/x/exp v0.0.0-20250128182459-e0ece0dbea4c
```
huntergregory, line 6: non-blocking: I'm not a fan of the experimental library because it makes breaking changes all the time

hack/tools/kapinger/clients/http.go
```diff
@@ -41,7 +41,7 @@ func NewKapingerHTTPClient(clientset *kubernetes.Clientset, labelselector string
 			Transport: &http.Transport{
 				DisableKeepAlives: true,
 			},
-			Timeout: 3 * time.Second,
+			Timeout: 15 * time.Second,
```
huntergregory, line 44: What if there's a networkpolicy dropping traffic in our testing? This would reduce that amount of traffic by 80%?

huntergregory: perhaps make this timeout configurable?

hack/tools/kapinger/clients/http.go
```diff
@@ -80,22 +80,41 @@ func NewKapingerHTTPClient(clientset *kubernetes.Clientset, labelselector string
...
+			if len(sem) >= cap(sem) {
+				log.Printf("max http clients reached, skipping interval")
+				continue
+			}
 			go func() {
+				sem <- struct{}{}
+				defer func() { <-sem }()
+
 				for i := 0; i < k.volume; i++ {
-					for _, url := range k.urls {
```
huntergregory, line 91: Are we killing the `TARGET_TYPE=pod` functionality? I'd remove any dead code or add comments to it to fix later
https://github.com/microsoft/retina/blob/d1378d10c3bc20599475cd505ce3b970ccc4160c/hack/tools/kapinger/clients/http.go#L61-L73

hack/tools/kapinger/servers/http.go
timraymond: @huntergregory that's fairly common, but also not my most favorite thing. It uses `init()` to attach pprof handlers to the default `net/http` handler. Probably fine here, where it's for testing, but for production pprof handlers I usually like to explicitly attach those handlers to my routes. This is documented in the package docs here: https://pkg.go.dev/runtime/pprof .

Issue #1280, opened 2025-01-28
Continuous error message about interfaces skipped by Linuxutil


anubhabMajumdar:
**Describe the bug**

In a cluster, I am continuously seeing the following error

```bash
level=error caller=linuxutil/ethtool_stats_linux.go:81 msg="Error while getting ethtool:" ifacename=cilium_vxlan error="interface not supported while retrieving stats: operation not supported"
```

This interface should be skipped next time.

**To Reproduce**

1. Deploy a Cilium cluster
2. Install Retina with `make helm-install` 

**Expected behavior**

Error should be reported once.

**Platform (please complete the following information):**

- OS: Linux
- Kubernetes Version: 1.29
- Host: AKS
- Retina Version: v0.0.17

xiaozhiche320:
This issue has been fixed with pr: https://github.com/microsoft/retina/pull/1296

PR #1296, opened 2025-01-31
fix: fix ethtool issue when trying to skip unsupported interface

xiaozhiche320:
# Description

This PR addresses an issue where the plugin repeatedly logged the same operation not supported error for interfaces on every metrics collection cycle. The root cause ethtool handle is created and closed during each interval, which invalidated any cached (LRU) information on unsupported interfaces. Now, once an interface is identified as unsupported, it won’t repeatedly log the same error. 

## Related Issue

https://github.com/microsoft/retina/issues/1280
## Checklist

Logs before fixing the issue:
Before
Logs after fixing the issue, the unsupported cache will only log once now (add debug message for test and validate behaviour of LRU, which has been removed in PR):
After

## Additional Notes

---


pkg/plugin/linuxutil/linuxutil_linux.go
```diff
@@ -61,6 +61,26 @@ func (lu *linuxUtil) SetupChannel(chan *hubblev1.Event) error {
...
+		lu.l.Error("Error while creating ethHandle: %v\n", zap.Error(err))
+		return err
+	}
+
+	ethReader := NewEthtoolReader(ethtoolOpts, ethHandle)
+	if ethReader == nil {
+		lu.l.Error("Error while creating ethReader")
+		return fmt.Errorf("error while creating ethReader")
+	}
+	defer ethHandle.Close()
```
ritwikranjan, line 82: nit: we can register the close just after initializing ethHandle

pkg/plugin/linuxutil/ethtool_stats_linux.go
```diff
@@ -58,7 +58,7 @@ func (er *EthtoolReader) readInterfaceStats() error {
 		return err
 	}
 
-	defer er.ethHandle.Close()
+	// defer er.ethHandle.Close()
```
ritwikranjan, line 61: nit: any particular reason for leaving it commented and not removing it completely? same for the test as well

pkg/plugin/linuxutil/ethtool_handle_linux.go
```diff
@@ -42,9 +43,9 @@ func (ce *CachedEthtool) Stats(intf string) (map[string]uint64, error) {
 	if err != nil {
 		if strings.Contains(err.Error(), "operation not supported") {
 			ce.unsupported.Add(intf, struct{}{})
-			return nil, errors.Wrap(err, "interface not supported while retrieving stats")
+			return nil, fmt.Errorf("interface %q not supported while retrieving stats: %w", intf, err)
```
ritwikranjan: While it is great to have less cluttered error we lose context as well. If we are reducing the presence of error to just one interface, I don't see any harm in having extra information in the event we might have to use this for additional debug.

pkg/plugin/linuxutil/ethtool_stats_linux.go
ritwikranjan: nvm, just saw linters already flagged it!

SRodi (approved):
LGTM, pending @ritwikranjan re-review.

pkg/plugin/linuxutil/linuxutil_linux.go
```diff
@@ -61,6 +62,26 @@ func (lu *linuxUtil) SetupChannel(chan *hubblev1.Event) error {
...
+		lu.l.Error("Error while creating ethHandle: %v\n", zap.Error(err))
+		return fmt.Errorf("failed to create ethHandle: %w", err)
+	}
+	defer ethHandle.Close()
+
+	ethReader := NewEthtoolReader(ethtoolOpts, ethHandle)
+	if ethReader == nil {
+		lu.l.Error("Error while creating ethReader")
+		return errors.New("error while creating ethReader")
+	}
```
anubhabMajumdar, line 83: One of the reason for keeping the `ethHandle` inside the loop is to make sure we are not bringing down Retina agent for transient error. Do you think the `ethHandle` error is permanent - as in if it fails once, it cannot succeed on retry? 
If you return error here, `PluginManager` will treat it as a fatal error.

pkg/plugin/linuxutil/ethtool_stats_linux.go
```diff
@@ -58,8 +58,6 @@ func (er *EthtoolReader) readInterfaceStats() error {
 		return err
 	}
 
-	defer er.ethHandle.Close()
```
anubhabMajumdar, line 61: I don't like closing this here, thanks for refactoring this.

pkg/plugin/linuxutil/linuxutil_linux.go
anubhabMajumdar: Brining this out is probably ok, but I would retry few time (with exponential backoff) before reporting failure.

anubhabMajumdar (changes requested):

pkg/plugin/linuxutil/linuxutil_linux.go
anubhabMajumdar: I think your cache shouldn't be tied to the lifecycle of the tool, it should persist for agent's lifecycle.

pkg/plugin/linuxutil/linuxutil_linux.go
```diff
@@ -11,15 +11,19 @@ import (
...
 	"github.com/microsoft/retina/pkg/log"
 	"github.com/microsoft/retina/pkg/plugin/registry"
+	"github.com/pkg/errors"
 	"github.com/safchain/ethtool"
 	"go.uber.org/zap"
 )
 
 const defaultLimit = 2000
 
+var globalLruCache *lru.Cache[string, struct{}]
```
anubhabMajumdar, line 25: Don't use global variable, inject the cache to the functions. Something like
```
unsupportedInterfacesCache, err = lru.New[string, struct{}](int(defaultLimit))
...
ethReader := NewEthtoolReader(ethtoolOpts, ethHandle, unsupportedInterfacesCache)
...
```

Decide on the cache name if it contains something other than unsupported interfaces.

pkg/plugin/linuxutil/ethtool_stats_linux.go
```diff
@@ -78,6 +76,8 @@ func (er *EthtoolReader) readInterfaceStats() error {
 		if err != nil {
 			if errors.Is(err, errskip) {
 				er.l.Debug("Skipping unsupported interface", zap.String("ifacename", i.Name))
+			} else if strings.Contains(err.Error(), "interface not supported while retrieving stats") {
+				er.l.Warn("Unsupported interface detected:", zap.String("ifacename", i.Name), zap.Error(err))
```
anubhabMajumdar, line 80: Why handle this separately? L82 would print it as well.

pkg/plugin/linuxutil/ethtool_stats_linux.go
```diff
@@ -78,6 +76,8 @@ func (er *EthtoolReader) readInterfaceStats() error {
...
 			} else {
 				er.l.Error("Error while getting ethtool:", zap.String("ifacename", i.Name), zap.Error(err))
```
anubhabMajumdar, line 82: Update the log message. It's `Error while retrieving stats...`.

pkg/plugin/linuxutil/linuxutil_linux.go
```diff
@@ -61,6 +65,15 @@ func (lu *linuxUtil) SetupChannel(chan *hubblev1.Event) error {
 
 func (lu *linuxUtil) run(ctx context.Context) error {
 	lu.l.Info("Running linuxutil plugin...")
+
+	// create a global LRU cache
```
anubhabMajumdar, line 69: Add a more detailed comment about why we are maintaining this.

pkg/plugin/linuxutil/linuxutil_linux.go
```diff
@@ -11,15 +11,19 @@ import (
...
+	lru "github.com/hashicorp/golang-lru/v2"
 	kcfg "github.com/microsoft/retina/pkg/config"
 	"github.com/microsoft/retina/pkg/log"
 	"github.com/microsoft/retina/pkg/plugin/registry"
+	"github.com/pkg/errors"
 	"github.com/safchain/ethtool"
 	"go.uber.org/zap"
 )
 
 const defaultLimit = 2000
```
anubhabMajumdar, line 23: You can relocate this to `types_linux.go` if you want.

anubhabMajumdar (approved):

Issue #1365, opened 2025-02-18
POC - Implement bootstrap manager to split up data plane and control plane


BeegiiK:
**Is your feature request related to a problem? Please describe.**
The data plane (plugins) can only work when the control plane is started. If the host machine is running on a non-k8s container orchestration, nothing will work because the control plane is tightly coupled to kubernetes constructs

**Describe the solution you'd like**
Allow the data plane to start regardless of what the underlying container orchestration is (ACI enablement or kubernetes) and make sure no unexpected side effects are unaccounted for. 

The plugins of interest should be the Windows HNS stats as the POC was decided to use this OS to begin with.

**Describe alternatives you've considered**

**Additional context**

PR #1385, opened 2025-02-24
feat(windows): Create standalone daemon for non-k8s orchestration and retrieve pod level metadata on running VM

BeegiiK:
# Description

This PR is a POC on enabling Retina to work outside of Kubernetes (ACI) and collect pod level information such as name and namespace by running it as a binary on the host VM.

The two main efforts are the following:
* Decoupling the dependence for the K8s configuration files to allow bootstrapping to proceed. Currently, if the control plane fails to boot up, the data plane will also too. In the scenario of running on ACI, we need to reconfigure how Retina starts as it expects a lot of K8s information so that the responsible plugin (HNS stats - feature only supports windows as of now) can start collecting information. This behavior can be controlled by the new toggle, i.e. `enableStandalone` and by default it will be set to false.

* Collecting the pod related information. There are two ways that we can retrieve the pod name and its corresponding namespace for a given (HNS endpoint / IP address). Either through containerd or reading the CNI state file. Appropriate unit testing were also added. This behavior can be controlled by another new toggle, i.e. `enableCrictl`. Once these were collected, the new labels are returned in the new set of advanced metrics for windows (more work is needed to optimize the design as there's still some repetition - encapsulated in this GH issue https://github.com/microsoft/retina/issues/1547

Retina would start as a normal binary with the toggle set to true in the `config.yaml` file. All necessary services start in the control plane such as the HTTP server. The HNS plugin would start and begin polling for HNS endpoints. Retina would then correlate the HNS endpoints to running pods on the host VM and if there is a match, it would be added to cache so it can be retrieved later. If pods are removed then the cache will be updated accordingly. 

TTL parameter was introduced because there is no guarantee that the next set of HNS endpoints will return the same endpoints in the previous iteration. These entries would exist in cache and therefore become stale. The TTL would ensure that the cache will always reflect the active endpoints after the time is up.

## Related Issue

https://github.com/microsoft/retina/issues/1365

## Checklist

### AKS testing to ensure normal Retina works
image
image

#### CriCtl
image
image
image

#### State-file
image
image

Manual deletion in the state-file to reflect cache is updated
image

## ACI Dev Node 
image

## Additional Notes
* Demo explaining this feature to the wider ACN audience - https://microsofteur-my.sharepoint.com/personal/mdegutyte_microsoft_com/_layouts/15/stream.aspx?id=%2Fpersonal%2Fmdegutyte_microsoft_com%2FDocuments%2FRecordings%2FACN%20-%20Standalone%20Retina%20Demo-20250417_170310-Meeting%20Recording%2Emp4&nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbFZpZXciOiJTaGFyZURpYWxvZy1MaW5rIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXcifX0%3D&nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbFZpZXciOiJTaGFyZURpYWxvZy1MaW5rIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXcifX0=&ga=1

---


cmd/bootstrap_manager.go
```diff
@@ -0,0 +1,107 @@
...
+		if err != nil {
+			return fmt.Errorf("creating controller-runtime manager: %w", err)
+		}
+		return b.startDaemon(cfg, daemonConfig, zl)
+	}
+
+	cfg, err = kcfg.GetConfig()
+	if err != nil {
+		panic(err)
+	}
```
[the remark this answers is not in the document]
BeegiiK: Good point. I initially had the implementation without the standalone toggle. I'll put it back into the standard daemon. Thanks Alex!

cmd/standalone_daemon.go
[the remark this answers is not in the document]
BeegiiK: Good point! Addressed in the next commit. Thanks Alex!

pkg/controllers/cache/standalone_cache.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/microsoft/retina/pkg/log"
+	"go.uber.org/zap"
+)
+
+type PodInfo struct {
+	Name      string
+	Namespace string
+}
+
```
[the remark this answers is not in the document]
BeegiiK: Yes, it's still valid. Pod names and Namespaces is not strictly a kubernetes construct and is used in other orchestrations such as Containerd. This work is primarily focused on enabled pod level metrics in standalone mode. Hope this made it clearer. Thanks Alex!

pkg/enricher/enricher.go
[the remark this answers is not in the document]
BeegiiK: Done! Thanks Alex!

pkg/enricher/enricher.go
```diff
@@ -182,6 +182,26 @@ func (e *Enricher) getWorkloads(ownerRefs []*common.OwnerReference) []*flow.Work
...
+			Name:      o.Name(),
+			Namespace: o.Namespace(),
+		}
+
+	default:
+		e.l.Debug("received unknown type from cache", zap.Any("obj", obj), zap.Any("type", reflect.TypeOf(obj)))
+		return nil
+	}
+}
+
```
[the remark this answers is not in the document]
BeegiiK: Since this is not used (as of now), I'll remove it from the PR. Thanks Alex! I agree. There is definitely a better approach to this.

pkg/enricher/standalone_enricher.go
```diff
@@ -0,0 +1,101 @@
...
+					return
+				}
+				e.l.Debug("Processing event", zap.String("ip", event.IP))
+				e.processEvent(event.IP)
+			}
+		}
+	}()
+}
+
+func (e *StandaloneEnricher) processEvent(ip string) {
```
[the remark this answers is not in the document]
BeegiiK: Good point. This method is responsible for obtaining the labels for enrichment. Addressed in the next commit! Thanks Alex!

pkg/enricher/standalone_enricher_test.go
```diff
@@ -0,0 +1,86 @@
...
+	statefile.StateFileLocation = testfile
+	MaxStandaloneCacheEventSize = 1
+	testCache := cache.NewStandaloneCache()
+	enricher := NewStandaloneEnricher(context.Background(), testCache)
+
+	go enricher.Run()
+	defer enricher.Stop()
+
+	err1 := enricher.PublishEvent(ip)
+	err2 := enricher.PublishEvent(invalidIP)
```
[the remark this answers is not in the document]
BeegiiK: The naming of the variable is misleading. I'm trying to test how the enricher reacts when the channel is full. I'll update the tests soon using the table tests approach

deploy/standard/manifests/controller/helm/retina/values.yaml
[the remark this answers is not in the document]
BeegiiK: Nice catch Alex! I've removed the toggles from the helm files. Addressed in the next push!

pkg/controllers/cache/standalone_cache_test.go
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package cache
```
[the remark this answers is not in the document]
BeegiiK: I'll keep this in mind with my tests. I'll update the current tests I've written to follow this approach. Thanks Alex!

cmd/standalone_daemon.go
```diff
@@ -0,0 +1,90 @@
...
+	g.Go(func() error {
+		return sd.pluginManager.Start(ctx)
+	})
+	g.Go(func() error {
+		return sd.httpServer.Start(ctx)
+	})
+
+	if err := g.Wait(); err != nil {
+		mainLogger.Panic("Error running standalone daemon", zap.Error(err))
+	}
```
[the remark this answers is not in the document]
BeegiiK: Done! Thanks Alex

cmd/root.go
```diff
@@ -27,9 +26,9 @@ var (
 		Long:  "Start Retina Agent",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// Do Stuff Here
-			fmt.Println("Starting Retina Agent")
-			d := standard.NewDaemon(metricsAddr, probeAddr, cfgFile, enableLeaderElection)
-			if err := d.Start(); err != nil {
+			fmt.Println("Bootstrapping Retina")
+			b := NewBootstrapManager(metricsAddr, probeAddr, cfgFile, enableLeaderElection)
```
anubhabMajumdar: I think `standalone` should be a new command (similar to `hubble-control-plane`)
```bash
$ ./retina-agent -h
Usage:
  retina-agent [flags]
  retina-agent [command]

Available Commands:
  completion           Generate the autocompletion script for the specified shell
  help                 Help about any command
  hubble-control-plane Start Hubble control plane
  standalone Start retina without K8s control plane
```

This untangles the standalone from standard. Standalone is a completely separate mode and should be handled as such.

anubhabMajumdar: Also, this isolates Standalone from impacting Standard in any way.

pkg/config/config.go
```diff
@@ -68,6 +68,8 @@ type Config struct {
 	EnableTelemetry          bool          `yaml:"enableTelemetry"`
 	EnableRetinaEndpoint     bool          `yaml:"enableRetinaEndpoint"`
 	EnablePodLevel           bool          `yaml:"enablePodLevel"`
+	EnableStandalone         bool          `yaml:"enableStandalone"`
+	EnableCrictl             bool          `yaml:"enableCrictl"`
```
anubhabMajumdar: The fields of this config structure were modeled for standard control plane, with many of them specific for a K8s control plane. Instead of adding fields for standalone here, we should have a new config structure specifically for standalone. This makes it cleaner and easier to maintain.

pkg/controllers/daemon/standalone/source/ctrinfo.go
```diff
@@ -0,0 +1,101 @@
...
+	crictlCommand = runCommand
+
+	errGetPods    = errors.New("failed to get running pods")
+	errInspectPod = errors.New("failed to inspect pod information")
+	errJSONRead   = errors.New("error unmarshalling JSON")
+)
+
+func (c *Ctrinfo) GetAllEndpoints() ([]*common.RetinaEndpoint, error) {
+	// Using crictl to get all running pods
+	runningPods, err := crictlCommand("cmd", "/c", "crictl", "pods", "-q")
```
anubhabMajumdar: `cmd /c` is Windows specific. Why is that the default? I would imagine that would be decided based on where agent is running.

pkg/controllers/daemon/standalone/source/ctrinfo.go
```diff
@@ -0,0 +1,101 @@
...
+			spec.Status.Metadata.Namespace,
+			common.NewIPAddress(ip, nil),
+		))
+	}
+
+	return endpoints, nil
+}
+
+func runCommand(command string, args ...string) (string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
anubhabMajumdar: Let's not hard code this. Should we make this configurable?

pkg/controllers/daemon/standalone/source/statefile.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"net"
+	"os"
+
+	"github.com/microsoft/retina/pkg/common"
+)
+
+type Statefile struct{}
+
+var StateFileLocation = "C:/Windows/System32/azure-vnet.json"
```
anubhabMajumdar: This should be configurable.
Also, this seems to be Azure/Windows specific. And I would name the file as such.

pkg/config/config.go
anubhabMajumdar: I don't like this config setup like this. Would rather like to see something like
`--enrichment-mode=crictl/azure-vnet-statefile` (please decide on a better flag name).

pkg/enricher/enricher.go
```diff
@@ -98,8 +105,37 @@ func (e *Enricher) Run() {
 	}()
 }
 
-// enrich takes the flow and enriches it with the information from the cache
 func (e *Enricher) enrich(ev *v1.Event) {
+	if e.enableStandalone {
```
anubhabMajumdar: Don't inject the standalone/standard config in modules. Rather refactor to have separate enricher for Standard and Standalone. Common code can reside in a common package.

pkg/managers/controllermanager/controllermanager.go
```diff
@@ -83,7 +83,7 @@ func (m *Controller) Init(ctx context.Context) error {
 		m.cache = cache.New(m.pubsub)
 
 		// create enricher instance
-		m.enricher = enricher.New(ctx, m.cache)
+		m.enricher = enricher.New(ctx, m.cache, m.conf.EnableStandalone)
```
anubhabMajumdar: Reiterating - Don't inject the standalone/standard config in modules. Have separate modules for Standalone.

cmd/root.go
BeegiiK: Thanks for the review Anubhab! I've updated the PR in the latest commit to address this

pkg/controllers/daemon/standalone/source/ctrinfo.go
BeegiiK: Good catch, I wasn't thinking operations on the Linux VM. Updated in the latest commit

PR #1452, opened 2025-03-20
fix(basic-metrics): improve throughput by removing some kprobes 

### LINKED ISSUE #1451: Some drop reasons kernel return codes are ignored
`nf_hook_slow_ret` ignores the return code it gets from the kretprobe and always passes 0 along with its drop reason. This accidentally reduces the cadrinality of this metric, but also strips away potentially useful diagnostic info.
https://github.com/microsoft/retina/blob/main/pkg/plugin/dropreason/_cprog/drop_reason.c#L288
Same for `nf_nat_inet_fn_ret`
https://github.com/microsoft/retina/blob/main/pkg/plugin/dropreason/_cprog/drop_reason.c#L413

rectified95:
Make basic mode only track the number of dropped packets (skipping byte count) - that allows us not to register the kprobes, but only the matching retprobes, cutting program invocatinos in half. 
On the test cluster, throughput went up from 17.5 to 19.5 Gbps (vs. 22.5 baseline)

This PR adds new probes, rather than trying to modify the existing ones with pre-processor directives.
There's new logic that loads the right set of probes depending on basic/advanced mode.

Also Fixes #1451


rectified95:
Want to add latency numbers apart from tput before merging.

rectified95:
Superseded by #1458

Issue #715, opened 2024-09-10
inet_csk_accept ebpf program doesn't work properly


rectified95:
`kretprobe/inet_csk_accept` (reason `TCP_ACCEPT_BASIC`) has two big issues:

1. it doesn't track **bytes dropped** for any packet it processes (it calls `get_packet_from_sock` which doesn't populate the packet's `skb_len`, since that value is not available in the socket struct this ebpf hook receives as input; this means that the subsequent call to `update_metrics_map` doesn't increase the drop bytes metrics)
2. it misses many drops (it does a single rather than double pointer dereference of the value retrieved from `accept_pids` map; the rerieved pointer (and not underlying value) is assigned to `int` which given 2's complement notation means that anytime the memory address of the `err` value has the most significant bit == 0 (positive number), it returns early thinking the underlying function didn't indicate a drop)

Solutions: 
1. We'd need to figure out how to get the sk_buff from the socket.
2. I tried 2 approaches unsuccessfully so far: doing the second pointer deref (ebpf verifier failure); storing the actual error value from `inet_csk_accept` rather than the pointer, as a signed int, which obviates the need for the double deref, but also got a verifier error.

rectified95:
Re. 2 We need to store the pointer as we are doing now, because the underlying value is not set until the function completes. It is an out-parameter, since the return value is a socket. Still, it seems like this funciton doesn't track this drop reason correctly.

PR #1458, opened 2025-03-22
fix(basic-metrics): Use fexit instead of kprobes for DropReason

### LINKED ISSUE #1451: Some drop reasons kernel return codes are ignored

### LINKED ISSUE #1466: Some kernel return codes are always corrupted for DropReason
A few instances of passing a negative return code to an unsigned map key. THis always overflows, and the resulting metric value is partially meaningless.
Eg.: https://github.com/microsoft/retina/blob/main/pkg/plugin/dropreason/_cprog/drop_reason.c#L310

Related to #1451

### LINKED ISSUE #1469: DropReason should use atomics for drop count
__sync_fetch_and_add(&entry->count, 1);

rectified95:
### Description
`fexit` programs capture both the function arguments and the return value. By using them we don't need the pairs of kprobe-kretprobe programs to correlate program invocations for packets.

This reduces the number of syscalls by half, while this program type is also more efficient than kprobe, as measured by this benchmark: https://github.com/cloudflare/ebpf_exporter/tree/master/benchmark
I've reproduced it and have seen more moderate but noticeable differences.

The largest perf gain comes from slashing the number of eBPF program invocations.
This PR covers the "basic metrics" mode, and a follow-up is ready for the advanced scenario, removing kprobes altogether.

We use the new program type for kernel versions >= 5.5 (AMD64), 6.0 (ARM64) the existing kprobes for older kernels.
On Azure Linux (Mariner) we skip attaching certain eBPF programs; this distro currently doesn't contain BTF info for kernel modules, and some of the functions we attach to are outside of the core kernel.

#### 99% CPU 50 streams, MTU 100 B, cross-node pods
- w/o Retina: **8.4**
- w/ this PR: **8.17**
- w/o this PR: **6.2**

#### Single TCP stream, same node pods
- w/o Retina: **21.2**
- w/ this PR: **20.5**
- w/ this PR, all dropped: 19.7 (unconditionally execute entire programs, forcing map lookups)
- w/o this PR: **17.2**

#### 50 streams, default MTU, same node
- w/o Retina: **420**
- w/ this PR: **400**
- w/o this PR: **40**

#### 50 streams, default MTU, cross-node
30Gbps achieved for all.

### Other changes: 
- fix overflow when storing kernel nagative return codes in an unsigned type
- fix some kernel return codes being ignored and not recorded
- fix drop packet counter not using atomic increment
- fix up error formatting using `zap`

Fixes #1469 
Fixes #1466
Fixes #1451
Related https://github.com/microsoft/retina/issues/715


pkg/plugin/dropreason/_cprog/drop_reason.c
```diff
@@ -28,7 +28,8 @@ char __license[] SEC("license") = "Dual MIT/GPL";
 struct metrics_map_key
 {
     __u16 drop_type;
-    __u32 return_val;
+    __u8 padding[2];
```
rectified95, line 31: Need aligned access, clang didn't add padding

pkg/plugin/dropreason/_cprog/drop_reason.c
```diff
@@ -108,11 +110,12 @@ struct
             member_address(source_struct, source_member));     \
     } while (0)
 
+// Updates dropped packets stats - ONLY use in Retina advanced.
 void update_metrics_map(void *ctx, drop_reason_t drop_type, int ret_val, struct packet *p)
 {
     struct metrics_map_value *entry, new_entry = {};
     struct metrics_map_key key;
-    __builtin_memset(&key, 0, sizeof(p));
+    __builtin_memset(&key, 0, sizeof(key));
```
rectified95, line 118: this was wrong

pkg/plugin/dropreason/_cprog/drop_reason.c
```diff
@@ -285,15 +303,27 @@ int BPF_KRETPROBE(nf_hook_slow_ret, int retVal)
 
     bpf_map_delete_elem(&retina_dropreason_drop_pids, &pid);
 
-    update_metrics_map(ctx, IPTABLE_RULE_DROP, 0, p);
+    update_metrics_map(ctx, IPTABLE_RULE_DROP, retVal, p);
```
rectified95, line 306: pass actual return val isntead of zero

nddq (approved):
some minor comments but otherwise lgtm

pkg/plugin/dropreason/dropreason_linux.go
```diff
@@ -444,13 +398,147 @@ func (dr *dropReason) processMapValue(dataKey dropMetricKey, dataValue dropMetri
...
 	dr.l.Debug("DATA From the DropReason Map", zap.String("Droptype", dataKey.getType()),
-		zap.Uint32("Return Val", dataKey.ReturnVal),
+		zap.Int32("Return Val", dataKey.ReturnVal),
 		zap.Int("DropCount", int(pktCount)),
 		zap.Int("DropBytes", int(pktBytes)))
 
 	dr.dropMetricAdd(dataKey.getType(), dataKey.getDirection(), pktCount, pktBytes)
 }
 
+func (dr *dropReason) attachKprobes(objs *kprobeObjects) error {
```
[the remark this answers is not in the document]
rectified95: i can do that in the follow-up PR for advanced metrics

pkg/plugin/dropreason/dropreason_linux.go
```diff
@@ -444,13 +398,147 @@ func (dr *dropReason) processMapValue(dataKey dropMetricKey, dataValue dropMetri
...
+	var err error
+
+	dr.KNfHook, err = link.Kprobe(nfHookSlowFn, objs.NfHookSlow, nil)
+	if err != nil {
+		dr.l.Error("opening kprobe", zap.Error(err))
+		return fmt.Errorf("Failed to attach program: %w", err) //nolint:goerr113 //wrapping error from external module
```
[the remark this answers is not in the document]
rectified95: doesn't fmt.Errorf wrap as well?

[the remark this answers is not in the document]
nddq: does it? if we remove `//nolint:goerr113`, does the linter has an issue with it?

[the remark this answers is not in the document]
rectified95: It does, see here: https://docs.datadoghq.com/security/code_security/static_analysis/static_analysis_rules/go-best-practices/errors-new-errorf/

I added this linter ignore because somehow it doesn't correctly recognize the wrapping

deploy/standard/manifests/controller/helm/retina/values.yaml
```diff
@@ -98,6 +98,7 @@ volumeMounts:
   cgroup: /sys/fs/cgroup
   tmp: /tmp
   config: /retina/config
+  host-os-release: /etc/os-release
```
rectified95, line 101: This volumeMount is needed to determine if the host is Mariner, so we can decide which ebpf programs to attach. 
Retina's base image is Mariner, so the file scan of /etc/os-release would always return true otherwise.

pkg/plugin/common/common_linux.go
```diff
@@ -89,3 +94,47 @@ func NewPerfReader(l *log.ZapLogger, m *ebpf.Map, max, min int) (*perf.Reader, e
 	}
 	return nil, errors.New("failed to create perf reader")
 }
+
+// Function to only extract the X.Y version components, used as fallback for version.GetKernelVersion()
+// Needed for images which don't follow standard Linux version format, e.g.: 5.15.153.1-microsoft-standard-WSL2
+func GetKernelVersionMajMin() (semver.Version, error) {
+	var unameBuf unix.Utsname
+	if err := unix.Uname(&unameBuf); err != nil {
+		return semver.Version{}, err //nolint:wrapcheck // no additional context needed
```
MikeZappa87, line 103: return nil, err?

test/plugin/dropreason/main_linux.go
```diff
@@ -67,31 +67,31 @@ func main() {
 
 	err = tt.Stop()
 	if err != nil {
-		l.Error("Stop failed:%v", zap.Error(err))
+		l.Error("Stop failed:", zap.Error(err))
```
MikeZappa87, line 70: nit: these log statement changes should have been in a different PR.

deploy/standard/manifests/controller/helm/retina/values.yaml
MikeZappa87, line 101: I have come across distros where this file is missing since this is open source we should be aware of this. I don't believe Linux states this MUST be present.

test/plugin/dropreason/main_linux.go
rectified95: Not required, just saw garbled logs so fixed the,

deploy/standard/manifests/controller/helm/retina/values.yaml
rectified95: Right, it's a standard, for not required. Hb this: I remove `Type: File`, and we'll avoid pod start failures if file is missing, k8s will mount and empty dir, which the code handles.

rectified95: @MikeZappa87 Changed this to `FileOrCreate` to avoid pods stuck in Init. Mariner will always have this file non-empty, so this fallback will correctly apply for non-Mariner distros. 
I tested this logic manually.

matmerr (approved):

PR #1468, opened 2025-03-25
ebpfwindows-plugin

vpidatala94:
ebpfwindows plugin to create basic and advanced metrics
advanced metrics for trace and drop events are created by using hubble parser

### DIFF
--- a/pkg/plugin/ebpfwindows/ebpf_windows.go
+++ b/pkg/plugin/ebpfwindows/ebpf_windows.go
@@ -1,0 +1,277 @@
+package ebpfwindows
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"unsafe"
+
+	v1 "github.com/cilium/cilium/pkg/hubble/api/v1"
+	observer "github.com/cilium/cilium/pkg/hubble/observer/types"
+	hp "github.com/cilium/cilium/pkg/hubble/parser"
+	kcfg "github.com/microsoft/retina/pkg/config"
+	"github.com/microsoft/retina/pkg/enricher"
+	"github.com/microsoft/retina/pkg/log"
+	"github.com/microsoft/retina/pkg/metrics"
+	"github.com/microsoft/retina/pkg/plugin/registry"
+	"github.com/microsoft/retina/pkg/utils"
+	"github.com/sirupsen/logrus"
+	"go.uber.org/zap"
+)
+
+const (
+	// name of the ebpfwindows plugin
+	name string = "ebpfwindows"
+	// name of the metrics
+	packetsReceived        string = "win_packets_recv_count"
+	packetsSent            string = "win_packets_sent_count"
+	bytesSent              string = "win_bytes_sent_count"
+	bytesReceived          string = "win_bytes_recv_count"
+	droppedPacketsIncoming string = "win_packets_recv_drop_count"
+	droppedPacketsOutgoing string = "win_packets_sent_drop_count"
+	// metrics direction
+	ingressLabel = "ingress"
+	egressLabel  = "egress"
+)
+
+var (
+	ErrInvalidEventData = errors.New("The Event Data is invalid")
+	ErrNilEnricher      = errors.New("enricher is nil")
+)
+
+// Plugin is the ebpfwindows plugin
+type Plugin struct {
+	l               *log.ZapLogger
+	cfg             *kcfg.Config
+	enricher        *enricher.Enricher
+	externalChannel chan *v1.Event
+	parser          *hp.Parser
+}
+
+func init() {
+	registry.Add(name, New)
+}
+
+func New(cfg *kcfg.Config) registry.Plugin {
+	return &Plugin{
+		l:   log.Logger().Named(name),
+		cfg: cfg,
+	}
+}
+
+// Init is a no-op for the ebpfwindows plugin
+func (p *Plugin) Init() error {
+	parser, err := hp.New(logrus.WithField("windowsEbpf", "parser"),
+		// We use noop getters here since we will use our own custom parser in hubble
+		&NoopEndpointGetter,
+		&NoopIdentityGetter,
+		&NoopDNSGetter,
+		&NoopIPGetter,
+		&NoopServiceGetter,
+		&NoopLinkGetter,
+		&NoopPodMetadataGetter,
+	)
+
+	if err != nil {
+		p.l.Fatal("Failed to create parser", zap.Error(err))
+		return err
+	}
+
+	p.parser = parser
+	return nil
+}
+
+// Name returns the name of the ebpfwindows plugin
+func (p *Plugin) Name() string {
+	return name
+}
+
+// Start the plugin by starting a periodic timer.
+func (p *Plugin) Start(ctx context.Context) error {
+	p.l.Info("Start ebpfWindows plugin...")
+	p.pullMetricsAndEvents(ctx)
+	p.l.Info("Complete ebpfWindows plugin...")
+	return nil
+}
+
+// metricsMapIterateCallback is the callback function that is called for each key-value pair in the metrics map.
+func (p *Plugin) metricsMapIterateCallback(key *MetricsKey, value *MetricsValues) {
+	p.l.Debug("MetricsMapIterateCallback")
+	p.l.Debug("Key", zap.String("Key", key.String()))
+	p.l.Debug("Value", zap.String("Value", value.String()))
+	if key.IsDrop() {
+		if key.IsEgress() {
+			metrics.DropBytesGauge.WithLabelValues(DropReason(key.Reason), egressLabel).Set(float64(value.Bytes()))
+			metrics.DropPacketsGauge.WithLabelValues(DropReason(key.Reason), egressLabel).Set(float64(value.Count()))
+		} else if key.IsIngress() {
+			metrics.DropBytesGauge.WithLabelValues(DropReason(key.Reason), ingressLabel).Set(float64(value.Bytes()))
+			metrics.DropPacketsGauge.WithLabelValues(DropReason(key.Reason), ingressLabel).Set(float64(value.Count()))
+		}
+	} else {
+		if key.IsEgress() {
+			metrics.ForwardBytesGauge.WithLabelValues(egressLabel).Set(float64(value.Bytes()))
+			p.l.Debug("emitting bytes sent count metric", zap.Uint64(bytesSent, value.Bytes()))
+			metrics.ForwardBytesGauge.WithLabelValues(egressLabel).Set(float64(value.Count()))
+			p.l.Debug("emitting packets sent count metric", zap.Uint64(packetsSent, value.Count()))
+		} else if key.IsIngress() {
+			metrics.ForwardPacketsGauge.WithLabelValues(ingressLabel).Set(float64(value.Count()))
+			p.l.Debug("emitting packets received count metric", zap.Uint64(packetsReceived, value.Count()))
+			metrics.ForwardBytesGauge.WithLabelValues(ingressLabel).Set(float64(value.Bytes()))
+			p.l.Debug("emitting bytes received count metric", zap.Uint64(bytesReceived, value.Bytes()))
+		}
+	}
+}
+
+// eventsMapCallback is the callback function that is called for each value  in the events map.
+func (p *Plugin) eventsMapCallback(data unsafe.Pointer, size uint32) int {
+	p.l.Info("EventsMapCallback")
+	p.l.Info("Size", zap.Uint32("Size", size))
+	err := p.handleTraceEvent(data, size)
+	if err != nil {
+		p.l.Error("Error handling trace event", zap.Error(err))
+		return -1
+	}
+	return 0
+}
+
+func addEbpfToPath() error {
+	currPath := os.Getenv("PATH")
+	if strings.Contains(currPath, "ebpf-for-windows") {
+		return nil
+	}
+	programFiles := os.Getenv("ProgramFiles")
+	ebpfWindowsPath := programFiles + "\\ebpf-for-windows\\"
+	newPath := currPath + ";" + ebpfWindowsPath
+	if err := os.Setenv("PATH", newPath); err != nil {
+		return fmt.Errorf("error setting PATH environment variable: %v", err)
+	}
+
+	return nil
+}
+
+func (p *Plugin) pullMetricsAndEvents(ctx context.Context) {
+	eventsMap := NewEventsMap()
+	metricsMap := NewMetricsMap()
+
+	err := addEbpfToPath()
+	if err != nil {
+		return
+	}
+
+	if enricher.IsInitialized() {
+		p.enricher = enricher.Instance()
+	} else {
+		p.l.Warn("retina enricher is not initialized")
+	}
+
+	if p.enricher != nil {
+		err := eventsMap.RegisterForCallback(p.eventsMapCallback)
+		if err != nil {
+			p.l.Error("Error registering for events map callback", zap.Error(err))
+			return
+		}
+
+		defer func() {
+			p.l.Error("ebpfwindows plugin canceling", zap.Error(ctx.Err()))
+			eventsMap.UnregisterForCallback()
+		}()
+	}
+
+	ticker := time.NewTicker(p.cfg.MetricsInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			err := metricsMap.IterateWithCallback(p.metricsMapIterateCallback)
+			if err != nil {
+				p.l.Error("Error iterating metrics map", zap.Error(err))
+			}
+		case <-ctx.Done():
+			p.l.Error("ebpfwindows plugin canceling", zap.Error(ctx.Err()))
+			err := eventsMap.UnregisterForCallback()
+
+			if err != nil {
+				p.l.Error("Error Unregistering Events Map callback", zap.Error(err))
+			}
+			return
+		}
+	}
+}
+
+// SetupChannel saves the external channel to which the plugin will send events.
+func (p *Plugin) SetupChannel(ch chan *v1.Event) error {
+	p.externalChannel = ch
+	return nil
+}
+
+// Stop the plugin by cancelling the periodic timer.
+func (p *Plugin) Stop() error {
+	p.l.Info("Stop ebpfWindows plugin...")
+	return nil
+}
+
+// Compile is a no-op for the ebpfwindows plugin
+func (p *Plugin) Compile(context.Context) error {
+	return nil
+}
+
+// Generate is a no-op for the ebpfwindows plugin
+func (p *Plugin) Generate(context.Context) error {
+	return nil
+}
+
+func (p *Plugin) handleTraceEvent(data unsafe.Pointer, size uint32) error {
+	if uintptr(size) < unsafe.Sizeof(uint8(0)) {
+		return ErrInvalidEventData
+	}
+	eventType := *(*uint8)(data)
+	switch eventType {
+	case NotifyDrop:
+		if uintptr(size) < unsafe.Sizeof(DropNotify{}) {
+			p.l.Error("Invalid DropNotify data size", zap.Uint32("size", size))
+			return ErrInvalidEventData
+		}
+		e, err := p.parser.Decode(&observer.MonitorEvent{
+			Payload: &observer.PerfEvent{
+				Data: (*[unsafe.Sizeof(DropNotify{})]byte)(data)[:],
+			},
+		})
+		if err != nil {
+			p.l.Error("Could not convert event to flow", zap.Any("handleTraceEvent", data), zap.Error(err))
+			return ErrInvalidEventData
+		}
+		meta := &utils.RetinaMetadata{}
+		// Add packet size to the flow's metadata.
+		utils.AddPacketSize(meta, 128)
+		fl := e.GetFlow()
+		dropNotify := (*DropNotify)(data)
+		meta.DropReason = utils.DropReason(dropNotify.Subtype)
+		utils.AddRetinaMetadata(fl, meta)
+		p.enricher.Write(e)
+	case NotifyTrace:
+		if uintptr(size) < unsafe.Sizeof(TraceNotify{}) {
+			p.l.Error("Invalid TraceNotify data size", zap.Uint32("size", size))
+			return ErrInvalidEventData
+		}
+		e, err := p.parser.Decode(&observer.MonitorEvent{
+			Payload: &observer.PerfEvent{
+				Data: (*[unsafe.Sizeof(TraceNotify{})]byte)(data)[:],
+			},
+		})
+		if err != nil {
+			p.l.Error("Could not convert event to flow", zap.Any("handleTraceEvent", data), zap.Error(err))
+			return ErrInvalidEventData
+		}
+		meta := &utils.RetinaMetadata{}
+		// Add packet size to the flow's metadata.
+		utils.AddPacketSize(meta, 128)
+		fl := e.GetFlow()
+		utils.AddRetinaMetadata(fl, meta)
+		p.enricher.Write(e)
+	}
+	return nil
+}
--- a/pkg/plugin/ebpfwindows/eventsmap_types_windows.go
+++ b/pkg/plugin/ebpfwindows/eventsmap_types_windows.go
@@ -1,0 +1,123 @@
+package ebpfwindows
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"net"
+)
+
+// IP represents an IPv4 or IPv4 or IPv6 address
+type IP struct {
+	Address uint32
+	Pad1    uint32
+	Pad2    uint32
+	Pad3    uint32
+}
+
+// TraceSockNotify is the notification for a socket trace
+type TraceSockNotify struct {
+	Type       uint8
+	XlatePoint uint8
+	DstIP      IP
+	DstPort    uint16
+	SockCookie uint64
+	CgroupID   uint64
+	L4Proto    uint8
+	Ipv6       bool
+	Data       [128]byte
+}
+
+// NotifyCommonHdr is the common header for all notifications
+type NotifyCommonHdr struct {
+	Type    uint8
+	Subtype uint8
+	Source  uint16
+	Hash    uint32
+}
+
+// NotifyCaptureHdr is the common header for all capture notifications
+type NotifyCaptureHdr struct {
+	NotifyCommonHdr
+	LenOrig uint32 // Length of original packet
+	LenCap  uint16 // Length of captured bytes
+	Version uint16 // Capture header version
+}
+
+// DropNotify is the notification for a packet drop
+type DropNotify struct {
+	NotifyCaptureHdr
+	SrcLabel uint32
+	DstLabel uint32
+	DstID    uint32 // 0 for egress
+	Line     uint16
+	File     uint8
+	ExtError int8
+	Ifindex  uint32
+	Data     [128]byte
+}
+
+// TraceNotify is the notification for a packet trace
+type TraceNotify struct {
+	NotifyCaptureHdr
+	SrcLabel uint32
+	DstLabel uint32
+	DstID    uint16
+	Reason   uint8
+	Ipv6     bool
+	Ifindex  uint32
+	OrigIP   IP
+	Data     [128]byte
+}
+
+// Notification types
+const (
+	NotifyUnspec        = 0
+	NotifyDrop          = 1
+	NotifyDebugMessage  = 2
+	NotifyDebugCapture  = 3
+	NotifyTrace         = 4
+	NotifyPolicyVerdict = 5
+	NotifyCapture       = 6
+	NotifyTraceSock     = 7
+)
+
+func (ip *IP) ConvertToString(Ipv6 bool) string {
+	var ipAddress string
+	var buf bytes.Buffer
+
+	err := binary.Write(&buf, binary.BigEndian, *ip)
+
+	if err != nil {
+		return ""
+	}
+
+	byteArray := buf.Bytes()
+
+	if Ipv6 {
+		ipAddress = net.IP(byteArray[:16]).String()
+	} else {
+		ipAddress = net.IP(byteArray[:4]).String()
+	}
+
+	return ipAddress
+
+}
+
+// String returns a string representation of the DropNotify
+func (k *DropNotify) String() string {
+
+	return fmt.Sprintf("Ifindex: %d, SrcLabel:%d, DstLabel:%d, File: %s, Line: %d", k.Ifindex, k.SrcLabel, k.DstLabel, BPFFileName(k.File), k.Line)
+}
+
+// String returns a string representation of the TraceNotify
+func (k *TraceNotify) String() string {
+	ipAddress := k.OrigIP.ConvertToString(k.Ipv6)
+	return fmt.Sprintf("Ifindex: %d, SrcLabel:%d, DstLabel:%d, IpV6:%t, OrigIP:%s", k.Ifindex, k.SrcLabel, k.DstLabel, k.Ipv6, ipAddress)
+}
+
+// String returns a string representation of the TraceSockNotify
+func (k *TraceSockNotify) String() string {
+	ipAddress := k.DstIP.ConvertToString(k.Ipv6)
+	return fmt.Sprintf("DstIP:%s, DstPort:%d, SockCookie:%d, CgroupID:%d, L4Proto:%d, IPv6:%t", ipAddress, k.DstPort, k.SockCookie, k.CgroupID, k.L4Proto, k.Ipv6)
+}
--- a/pkg/plugin/ebpfwindows/eventsmap_windows.go
+++ b/pkg/plugin/ebpfwindows/eventsmap_windows.go
@@ -1,0 +1,75 @@
+package ebpfwindows
+
+import (
+	"syscall"
+	"unsafe"
+)
+
+var (
+	registerEventsMapCallback   = retinaEbpfAPI.NewProc("RetinaRegisterEventsMapCallback")
+	unregisterEventsMapCallback = retinaEbpfAPI.NewProc("RetinaUnregisterEventsMapCallback")
+)
+
+type eventsMapCallback func(data unsafe.Pointer, size uint32) int
+
+// Callbacks in Go can only be passed as functions with specific signatures and often need to be wrapped in a syscall-compatible function.
+var eventsCallback eventsMapCallback
+
+// This function will be passed to the Windows API
+func eventsMapSysCallCallback(data unsafe.Pointer, size uint32) uintptr {
+
+	if eventsCallback != nil {
+		return uintptr(eventsCallback(data, size))
+	}
+
+	return 0
+}
+
+// EventsMap interface represents a events map
+type EventsMap interface {
+	RegisterForCallback(eventsMapCallback) error
+	UnregisterForCallback() error
+}
+
+type eventsMap struct {
+	ringBuffer uintptr
+}
+
+// NewEventsMap creates a new metrics map
+func NewEventsMap() EventsMap {
+	return &eventsMap{ringBuffer: 0}
+}
+
+// RegisterForCallback registers a callback function to be called when a new event is added to the events map
+func (e *eventsMap) RegisterForCallback(cb eventsMapCallback) error {
+
+	eventsCallback = cb
+
+	// Convert the Go function into a syscall-compatible function
+	callback := syscall.NewCallback(eventsMapSysCallCallback)
+
+	// Call the API
+	ret, _, err := registerEventsMapCallback.Call(
+		uintptr(callback),
+		uintptr(unsafe.Pointer(&e.ringBuffer)),
+	)
+
+	if ret != 0 {
+		return err
+	}
+
+	return nil
+}
+
+// UnregisterForCallback unregisters the callback function
+func (e *eventsMap) UnregisterForCallback() error {
+
+	// Call the API
+	ret, _, err := unregisterEventsMapCallback.Call(e.ringBuffer)
+
+	if ret != 0 {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/plugin/ebpfwindows/metricsmap_windows.go
+++ b/pkg/plugin/ebpfwindows/metricsmap_windows.go
@@ -1,0 +1,190 @@
+package ebpfwindows
+
+import (
+	"fmt"
+	"reflect"
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+)
+
+const (
+	dirUnknown = 0
+	dirIngress = 1
+	dirEgress  = 2
+	dirService = 3
+)
+
+// direction is the metrics direction i.e ingress (to an endpoint),
+// egress (from an endpoint) or service (NodePort service being accessed from
+// outside or a ClusterIP service being accessed from inside the cluster).
+// If it's none of the above, we return UNKNOWN direction.
+var direction = map[uint8]string{
+	dirUnknown: "UNKNOWN",
+	dirIngress: "INGRESS",
+	dirEgress:  "EGRESS",
+	dirService: "SERVICE",
+}
+
+// Value must be in sync with struct metrics_key in <bpf/lib/common.h>
+type MetricsKey struct {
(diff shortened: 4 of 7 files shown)


pkg/plugin/ebpfwindows/eventsmap_types_windows.go
```diff
@@ -0,0 +1,123 @@
...
+	NotifyDrop          = 1
+	NotifyDebugMessage  = 2
+	NotifyDebugCapture  = 3
+	NotifyTrace         = 4
+	NotifyPolicyVerdict = 5
+	NotifyCapture       = 6
+	NotifyTraceSock     = 7
+)
+
+func (ip *IP) ConvertToString(Ipv6 bool) string {
```
matmerr, line 85: dupe on net.IP comment, these string methods are already implemented

https://pkg.go.dev/net/netip#Addr.String

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return ipAddress
+
+}
+
+// String returns a string representation of the DropNotify
+func (k *DropNotify) String() string {
+
+	return fmt.Sprintf("Ifindex: %d, SrcLabel:%d, DstLabel:%d, File: %s, Line: %d", k.Ifindex, k.SrcLabel, k.DstLabel, BPFFileName(k.File), k.Line)
```
matmerr, line 110: Overriding `String()` and manually printing out properties can be a slippery slope, especially if the struct has added types later on and the `String()` method isn't updated, and then when used to debug context can be lost in the future. Example is this case here, we don't print `ExtError` which is a property of `DropNotify`, and if that's relevant in the future we'd be missing it

I think I'd prefer to leave this Stringer as is, and when calling `fmt.Printf("%v", k)`, also call `BPFFileName(k.File)` in the log line, or if really required add a unreserved nonstringer method with a name like `k.GetString()` or otherwise

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// String returns a string representation of the TraceNotify
+func (k *TraceNotify) String() string {
+	ipAddress := k.OrigIP.ConvertToString(k.Ipv6)
+	return fmt.Sprintf("Ifindex: %d, SrcLabel:%d, DstLabel:%d, IpV6:%t, OrigIP:%s", k.Ifindex, k.SrcLabel, k.DstLabel, k.Ipv6, ipAddress)
```
matmerr, line 116: +1 hesitant to override the stringer, for same reasons as dropnotify stringer, things such as `DstID` is missing

pkg/plugin/ebpfwindows/eventsmap_windows.go
```diff
@@ -0,0 +1,75 @@
...
+	callback := syscall.NewCallback(eventsMapSysCallCallback)
+
+	// Call the API
+	ret, _, err := registerEventsMapCallback.Call(
+		uintptr(callback),
+		uintptr(unsafe.Pointer(&e.ringBuffer)),
+	)
+
+	if ret != 0 {
+		return err
```
matmerr, line 58: unwrapped error, this seems like a pretty bad case if we hit this, so any extra context would be helpful

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,277 @@
...
+	return nil
+}
+
+func (p *Plugin) pullMetricsAndEvents(ctx context.Context) {
+	eventsMap := NewEventsMap()
+	metricsMap := NewMetricsMap()
+
+	err := addEbpfToPath()
+	if err != nil {
+		return
```
matmerr, line 162: log this error, if we return here and we exit Start() there's really no indication of what happened

pkg/plugin/ebpfwindows/metricsmap_windows.go
```diff
@@ -0,0 +1,190 @@
...
+
+// IterateWithCallback iterates through all the keys/values of a metrics map,
+// passing each key/value pair to the cb callback
+func (m metricsMap) IterateWithCallback(cb IterateCallback) error {
+
+	// Define the callback function in Go
+	enumCallBack = func(key unsafe.Pointer, value unsafe.Pointer, valueSize int) int {
+
+		var metricsValues MetricsValues
+		sh := (*reflect.SliceHeader)(unsafe.Pointer(&metricsValues))
```
matmerr, line 100: I don't see any tests around this callback paradigm, can we add one? Also hesitant to use reflection with anything that has potential heavy invocation, even in the reflect package it warns deprecated

> It cannot be used safely or portably and its representation may change in a later release. 

https://cs.opensource.google/go/go/+/go1.24.2:src/reflect/value.go;l=2558

pkg/plugin/ebpfwindows/metricsmap_windows.go
```diff
@@ -0,0 +1,190 @@
...
+	return k.Dir == dirIngress
+}
+
+// IsEgress checks if the direction is egress or not.
+func (k *MetricsKey) IsEgress() bool {
+	return k.Dir == dirEgress
+}
+
+// Count returns the sum of all the per-CPU count values
+func (vs MetricsValues) Count() uint64 {
```
matmerr, line 169: `Count()` naming sound like `len(vs)` despite it being a sum, can we rename to `MetricsSum()` or similar?

pkg/plugin/ebpfwindows/metricsmap_windows.go
```diff
@@ -0,0 +1,190 @@
...
+	c := uint64(0)
+	for _, v := range vs {
+		c += v.Count
+	}
+
+	return c
+}
+
+// Bytes returns the sum of all the per-CPU bytes values
+func (vs MetricsValues) Bytes() uint64 {
```
matmerr, line 179: `Bytes()` naming could also be ambiguous since other usage of Bytes() methods may expect a byte array returned, maybe `BytesSum()`?

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,277 @@
...
+				Data: (*[unsafe.Sizeof(DropNotify{})]byte)(data)[:],
+			},
+		})
+		if err != nil {
+			p.l.Error("Could not convert event to flow", zap.Any("handleTraceEvent", data), zap.Error(err))
+			return ErrInvalidEventData
+		}
+		meta := &utils.RetinaMetadata{}
+		// Add packet size to the flow's metadata.
+		utils.AddPacketSize(meta, 128)
```
matmerr, line 249: `128` here seems like a magic number given the context, can we assign this as named variable? or determine packet size programmatically?

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,277 @@
...
+				Data: (*[unsafe.Sizeof(TraceNotify{})]byte)(data)[:],
+			},
+		})
+		if err != nil {
+			p.l.Error("Could not convert event to flow", zap.Any("handleTraceEvent", data), zap.Error(err))
+			return ErrInvalidEventData
+		}
+		meta := &utils.RetinaMetadata{}
+		// Add packet size to the flow's metadata.
+		utils.AddPacketSize(meta, 128)
```
matmerr, line 271: same as magic number comment above

matmerr (changes requested):

windows/manifests/windows.yaml
```diff
@@ -66,5 +66,5 @@ data:
       host: "0.0.0.0"
       port: 10093
     logLevel: info
-    enabledPlugin: ["hnsstats"] 
+    enabledPlugin: ["hnsstats","ebfforwindows"]
```
matmerr, line 69:
```suggestion
    enabledPlugin: ["hnsstats","ebpfforwindows"]
```

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,277 @@
...
+func (p *Plugin) Compile(context.Context) error {
+	return nil
+}
+
+// Generate is a no-op for the ebpfwindows plugin
+func (p *Plugin) Generate(context.Context) error {
+	return nil
+}
+
+func (p *Plugin) handleTraceEvent(data unsafe.Pointer, size uint32) error {
```
anubhabMajumdar, line 227: How are we handling this error? Will this stop the plugin? If not, we would need to log the error so that we can debug the issue.

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,277 @@
...
+		return
+	}
+
+	if enricher.IsInitialized() {
+		p.enricher = enricher.Instance()
+	} else {
+		p.l.Warn("retina enricher is not initialized")
+	}
+
+	if p.enricher != nil {
```
anubhabMajumdar, line 171: We should check for `cfg.EnablePodLevel` before providing pod level metrics.

pkg/plugin/ebpfwindows/eventsmap_windows.go
```diff
@@ -0,0 +1,75 @@
...
+	ringBuffer uintptr
+}
+
+// NewEventsMap creates a new metrics map
+func NewEventsMap() EventsMap {
+	return &eventsMap{ringBuffer: 0}
+}
+
+// RegisterForCallback registers a callback function to be called when a new event is added to the events map
+func (e *eventsMap) RegisterForCallback(cb eventsMapCallback) error {
```
anubhabMajumdar, line 44: In my experience, under high traffic, the plugin would fail to keep up with the perf array in kernel. What's the mechanism here to track events overwritten in the ring-buffer? We should track that as we do here - https://github.com/microsoft/retina/blob/166256766a49dcbe48ffb3762b74401ff93ff6d9/pkg/plugin/packetparser/packetparser_linux.go#L685 .

Also, can have multiple handler instances running in parallel? If not, we can look into maintaining a queue in the user space and have multiple workers. Ref: https://github.com/microsoft/retina/blob/main/pkg/plugin/packetparser/packetparser_linux.go#L533-L541

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
```diff
@@ -0,0 +1,123 @@
...
+// NotifyCaptureHdr is the common header for all capture notifications
+type NotifyCaptureHdr struct {
+	NotifyCommonHdr
+	LenOrig uint32 // Length of original packet
+	LenCap  uint16 // Length of captured bytes
+	Version uint16 // Capture header version
+}
+
+// DropNotify is the notification for a packet drop
+type DropNotify struct {
```
anubhabMajumdar, line 48: About Drop/TraceNotify -> Are we saying we need these structs defined here because they are different from Cilium? The kernel events are different for windows ebpf? If that's the case, then why not name them something else?

windows/manifests/windows.yaml
anubhabMajumdar, line 69: Why should we enable both at the same time? They emit same metrics, so we would be overwriting value or double count metrics.

anubhabMajumdar (changes requested):

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,277 @@
...
+	eventType := *(*uint8)(data)
+	switch eventType {
+	case NotifyDrop:
+		if uintptr(size) < unsafe.Sizeof(DropNotify{}) {
+			p.l.Error("Invalid DropNotify data size", zap.Uint32("size", size))
+			return ErrInvalidEventData
+		}
+		e, err := p.parser.Decode(&observer.MonitorEvent{
+			Payload: &observer.PerfEvent{
+				Data: (*[unsafe.Sizeof(DropNotify{})]byte)(data)[:],
```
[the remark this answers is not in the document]
vpidatala94: yes I see *[unsafe.Sizeof(DropNotify{})]byte)(data)[:] this isn't necessary

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
```diff
@@ -0,0 +1,123 @@
...
+type TraceSockNotify struct {
+	Type       uint8
+	XlatePoint uint8
+	DstIP      IP
+	DstPort    uint16
+	SockCookie uint64
+	CgroupID   uint64
+	L4Proto    uint8
+	Ipv6       bool
+	Data       [128]byte
```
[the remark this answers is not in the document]
vpidatala94: Data       [128]byte --- contains the packet payload in the absence of perf-event array. To be replaced once perf-event-array is available.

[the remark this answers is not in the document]
vpidatala94: (interim)

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
```diff
@@ -0,0 +1,123 @@
...
+type DropNotify struct {
+	NotifyCaptureHdr
+	SrcLabel uint32
+	DstLabel uint32
+	DstID    uint32 // 0 for egress
+	Line     uint16
+	File     uint8
+	ExtError int8
+	Ifindex  uint32
+	Data     [128]byte
```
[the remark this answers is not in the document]
vpidatala94: (interim) Data [128]byte --- contains the packet payload in the absence of perf-event array. To be replaced once perf-event-array is available.

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"net"
+)
+
+// IP represents an IPv4 or IPv4 or IPv6 address
+type IP struct {
```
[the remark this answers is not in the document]
vpidatala94: testing and replcing with net.IP

windows/manifests/windows.yaml
vpidatala94: the correct solution would be to programmatically determine the underlying networking stack VPF(Legacy)/L3(new) as based on that either "hnsstats" or "ebfforwindows" should be enabled, that API is still being built so should I restore previous changes and only enable ebfforwindows during e2e comp testing?

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
vpidatala94: no, we are keeping them to validate whether size of dropNotify or traceNotify is appropriate size.

deploy/legacy/manifests/controller/helm/retina/values.yaml
```diff
@@ -35,9 +35,9 @@ image:
   # Overrides the image tag whose default is the chart appVersion.
   tag: "v0.0.2"
 
-enablePodLevel: false
+enablePodLevel: true
```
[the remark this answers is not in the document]
vpidatala94: [Intentional]
enablePodLevel: false
enablePodLevel: true --- we need this as I had highlighted as in the bug in retina windows, which is --- helm upgrade in windows doesn't perform a kube rollout causing plugins to never take in the config.

controller/Dockerfile
```diff
@@ -11,17 +11,24 @@ FROM --platform=$TARGETPLATFORM mcr.microsoft.com/cbl-mariner/base/core@sha256:7
...
 
 # intermediate go generate stage
-FROM golang AS intermediate 
+
+FROM golang AS eBPFRetinaStage
+WORKDIR /tmp
+RUN tdnf install -y unzip && \
+    curl -L -o eBPFRetina.zip https://www.nuget.org/api/v2/package/Microsoft.Wcn.Observability.eBPF.Retina.x64/0.1.0-prerelease.3 && \
+    mkdir eBPFRetina && \
+    unzip -d eBPFRetina eBPFRetina.zip || true
```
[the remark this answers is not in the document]
vpidatala94: unzip creates warning with exit code to be non zero, but this does not functionally affect the unzipping process, thus || true to override it.

pkg/plugin/ebpfwindows/ebpf_windows.go
vpidatala94: ins't enricher being initialed and enablepodlevel metrics both mutually inclusive?

pkg/plugin/ebpfwindows/ebpf_windows.go
vpidatala94: will not stop the plugin, we are logging the error here -- err := p.handleTraceEvent(data, size)
	if err != nil {
		p.l.Error("Error handling trace event", zap.Error(err))
		return -1
	}

pkg/plugin/ebpfwindows/ebpf_windows.go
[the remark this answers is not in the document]
vpidatala94: [interim code] --- to be replaced once perf-event-array code is checked in. Since Ring buffer code does not append packet data. we are placing it with the event-structures.

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,277 @@
...
+	p.pullMetricsAndEvents(ctx)
+	p.l.Info("Complete ebpfWindows plugin...")
+	return nil
+}
+
+// metricsMapIterateCallback is the callback function that is called for each key-value pair in the metrics map.
+func (p *Plugin) metricsMapIterateCallback(key *MetricsKey, value *MetricsValues) {
+	p.l.Debug("MetricsMapIterateCallback")
+	p.l.Debug("Key", zap.String("Key", key.String()))
+	p.l.Debug("Value", zap.String("Value", value.String()))
```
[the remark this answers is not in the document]
vpidatala94: removed -- p.l.Debug("MetricsMapIterateCallback")
	p.l.Debug("Key", zap.String("Key", key.String()))
	p.l.Debug("Value", zap.String("Value", value.String()))

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,277 @@
...
+	ebpfWindowsPath := programFiles + "\\ebpf-for-windows\\"
+	newPath := currPath + ";" + ebpfWindowsPath
+	if err := os.Setenv("PATH", newPath); err != nil {
+		return fmt.Errorf("error setting PATH environment variable: %v", err)
+	}
+
+	return nil
+}
+
+func (p *Plugin) pullMetricsAndEvents(ctx context.Context) {
```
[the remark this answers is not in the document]
vpidatala94: it was written this way because it was called in Start method which do not generally return anything

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,277 @@
...
+		&NoopIdentityGetter,
+		&NoopDNSGetter,
+		&NoopIPGetter,
+		&NoopServiceGetter,
+		&NoopLinkGetter,
+		&NoopPodMetadataGetter,
+	)
+
+	if err != nil {
+		p.l.Fatal("Failed to create parser", zap.Error(err))
```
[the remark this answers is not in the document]
vpidatala94: done

windows/manifests/windows.yaml
vpidatala94: the correct solution would be to programmatically determine the underlying networking stack VPF(Legacy)/L3(new) as based on that either "hnsstats" or "ebfforwindows" should be enabled, that API is still being built so should I restore previous changes and only enable ebfforwindows during e2e comp testing?

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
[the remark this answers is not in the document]
vpidatala94: [interim code] --- to be replaced once perf-event-array code is checked in. Since Ring buffer code does not append packet data. we are placing it with the event-structures.

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+// Generate is a no-op for the ebpfwindows plugin
+func (p *Plugin) Generate(context.Context) error {
+	return nil
+}
+
+func (p *Plugin) handleTraceEvent(data unsafe.Pointer, size uint32) error {
+	if uintptr(size) < unsafe.Sizeof(uint8(0)) {
+		return ErrInvalidEventData
```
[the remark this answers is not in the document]
vpidatala94: fixed this

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
vpidatala94: removed them

windows/manifests/windows.yaml
vpidatala94: proceeding to uncomment changes

pkg/plugin/ebpfwindows/metricsmap_windows.go
vpidatala94: replaced with sum()

pkg/plugin/ebpfwindows/metricsmap_windows.go
vpidatala94: replaced

pkg/plugin/ebpfwindows/ebpf_windows.go
vpidatala94: determine packet size programmatically

pkg/plugin/ebpfwindows/ebpf_windows.go
vpidatala94: determine packet size programmatically

windows/manifests/windows.yaml
vpidatala94: only enabling ebpfforwindows plugin in comp tests

pkg/plugin/ebpfwindows/metricsmap_windows.go
vpidatala94: my take is ---  we are testing this code-flow anyway using e2e comp tests

pkg/plugin/ebpfwindows/ebpf_windows.go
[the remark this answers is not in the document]
vpidatala94: replaced this by programmatically determining packet size

pkg/plugin/ebpfwindows/ebpf_windows.go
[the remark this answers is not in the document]
vpidatala94: far point, as discussed, the solution is 
version check:
The idea is programmatically ensure that the cilium plugin version and CNC referenced cilium version are identical to avoid crashes.  To open an issue to track this and close this at the time of CNC release, and not block this PR

deploy/legacy/manifests/controller/helm/retina/values.yaml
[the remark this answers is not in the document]
vpidatala94: removed -- not needed as in e2e comp tests, we are installing instead of upgrading retina for advanced metrics

pkg/plugin/ebpfwindows/eventsmap_windows.go
vpidatala94: @Vinod - could you please take a look.

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
vpidatala94: The original request was to avoid references to Cilium in the plugin (which is no longer necessary).
We do want to use upstream Cilium structures in the plugin, as there is a risk that they could go out of sync with CNC cilium structures. Thus, keeping a copy of CNC cilium structures is a safer approach 
@saxena-anurag @shankarseal

pkg/plugin/ebpfwindows/eventsmap_windows.go
shankarseal: This is a ring buffer. We will follow up with changes to consume perf buffer which has a drop callback in addition to sample callback. I also think that the efficiency has to be in the user mode code (Retina) to avoid dropped packets. Also tagging @mikeagun

mikeagun: In ebpf-for-windows we don't overwrite records, but drop records when full.

The perf event array map keeps a counter of lost records, so we have a count of lost writes.

We can also have a user-space queue as you mentioned if this is a problem.

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
shankarseal: I concur with Kartik here. I think it will cause un-necessary complications to pull in Cilium go modules in the new Windows Retina module (at least in the short term). We can revisit this in the long run.

pkg/plugin/ebpfwindows/metricsmap_windows.go
shankarseal: Given that this code path is executed in the e2e tests; let us move on by creatin an issue to add a component test for later.

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
vpidatala94: @Vinod could you please take a look at it

pkg/plugin/ebpfwindows/eventsmap_windows.go
vpidatala94: we are printing a log message at the caller of this function to provide context to the user -- 

err := eventsMap.RegisterForCallback(p.eventsMapCallback)
		if err != nil {
			p.l.Error("Error registering for events map callback", zap.Error(err))
			return
		}

controller/Dockerfile
```diff
@@ -11,17 +11,24 @@ FROM --platform=$TARGETPLATFORM mcr.microsoft.com/cbl-mariner/base/core@sha256:7
...
 # mcr.microsoft.com/windows/servercore:ltsc2022
 FROM --platform=$TARGETPLATFORM mcr.microsoft.com/windows/servercore@sha256:45952938708fbde6ec0b5b94de68bcdec3f8c838be018536b1e9e5bd95e6b943 AS ltsc2022
 
-
 # build stages
 
 # intermediate go generate stage
-FROM golang AS intermediate 
+
+FROM golang AS eBPFRetinaStage
```
matmerr, line 24: can you also include this stage in this Dockerfile, so azure pipelines can use it
https://github.com/microsoft/retina/blob/main/.pipelines/cg-pipeline.yaml#L245

it uses the `controller/Dockerfile.windows-native` image, usage here:
https://github.com/microsoft/retina/blob/main/windows/docker/DockerBuildModule.psm1#L37

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,265 @@
...
+			metrics.ForwardPacketsGauge.WithLabelValues(ingressLabel).Set(float64(value.Sum()))
+			metrics.ForwardBytesGauge.WithLabelValues(ingressLabel).Set(float64(value.BytesSum()))
+		}
+	}
+}
+
+// eventsMapCallback is the callback function that is called for each value  in the events map.
+func (p *Plugin) eventsMapCallback(data unsafe.Pointer, size uint32) int {
+	p.l.Debug("EventsMapCallback")
+	p.l.Debug("Size", zap.Uint32("Size", size))
```
anubhabMajumdar, line 124: Can we remove these? Will be too chatty to debug potential issues.

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	return nil
+}
+
+func (p *Plugin) pullMetricsAndEvents(ctx context.Context) {
+	eventsMap := NewEventsMap()
+	metricsMap := NewMetricsMap()
+
+	err := p.addEbpfToPath()
```
anubhabMajumdar, line 153: Question - instead of setting the path by relying on `ProgramFiles` environment variable, can we use the full path to the retinaebpf dll? Less chance of failure?

pkg/plugin/ebpfwindows/ebpf_windows.go
anubhabMajumdar: Stopping the plugin causes agent to stop and restart. Are you sure you would want to stop the plugin for all the error you encounter in this function? Some of them maybe intermittent or one-off.

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
anubhabMajumdar: I would push back against this. You are using the parser as-is from upstream, we should be able to pin to a particular version for the structs as well.

pkg/plugin/ebpfwindows/eventsmap_windows.go
anubhabMajumdar: We would need to surface the dropped events as metric. We do this for the Linux plugins. We need this here - otherwise how would we measure performance or missing metrics/flows.

pkg/plugin/ebpfwindows/ebpf_windows.go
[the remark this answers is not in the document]
anubhabMajumdar: If version check will come in future, how will you guarantee this won't have some kind of nil pointer exception? Seems dangerous to leave it like this?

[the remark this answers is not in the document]
anubhabMajumdar: Also, any comments on making this a library like I suggested?

pkg/plugin/ebpfwindows/ebpf_windows.go
vpidatala94: this is not for retinaebpfapi.dll, rather to add the path to ebpfapi.dll a dependency for retinaebpfapi.dll, it is located in C:\Program Files\ebpf-for-windows

controller/Dockerfile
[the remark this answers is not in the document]
vpidatala94: We can safely ignore this problem and the warning is seen consistently but this does not affect functionality

pkg/plugin/ebpfwindows/eventsmap_windows.go
vpidatala94: Acknowledged, can we take this as the next iteration? and create an issue for this because this would require another release of retinaebpapi.dll thus delay closing the PR?

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,277 @@
...
+func New(cfg *kcfg.Config) registry.Plugin {
+	return &Plugin{
+		l:   log.Logger().Named(name),
+		cfg: cfg,
+	}
+}
+
+// Init is a no-op for the ebpfwindows plugin
+func (p *Plugin) Init() error {
+	parser, err := hp.New(logrus.WithField("windowsEbpf", "parser"),
```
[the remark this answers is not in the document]
vpidatala94: p.l.Error("Failed to create parser", zap.Error(err)) here? Anything else needed

pkg/plugin/ebpfwindows/eventsmap_windows.go
vpidatala94: Filing a new issue

pkg/plugin/ebpfwindows/ebpf_windows.go
vpidatala94: Default log level is Info so it won't affect in regular cases. From our current debugging experience, these messages have been useful to check if the plugin is receiving callbacks from the kernel

controller/Dockerfile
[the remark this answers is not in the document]
vpidatala94: Sure we can add a check if .zip file exists before extracting it

pkg/plugin/ebpfwindows/eventsmap_types_windows.go
vpidatala94: With the new changes, we are proceeding to avoid any copy/duplication or cilium structures and rather use cilium package monitorapi to refer to the structures. We find the V1 structure compliant with our kernel checks

pkg/plugin/ebpfwindows/eventsmap_windows.go
vpidatala94: https://github.com/microsoft/retina/issues/1572

controller/Dockerfile
vpidatala94: done

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+}
+
+// eventsMapCallback is the callback function that is called for each value  in the events map.
+func (p *Plugin) eventsMapCallback(data unsafe.Pointer, size uint32) int {
+	p.l.Info("EventsMapCallback")
+	p.l.Info("Size", zap.Uint32("Size", size))
+	err := p.handleTraceEvent(data, size)
+	if err != nil {
+		p.l.Error("Error handling trace event", zap.Error(err))
```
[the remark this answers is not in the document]
vpidatala94: data is just a byte array, don't see much use in that, this was originally added to make sure if the packet data is appending to the back of the structure

pkg/plugin/ebpfwindows/ebpf_windows.go
vpidatala94: p.l.Error("Error setting PATH environment variable", zap.Error(err)) inside addEbpfToPath

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,258 @@
...
+			Payload: &observer.PerfEvent{
+				Data: perfData,
+			},
+		})
+		if err != nil {
+			return fmt.Errorf("could not convert event to flow: %w", err)
+		}
+		p.enricher.Write(e)
+		meta := &utils.RetinaMetadata{}
+		utils.AddPacketSize(meta, size-uint32(unsafe.Sizeof(monitor.DropNotify{})))
```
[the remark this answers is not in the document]
vpidatala94: this is to increment byte counter for advanced metrics forward and drop bytes
networkobservability_adv_forward,drop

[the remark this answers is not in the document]
vpidatala94: this is actual packet size , the packet is appended to the end of the event

pkg/plugin/ebpfwindows/metricsmap_windows.go
```diff
@@ -0,0 +1,207 @@
...
+type MetricsMap interface {
+	IterateWithCallback(IterateCallback) error
+}
+
+type metricsMap struct {
+}
+
+var (
+	// Load the retinaebpfapi.dll
+	retinaEbpfAPI = windows.NewLazyDLL("retinaebpfapi.dll")
```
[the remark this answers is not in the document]
vpidatala94: cannot do that, it causes syntax error -- 
windows.NewLazyDLL("retinaebpfapi.dll") (value of type *"golang.org/x/sys/windows".LazyDLL) is not constantcompiler[InvalidConstInit](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#InvalidConstInit)

pkg/plugin/ebpfwindows/eventsmap_windows.go
```diff
@@ -0,0 +1,78 @@
...
+func NewEventsMap() EventsMap {
+	return &eventsMap{perfBuffer: 0}
+}
+
+// RegisterForCallback registers a callback function to be called when a new event is added to the events map
+func (e *eventsMap) RegisterForCallback(cb eventsMapCallback) error {
+
+	eventsCallback = cb
+
+	fmt.Println("Attempting to register")
```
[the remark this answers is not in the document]
vpidatala94: done

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,258 @@
...
+	case monitorapi.MessageTypeTrace:
+		if size <= uint32(unsafe.Sizeof(monitor.TraceNotify{})) {
+			return fmt.Errorf("invalid size for TraceNotify %d", size)
+		}
+		e, err := p.parser.Decode(&observer.MonitorEvent{
+			Payload: &observer.PerfEvent{
+				Data: perfData,
+			},
+		})
+		p.enricher.Write(e)
```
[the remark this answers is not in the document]
vpidatala94: done

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,258 @@
...
+		}
+		e, err := p.parser.Decode(&observer.MonitorEvent{
+			Payload: &observer.PerfEvent{
+				Data: perfData,
+			},
+		})
+		if err != nil {
+			return fmt.Errorf("could not convert event to flow: %w", err)
+		}
+		p.enricher.Write(e)
```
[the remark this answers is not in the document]
vpidatala94: done

pkg/plugin/ebpfwindows/metricsmap_windows.go
```diff
@@ -0,0 +1,207 @@
...
+// outside or a ClusterIP service being accessed from inside the cluster).
+// If it's none of the above, we return UNKNOWN direction.
+var direction = map[uint8]string{
+	dirUnknown: "UNKNOWN",
+	dirIngress: "INGRESS",
+	dirEgress:  "EGRESS",
+	dirService: "SERVICE",
+}
+
+// Value must be in sync with struct metrics_key in <bpf/lib/common.h>
```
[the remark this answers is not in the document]
saxena-anurag: The BPF program has the same struct defined in a header file that it includes. For user mode code, the definition is a "copy" of the same struct as we can't include the header file. The same is true for other structs also (TRACE_NOTIFY, DROP_PNOTIFY, etc.), and also in the upstream code.

But I think we can remove this comment when we add code to handle versioning logic.

pkg/plugin/ebpfwindows/metricsmap_windows.go
vpidatala94: added unit tests in ebpf_windows_test.go

matmerr (approved):
approving from my side, @anubhabMajumdar also to sign off

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,279 @@
...
+		if err != nil {
+			p.l.Error("Error registering for events map callback", zap.Error(err))
+			return
+		}
+
+		defer func() {
+			p.l.Error("ebpfwindows plugin canceling", zap.Error(ctx.Err()))
+			eventsMap.UnregisterForCallback()
+		}()
+	}
```
ritwikranjan, line 175: I believe this whole thing is required only when EnablePodLevel  is true, It won't make sense to log the warning msg when `EnablePodLevel` is false

```suggestion
	if p.cfg.EnablePodLevel {
		if enricher.IsInitialized() {
			p.enricher = enricher.Instance()
		} else {
			p.l.Warn("retina enricher is not initialized")
		}

		if p.enricher != nil {
			err := eventsMap.RegisterForCallback(p.l, p.eventsMapCallback)
			if err != nil {
				p.l.Error("Error registering for events map callback", zap.Error(err))
				return
			}

			defer func() {
				p.l.Error("ebpfwindows plugin canceling", zap.Error(ctx.Err()))
				eventsMap.UnregisterForCallback()
			}()
		}
	}
```

pkg/plugin/ebpfwindows/eventsmap_windows.go
```diff
@@ -0,0 +1,84 @@
...
+	l.Info("Attempting to register")
+	// Convert the Go function into a syscall-compatible function
+	callback := syscall.NewCallback(eventsMapSysCallCallback)
+
+	// Call the API
+	ret, _, err := callRegisterEventsMapCallback(uintptr(callback), e.perfBuffer)
+
+	if ret != 0 {
+		l.Error("Error registering for events map callback")
+		return err
```
ritwikranjan, line 63:
```suggestion
		errors.Wrap(err, "Error registering for events map callback")
```

PR #1470, opened 2025-03-25
ebpfwindows e2e tests

vpidatala94:
Changes being introduced.

test\e2e\tools\event-writer : this is an test tool that helps generating mock trace and drop events using a bpf program based on customer inputs. The tools consists a BPF program and C++ Application that is used to load/unload and attach/update filter
Tool is intentionally building only for Release config variant for x64.

Additional changes in E2E with jobs
a. to load program,maps and pin it because retina ebpforwindows plugin expects cilium pre-loaded.
b. to validate windows bpf metrics advanced + basic.
1. collect pre-test basic metrics
2. create a non-hpc-pod
3. capture ifindex, interface of that non-hpc-pod and attach the bpf program to that interface
4. set filter for event-writer to trace and initiate curl from the non-hpc-pod
5. capture post-test BPF program and check if the metrics have increased.
6. check for the presence of advanced metrics
c. to unpin bpf program, maps at the end of tests

### DIFF
--- a/.github/workflows/e2e-test-event-writer.yml
+++ b/.github/workflows/e2e-test-event-writer.yml
@@ -1,0 +1,87 @@
+name: Build E2E Test Event Writer
+on:
+  pull_request:
+    branches: [main]
+  merge_group:
+    types: [checks_requested]
+  workflow_dispatch:
+
+
+permissions:
+  contents: read
+  id-token: write
+
+env:
+  EVENT_WRITER_PATH: "${{ github.workspace }}/test/e2e/tools/event-writer/"
+
+jobs:
+  retina-win-e2e-bpf-images:
+    name: Build E2E Test Event Writer
+    runs-on: windows-2022
+
+    strategy:
+      matrix:
+        platform: ["windows"]
+        arch: ["amd64"]
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4.2.2
+
+      - name: Az CLI login
+        uses: azure/login@v2
+        with:
+          client-id: ${{ secrets.AZURE_CLIENT_ID }}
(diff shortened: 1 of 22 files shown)


test/e2e/retina_e2e_test.go
vpidatala94, line 1: retina_e2e_test.go changes not picked up -- probably a consequence of splitting the PR - fixing this

.github/workflows/e2e-test-event-writer.yml
```diff
@@ -0,0 +1,87 @@
...
+            echo "TAG=$TAG" >> $env:GITHUB_ENV
+          }
+        env:
+          IS_MERGE_GROUP: ${{ github.event_name == 'merge_group' }}
+
+      - name: Build Images
+        shell: cmd
+        run: |
+          if "%IS_MERGE_GROUP%"=="true" (
+            az acr login -n ${{ vars.ACR_NAME }} && docker build --file .\Dockerfile --tag ${{ vars.ACR_NAME }}/${{ github.repository }}/e2e-test-event-writer:${{ env.TAG }} . && docker build --file .\Dockerfile --tag ${{ vars.ACR_NAME }}/${{ github.repository }}/e2e-test-event-writer:latest . && docker push ${{ vars.ACR_NAME }}/${{ github.repository }}/e2e-test-event-writer:${{ env.TAG }} && docker push ${{ vars.ACR_NAME }}/${{ github.repository }}/e2e-test-event-writer:latest
```
matmerr, line 83: can we put this in the docker build powershell cmdlet which is reserved for windows container commands?

https://github.com/microsoft/retina/blob/main/windows/docker/DockerBuildModule.psm1

PR #1471, opened 2025-03-25
fix: enable annotations

alexcastilio:
# Description

**Feature description**: when using option `enableAnnotation: true` advanced metrics should only be generated for pods annotated with `retina.sh: observe` or pods in a namespace with that annotation or pods in kube-system namespace.

**Bug**: When `enableAnnotation: true` advanced metrics are still being generated for all pods.

**Fix**:
- add filtering to metrics module based on IPs stored on its cache
- add `kube-system` to list of namespaces to get metrics from

## Related Issue

x

## Checklist

## Additional Notes

---

### DIFF
--- a/pkg/controllers/daemon/namespace/namespace_controller.go
+++ b/pkg/controllers/daemon/namespace/namespace_controller.go
@@ -71,6 +71,8 @@ func (r *NamespaceReconciler) Start(ctx context.Context) {
 		select {
 		case <-ticker.C:
 			ns := r.cache.GetAnnotatedNamespaces()
+			// Add kube-system namespace to the list of namespaces to be included in metrics
+			ns = append(ns, "kube-system")
 			r.l.Debug("Reconciling metrics module", zap.Any("namespaces", ns))
 			spec := (&api.MetricsSpec{}).
 				WithIncludedNamespaces(ns).
--- a/pkg/module/metrics/metrics_module.go
+++ b/pkg/module/metrics/metrics_module.go
@@ -108,6 +108,9 @@ type Module struct {
 
 	// pubsub subscription uuid
 	pubsubPodSub string
+
+	// resetRegistry is the function to reset the registry
+	resetRegistry func(*Module, []api.MetricsContextOptions)
 }
 
 func InitModule(ctx context.Context,
@@ -133,6 +136,7 @@ func InitModule(ctx context.Context,
 			dirtyPods:     common.NewDirtyCache(),
 			pubsubPodSub:  "",
 			daemonConfig:  conf,
+			resetRegistry: resetRegistry,
 		}
 	})
 
@@ -162,7 +166,7 @@ func (m *Module) Reconcile(spec *api.MetricsSpec) error {
 	m.updateNamespaceLists(spec)
 
 	if m.currentSpec == nil || !validations.MetricsContextOptionsCompare(m.currentSpec.ContextOptions, spec.ContextOptions) {
-		m.updateMetricsContexts(spec)
+		m.updateMetricsContexts(spec, m.resetRegistry)
 	}
 
 	m.currentSpec = spec
@@ -202,15 +206,8 @@ func (m *Module) updateNamespaceLists(spec *api.MetricsSpec) {
 	}
 }
 
-func (m *Module) updateMetricsContexts(spec *api.MetricsSpec) {
-	// clean old metrics from registry (remove prometheus collectors and remove map entry)
-	// reset the advanced metrics registry
-	for key, metricObj := range m.registry {
-		metricObj.Clean()
-		delete(m.registry, key)
-	}
-
-	exporter.ResetAdvancedMetricsRegistry()
+func resetRegistry(m *Module, contextOptions []api.MetricsContextOptions) {
+	m.registry = make(map[string]AdvMetricsInterface)
 
 	ctxType := remoteContext
 	if m.daemonConfig != nil && !m.daemonConfig.RemoteContext {
@@ -220,7 +217,7 @@ func (m *Module) updateMetricsContexts(spec *api.MetricsSpec) {
 		ctxType = localContext
 	}
 
-	for _, ctxOption := range spec.ContextOptions {
+	for _, ctxOption := range contextOptions {
 		switch {
 		case strings.Contains(ctxOption.MetricName, forward):
 			fm := NewForwardCountMetrics(&ctxOption, m.l, ctxType)
@@ -263,6 +260,19 @@ func (m *Module) updateMetricsContexts(spec *api.MetricsSpec) {
 	}
 }
 
+func (m *Module) updateMetricsContexts(spec *api.MetricsSpec, resetRegistry func(*Module, []api.MetricsContextOptions)) {
+	// clean old metrics from registry (remove prometheus collectors and remove map entry)
+	// reset the advanced metrics registry
+	for key, metricObj := range m.registry {
+		metricObj.Clean()
+		delete(m.registry, key)
+	}
+
+	exporter.ResetAdvancedMetricsRegistry()
+
+	resetRegistry(m, spec.ContextOptions)
+}
+
 func (m *Module) run(newCtx context.Context) {
 	if m.isRunning {
 		m.l.Warn("Metric module is already running. Cannot start again.")
@@ -291,10 +301,18 @@ func (m *Module) run(newCtx context.Context) {
 				m.RLock()
 				f := ev.Event.(*flow.Flow)
 				m.l.Debug("converted flow object", zap.Any("flow l4", f.IP))
-				for _, metricObj := range m.registry {
-					metricObj.ProcessFlow(f)
+
+				srcIP := net.ParseIP(f.GetIP().GetSource())
+				dstIP := net.ParseIP(f.GetIP().GetDestination())
+
+				if !m.daemonConfig.EnableAnnotations ||
+					m.daemonConfig.EnableAnnotations && (m.filterManager.HasIP(srcIP) || m.filterManager.HasIP(dstIP)) {
+					for _, metricObj := range m.registry {
+						metricObj.ProcessFlow(f)
+					}
 				}
 				m.RUnlock()
+
 			case *flow.LostEvent:
 				ev := ev.Event.(*flow.LostEvent)
 				// the number of lost events == the size of the ring buffer initialized.
--- a/pkg/module/metrics/metrics_module_test.go
+++ b/pkg/module/metrics/metrics_module_test.go
@@ -4,10 +4,14 @@ package metrics
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"sync"
 	"testing"
+	"time"
 
+	"github.com/cilium/cilium/api/v1/flow"
+	v1 "github.com/cilium/cilium/pkg/hubble/api/v1"
 	"github.com/cilium/cilium/pkg/hubble/container"
 	api "github.com/microsoft/retina/crd/api/v1alpha1"
 	"github.com/microsoft/retina/pkg/common"
@@ -17,6 +21,7 @@ import (
 	"github.com/microsoft/retina/pkg/log"
 	"github.com/microsoft/retina/pkg/managers/filtermanager"
 	"github.com/microsoft/retina/pkg/pubsub"
+	"github.com/microsoft/retina/pkg/utils"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
@@ -437,10 +442,11 @@ func TestModule_Reconcile(t *testing.T) {
 				},
 			},
 			m: &Module{
-				RWMutex:   &sync.RWMutex{},
-				registry:  make(map[string]AdvMetricsInterface),
-				l:         l,
-				moduleCtx: context.Background(),
+				RWMutex:       &sync.RWMutex{},
+				registry:      make(map[string]AdvMetricsInterface),
+				l:             l,
+				moduleCtx:     context.Background(),
+				resetRegistry: resetRegistry,
 			},
 			expectErr: false,
 		},
@@ -468,8 +474,9 @@ func TestModule_Reconcile(t *testing.T) {
 					"drop_count":    testDropMetric,
 					"forward_count": testForwardMetric,
 				},
-				moduleCtx: context.Background(),
-				l:         l,
+				moduleCtx:     context.Background(),
+				l:             l,
+				resetRegistry: resetRegistry,
 			},
 			expectErr: false,
 		},
@@ -507,8 +514,9 @@ func TestModule_Reconcile(t *testing.T) {
 					"forward_count": testForwardMetric,
 					"forward_bytes": testForwardMetricBytes,
 				},
-				moduleCtx: context.Background(),
-				l:         l,
+				moduleCtx:     context.Background(),
+				l:             l,
+				resetRegistry: resetRegistry,
 			},
 			expectErr: false,
 		},
@@ -546,8 +554,9 @@ func TestModule_Reconcile(t *testing.T) {
 					"forward_count": testForwardMetric,
 					"forward_bytes": testForwardMetricBytes,
 				},
-				moduleCtx: context.Background(),
-				l:         l,
+				moduleCtx:     context.Background(),
+				l:             l,
+				resetRegistry: resetRegistry,
 			},
 			expectErr: true,
 		},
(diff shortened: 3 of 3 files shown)


pkg/controllers/daemon/namespace/namespace_controller.go
```diff
@@ -71,6 +71,8 @@ func (r *NamespaceReconciler) Start(ctx context.Context) {
 		select {
 		case <-ticker.C:
 			ns := r.cache.GetAnnotatedNamespaces()
+			// Add kube-system namespace to the list of namespaces to be included in metrics
+			ns = append(ns, "kube-system")
```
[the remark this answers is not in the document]
alexcastilio: if not added, adv metrics for pods in kube-system are not generated correctly as they will are getting filtered out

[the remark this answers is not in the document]
SRodi: In order to reduce the volume of metrics and improve performance, would it be worth to treat kube-system the same way as all other namespaces, requiring the annotation to exist? Is there any specific reason why we want kube-system metrics enabled by default?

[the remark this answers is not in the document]
alexcastilio: The idea of this PR is to fix the bug, bringing the behavior back to what is expected, which is described in documentation like that: https://retina.sh/docs/Metrics/annotations

We can have a discussion about changing it in a separate thread.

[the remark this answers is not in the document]
anubhabMajumdar: No, let's edit the documentation instead. Users would probably want to filter out kube-system metrics in the first place.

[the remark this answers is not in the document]
MikeZappa87: This is a timer function, so 'kube-system' is appended each time?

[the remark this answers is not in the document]
alexcastilio: @MikeZappa87 cache internally will not add duplicates: https://github.com/microsoft/retina/blob/0c96b7b35205762f8a1e68ea2fec8e50b29301ba/pkg/controllers/cache/cache.go#L479-L484

Anyway, that's now removed, so kube-system is not monitored by default. @anubhabMajumdar Documentation changed accordingly.

vakalapa:
Can we have a testing done section ?

alexcastilio:
> Can we have a testing done section ?

@vakalapa PR description updated with test details.

pkg/module/metrics/metrics_module.go
```diff
@@ -108,6 +108,9 @@ type Module struct {
 
 	// pubsub subscription uuid
 	pubsubPodSub string
+
+	// resetRegistry is the function to reset the registry
+	resetRegistry func(*Module, []api.MetricsContextOptions)
```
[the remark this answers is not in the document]
alexcastilio: resetRegistry is a function that receives *Module and resets it's registry. Originally there's some logic in the function with its own dependencies that is not needed in the test. I don't want to consider m.daemonConfig, ctxType, naming logic and functions for creating specific metrics for resetting registry in my test.

There's a test to check generation of advanced metrics. For that, I want to validate the creation of metrics without following the original logic of resetting registry.  So I only want to provide a simple way of registry reset: https://github.com/microsoft/retina/blob/feffe34ac63aa07ec848a10ff2f5ec398e3b2ef0/pkg/module/metrics/metrics_module_linux_test.go#L774-L778

If it was a method, this would be way more complex.

pkg/module/metrics/metrics_module.go
```diff
@@ -263,6 +260,21 @@ func (m *Module) updateMetricsContexts(spec *api.MetricsSpec) {
...
+	m.Lock()
+	for key, metricObj := range m.registry {
+		metricObj.Clean()
+		delete(m.registry, key)
+	}
+	m.Unlock()
+
+	exporter.ResetAdvancedMetricsRegistry()
+
+	resetRegistry(m, spec.ContextOptions)
```
vakalapa, line 275: Will we reset the registry for anytime the annotation changes for a given pod or only if the context options and the additional labels change for a given metric ? 

because resetting is very heavy and would lose the old data

alexcastilio: Registry is reset when there's a change in api.MetricsSpec (where ContextOptions is): 
https://github.com/microsoft/retina/blob/a3911a86ab1450e9a3ebf3c7bda332ee5dcc0a41/pkg/module/metrics/metrics_module.go#L168C2-L170C3

vakalapa (approved):

pkg/module/metrics/metrics_module.go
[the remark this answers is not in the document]
matmerr: In my opinion, mixing and matching CRUD operation patterns here is a slight code smell, especially when reading `(m *Module) updateMetricsContexts` where we have both 

```
	exporter.ResetAdvancedMetricsRegistry()
...
	resetRegistry(m, spec.ContextOptions)
```

Performing reset operations. Can we not perform any noop dependency injection to match the rest of metricsmodule?

pkg/module/metrics/metrics_module.go
```diff
@@ -291,10 +317,38 @@ func (m *Module) run(newCtx context.Context) {
 				m.RLock()
 				f := ev.Event.(*flow.Flow)
 				m.l.Debug("converted flow object", zap.Any("flow l4", f.IP))
-				for _, metricObj := range m.registry {
-					metricObj.ProcessFlow(f)
+
+				srcIP := net.ParseIP(f.GetIP().GetSource())
```
rectified95: this can be moved to the else clause

rectified95:
Can we identify when this regressed? It's a pretty massive functionality bug this PR is addressing

pkg/module/metrics/metrics_module.go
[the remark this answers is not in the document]
rectified95: Mat's comment was to add this resetregistry function on the Module struct, I believe.
It's just a matter of unifying the calling pattern - call a method on a struct instance, which implicitly apsses a pointer to it, rather than passing the pointer explicitly as an invocation argument.

Doing that will allow us to remove the mrr indirection and the noisy changes from the PR - they make it a bit harder to get at the actual changes adderssing the bug.

pkg/module/metrics/metrics_module.go
```diff
@@ -291,10 +317,38 @@ func (m *Module) run(newCtx context.Context) {
...
+				dstIP := net.ParseIP(f.GetIP().GetDestination())
+
+				if !m.daemonConfig.EnableAnnotations {
+					for _, metricObj := range m.registry {
+						metricObj.ProcessFlow(f)
+					}
+				} else {
+					if m.filterManager.HasIP(srcIP) && m.filterManager.HasIP(dstIP) {
```
rectified95: Is this correct? I'd imagine that a flow show be reported even if just src OR dst pod is tagged.

pkg/module/metrics/metrics_module.go
```diff
@@ -291,10 +317,38 @@ func (m *Module) run(newCtx context.Context) {
...
+						for _, metricObj := range m.registry {
+							metricObj.ProcessFlow(f)
+						}
+					} else {
+						// TODO: This is a temporary fix. Metrics module should not change the flow struct.
```
rectified95, line 330: what is the correct fix and do we have a plan to make it?

pkg/module/metrics/metrics_module.go
```diff
@@ -291,10 +317,38 @@ func (m *Module) run(newCtx context.Context) {
...
+						if !m.filterManager.HasIP(srcIP) && m.filterManager.HasIP(dstIP) {
+							f.IP.Source = ""
+							f.Source = nil
+							for _, metricObj := range m.registry {
```
rectified95: this can be refactored to avoid repeating the processflow calls - modify the flow object as needed first, then call processflow

rectified95: Also, that will make the if clause redundant (if both src and dst are there, no mods needed, just passthrough to the call line)

pkg/module/metrics/metrics_module.go
alexcastilio: That's what's happening under the else clause.

When there are flows that only Source or Destination are supposed to generate metrics, the side that does not generate metrics is deleted, so only metrics regarding the pod of interest are generated.

Another scenario is API Server, which many (all?) pods communicate with, so even pods without annotation would be monitored and would have network metrics generated. This would lead to confusion on why many metrics are generated for pods that are not annotated.

pkg/module/metrics/metrics_module.go
alexcastilio: I think there are some options to be analyzed here which might require some design discussions and refactoring. This is out of scope of this PR.

alexcastilio:
> Can we identify when this regressed? It's a pretty massive functionality bug this PR is addressing

I searched for it but could not identify regression.

pkg/module/metrics/metrics_module.go
rectified95: I'm not sure if this is out of scope, if the thing hasn't worked in a long time, we're not under pressure to make a hacky fix.
Can you share the options here?

rectified95: The way we do it in Cilium is the filters are applied in each metric handler, rather than flow being altered.
I think it would be good to spell out why it's wrong for this module to modify the flow.
I don't have a strong objection in merging it this way, though.

pkg/module/metrics/metrics_module.go
```diff
@@ -291,10 +317,34 @@ func (m *Module) run(newCtx context.Context) {
 				m.RLock()
 				f := ev.Event.(*flow.Flow)
 				m.l.Debug("converted flow object", zap.Any("flow l4", f.IP))
-				for _, metricObj := range m.registry {
-					metricObj.ProcessFlow(f)
+
+				if !m.daemonConfig.EnableAnnotations {
+					for _, metricObj := range m.registry {
```
rectified95, line 322: Thanks for removing the duplicate ProcessFlow calls. This can be further simplified: if CRD and annotations are mutually exclusive, then this if-clause is not needed.
ProcessFlow is called both when enableAnnotations is true or false. But in the latter case, we make some changes to the Flow object first.

```
if filtermgr[src] || filtermgr[dst]
  srcIP :=
  dstIP :=
  // fix up the Flow obj as needed

for m in metrics:
  m.ProcessFlow   
```

pkg/module/metrics/metrics_module.go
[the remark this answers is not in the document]
rectified95: @matmerr Did I read your comment right? I think this new interface is not needed, and we can just make the function work on this type.

pkg/module/metrics/metrics_module.go
rectified95: Reading through the code, I'm realizing the filterManager is shared for both the annotation and crd paths. My suggestion would potentially erase src/dst from Flow if annotations are disabled, but the CRD populated the filterManager.

I do wonder though, since annotation/crd are just two ways of populating the FilterManager, if annotations don't work, then can you please check the CRD does, as a sanity check?

rectified95:
Can you edit the PR description to demonstrate the before/after of the fix,? Currently it only shows that for applying an annotation; the current before annotation screenshot doesn't include kube-system, or any other namespace, so it doesn't show that metrics are collected for "all" namespaces".
For better readability, I'd just include a snippet of the metric output showing the relevant namespaces, no need to apste the whole thing.

pkg/module/metrics/metrics_module.go
```diff
@@ -291,10 +317,34 @@ func (m *Module) run(newCtx context.Context) {
...
+						metricObj.ProcessFlow(f)
+					}
+				} else {
+					srcIP := net.ParseIP(f.GetIP().GetSource())
+					dstIP := net.ParseIP(f.GetIP().GetDestination())
+
+					if m.filterManager.HasIP(srcIP) || m.filterManager.HasIP(dstIP) {
+						// TODO: This is a temporary fix. Metrics module should not change the flow struct.
+						if !m.filterManager.HasIP(srcIP) {
+							f.IP.Source = ""
```
rectified95, line 332: Could you explain why this actually fixes the issue? Why doesn't the filter manager respect the pod/namespaces added to it?

pkg/module/metrics/metrics_module.go
alexcastilio: Metrics module is 1. monitoring events to update cache with pods/namespaces of interest according to annotations/CRD configuration and 2. reading flows from Enricher and calling ProcessFlow for each Advanced Metric in its registry. Modifying the flow does not seem to fit in here in my opinion.

One option I see is to provide data about annotated pods/namespaces to eBPF plugins to generate flows only with data for pods of interest. E.g. a flow would be generated with source data and without destination data in case of source pod annotated and destination pod not annotated. 
I have the impression that this was the initial implementation thought because there's some logic for adding annotated pods to filter manager and to some filter map in kernel space: https://github.com/microsoft/retina/blob/0854cd491fafd0b479e30a0513fb029a9d266e26/pkg/module/metrics/metrics_module.go#L492-L530
https://github.com/microsoft/retina/blob/0854cd491fafd0b479e30a0513fb029a9d266e26/pkg/managers/filtermanager/manager_linux.go#L62-L94
I could not dig too much into that. It's not clear for me if this code is doing something else or if there is something wrong.

Another approach would be to provide details about pods of interest to each metric handler and let it generate metrics accordingly.

If we want to have the flow modified somewhere in the middle of this process, we need to add some component that is clearly doing so that developer following the flow processing path will easily know where this logic is

pkg/module/metrics/metrics_module.go
alexcastilio: Metrics Handlers just generate metrics based on the Flow struct provided. The only filtering it does is what labels to add depending on Retina's context configuration (direction, podname, namespace, port), not depending on pods details (annotations/CRD config). Metrics Module was only sending the flow received from Enricher to all Advanced Metric Handlers. Enricher is enriching flows with pod data cached (for both source and destination). Finally each flow was generated with source and destination IPs. So it seems that this control of metrics based on pods of interest was not happening anywhere in this path.

There are 2 details happening in the fix here:
1. Flows that both Source and Destination are not in Filter Manager are not processed any more.
2. Flows that only Source or Destination pod are in Filter Manager would still generate metrics for the pair pod that's not in Filter manager if they were processed straight away, because Metric Handlers would add labels according to both Source and Destination enriched data. That's why flow is edited to delete details of Source/Destination that is not in Filter Manager.

alexcastilio:
> Can you edit the PR description to demonstrate the before/after of the fix,? Currently it only shows that for applying an annotation. For better readability, I'd just include a snippet of the metric output showing the relevant namespaces, no need to apste the whole thing.

@rectified95 PR description updated

pkg/module/metrics/metrics_module.go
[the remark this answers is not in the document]
alexcastilio: @rectified95 The problem is to mock it in tests. By having that method provided by an inner struct field, I can replace it by a mock implementation in the test through Dependency Injection. I can test only what I want (specific Metrics Module logic regarding pods/namespace annotations) without bringing all Reset complexity (go through context options, check naming and create metrics with such logic) and dependencies to a test in which that's not relevant.

pkg/module/metrics/metrics_module.go
rectified95: See my comment above about the filter map. We should focus on that, and see why these flows are created from packets we don't care about. The condition in ebpf is "we don't care about the packet if src AND dst are in filter map". Perhaps we should change that to OR?. Let's continue the discussion in the other comment thread.

pkg/module/metrics/metrics_module.go
rectified95: > provide to eBPF plugins to generate flows only with data for pods of interest

We already do that, and if that doesn't work, we should investgate why.
The `retina_filter` map is populated with IPs of interest by FilterManager.AddIPs:
```
// Bulk add the IPs to the filter map.
// This is more efficient than adding one IP at a time.
err := f.fm.Add(ipsToAdd)
```
Then, `packetparser` calls `lookup` on the map and skips packets if both src and dst aren't present in the map.
```
#if BYPASS_LOOKUP_IP_OF_INTEREST == 0
if (!lookup(p.src_ip) && !lookup(p.dst_ip))
{
	return;
}
#endif
```

rectified95: @alexcastilio Can you try deploying Retina with this Helm value `bypassLookupIPOfInterest` set to true, **without** your PR and seeing if that's enough to make annotation work?

PR #1588, opened 2025-05-08
fix: include Mariner check for Hubble scenario

rectified95:
Fixes https://github.com/microsoft/retina/pull/1458

In the PR above, I only added the `/etc/host-os-release` volumeMount in 
`deploy/standard/manifests/controller/helm/retina/templates/daemonset.yaml`
but not in 
`deploy/hubble/manifests/controller/helm/retina/templates/agent/daemonset.yaml`

The omission was because in the former we iterate over the volume mount list defined in values.yaml, while the latter defines them explicitly in the daemonset.

### DIFF
--- a/deploy/hubble/manifests/controller/helm/retina/templates/agent/daemonset.yaml
+++ b/deploy/hubble/manifests/controller/helm/retina/templates/agent/daemonset.yaml
@@ -120,6 +120,8 @@ spec:
               mountPath: /sys/kernel/tracing
             - name: cilium
               mountPath: /var/run/cilium
+            - name: host-os-release
+              mountPath: /etc/os-release  
           {{- if .Values.hubble.tls.enabled }}
             - name: tls
               mountPath: /var/lib/cilium/tls/hubble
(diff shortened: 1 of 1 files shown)


deploy/hubble/manifests/controller/helm/retina/templates/agent/daemonset.yaml
```diff
@@ -146,6 +148,10 @@ spec:
         hostPath:
           path: /var/run/cilium
           type: DirectoryOrCreate
+      - name: host-os-release
+        hostPath:
+          path: /etc/os-release
+          type: FileOrCreate
```
alexcastilio, line 154: why is this new volume needed?

rectified95: context in https://github.com/microsoft/retina/pull/1458
We need to read the file on the host - it's the most reliable way to get distro info. All major distros have it, and I confirmed the string values will be stable with the Azure Linux team.
If the file is empty, our logic handles that fine, and says 'not mariner', which is correct, because Mariner will positively have this file.
FileOrCreate is there to avoid pod hanging in init state on some esoteric distro that misses this

alexcastilio (approved):

vakalapa (approved):

PR #1600, opened 2025-05-14
fix(fips): Grant permissions to support FIPS on Ubuntu

rectified95:
When running on the FIPS-compliant Ubuntu 20.04, Retina requires `SYS_RESOURCE` on top of `IPC_LOCK`.

Also, skip attaching to unavailable kernel hook points.

Merge after https://github.com/microsoft/retina/pull/1601 - otherwise the pod still fails, just at a later stage.

### DIFF
--- a/deploy/hubble/manifests/controller/helm/retina/values.yaml
+++ b/deploy/hubble/manifests/controller/helm/retina/values.yaml
@@ -108,9 +108,9 @@ securityContext:
   capabilities:
     add:
       - SYS_ADMIN
-      - SYS_RESOURCE
+      - SYS_RESOURCE # for setting rlimit
       - NET_ADMIN # for packetparser plugin
-      - IPC_LOCK # for mmap() calls made by NewReader(), ref: https://man7.org/linux/man-pages/man2/mmap.2.html
+      - IPC_LOCK # for mmap() calls made by NewReader(), ref: https://man7.org/linux/man-pages/man2/mmap.2.html 
   windowsOptions:
     runAsUserName: "NT AUTHORITY\\SYSTEM"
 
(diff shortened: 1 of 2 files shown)


pkg/plugin/dropreason/dropreason_linux.go
```diff
@@ -431,21 +432,31 @@ func (dr *dropReason) attachKprobes(kprobes, kprobesRet map[string]*ebpf.Program
 	for name := range kprobes {
 		progLink, err := link.Kprobe(name, kprobes[name], nil)
 		if err != nil {
-			dr.l.Error("Failed to attach kprobe", zap.String("program", name), zap.Error(err))
-			return fmt.Errorf("Failed to attach program: %w", err) //nolint:goerr113 //wrapping error from external module
+			if errors.Is(err, os.ErrNotExist) {
```
anubhabMajumdar, line 435: I am surprised that we changed the behavior of error handling by putting it in a loop. Can there be a situation where we do not attach any `kprobe` or `kretprobe` and report `dropreason` as successfully running and not generate any metrics. I think we should make some of the `kprobes` necessary and not optional (ex: the IPTable rule drop probe).

rectified95: Current behavior would never result in no probes attached and plugin showing up as successful: in fact the error handling is stricter than before my fexit change, and we error out in case of any failure - I will explain why this should be fine below.

This PR brings back the skipping logic we had before, because it was in your fips PR, but I think we should actually not need it.

The code iterates over different sets of programs depending on the scenario, informed by testing in different environments; we should never even enter this codepath. _We now understand which functions are available on each SKU/ARCH and have corresponding data types defined to prevent crashing on loading._ 

These attach warning were initially created to handle Mariner not allowing kprobes into kernel module functions, but now we don't even load those programs for Mariner, since fexit fails to even load for these.

rectified95: I'm moving this to the other PR, to keep this one scoped to the permissions.

anubhabMajumdar (approved):

PR #1601, opened 2025-05-14
fix(DropReason): Do not load fexit programs on kernels below 5.5

rectified95:
In #1458 I added a kernel version check to only attach `fexit` programs when they're supported.
We also need to ensure we don't even load them into the kernel.

### DIFF
--- a/pkg/plugin/dropreason/dropreason_linux.go
+++ b/pkg/plugin/dropreason/dropreason_linux.go
@@ -139,7 +139,7 @@ func (dr *dropReason) Init() error {
 		return err //nolint:wrapcheck // no additional context needed
 	}
 
-	objs, maps, isFexit, err := dr.getEbpfPayload()
+	objs, maps, supportsFexit, err := dr.getEbpfPayload()
 	if err != nil {
 		return err
 	}
@@ -167,14 +167,10 @@ func (dr *dropReason) Init() error {
 	progsKprobe, progsKprobeRet := buildKprobePrograms(objs)
 	progsFexit := buildFexitPrograms(objs)
 
-	if dr.cfg.EnablePodLevel {
-		err = dr.attachKprobes(progsKprobe, progsKprobeRet)
+	if supportsFexit {
+		err = dr.attachFexitPrograms(progsFexit)
 	} else {
-		if isFexit {
-			err = dr.attachFexitPrograms(progsFexit)
-		} else {
-			err = dr.attachKprobes(progsKprobe, progsKprobeRet)
-		}
+		err = dr.attachKprobes(progsKprobe, progsKprobeRet)
 	}
 
 	dr.metricsMapData = maps.RetinaDropreasonMetrics
--- a/pkg/plugin/dropreason/dropreason_linux.go
+++ b/pkg/plugin/dropreason/dropreason_linux.go
@@ -143,14 +143,34 @@ func (dr *dropReason) Init() error {
 	var objs interface{}
 	maps := &kprobeMaps{}
 	isMariner := plugincommon.IsAzureLinux()
+	dr.l.Info("Distro check:", zap.Bool("isMariner", isMariner))
 
-	if !isMariner {
+	var kv semver.Version
+	kv, err = version.GetKernelVersion()
+	if err != nil {
+		dr.l.Warn("Failed to get kernel version", zap.Error(err))
+
+		kv, err = plugincommon.GetKernelVersionMajMin()
+		if err != nil {
+			return fmt.Errorf("Failed to get kernel version: %w", err) //nolint:goerr113 //wrapping error from external module
+		}
+	}
+	dr.l.Info("Detected kernel >= ", zap.String("version", kv.String()))
+
+	minVersionAmd64, _ := versioncheck.Version("5.5")
+	minVersionArm64, _ := versioncheck.Version("6.0")
+	isMinVer := (runtime.GOARCH == "amd64" && kv.GTE(minVersionAmd64)) || (runtime.GOARCH == "arm64" && kv.GTE(minVersionArm64))
+
+	switch {
+	case !isMinVer:
+		objs = &kprobeObjectsOld{} //nolint:typecheck // this is a generated struct
+		maps = &objs.(*kprobeObjectsOld).kprobeMaps
+	case !isMariner:
 		objs = &kprobeObjects{} //nolint:typecheck // this is a generated struct
 		maps = &objs.(*kprobeObjects).kprobeMaps
-	} else {
+	default:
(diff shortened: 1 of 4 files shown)


pkg/plugin/dropreason/ebpfsetup_linux.go
```diff
@@ -0,0 +1,204 @@
...
+- nf_nat_inet_fn
+*/
+
+func (dr *dropReason) getEbpfPayload() (objs interface{}, maps *kprobeMaps, isFexit bool, err error) {
+	isMariner := plugincommon.IsAzureLinux()
+	dr.l.Info("Distro check:", zap.Bool("isMariner", isMariner))
+
+	kv, err := version.GetKernelVersion()
+	if err != nil {
+		dr.l.Warn("Failed to get kernel version", zap.Error(err))
```
anubhabMajumdar, line 70: This is not needed if fallback `GetKernelVersionMajMin` succeeds.

pkg/plugin/dropreason/ebpfsetup_linux.go
```diff
@@ -0,0 +1,204 @@
...
+
+	objs, maps, isFexit = resolvePayload(runtime.GOARCH, kv, isMariner)
+	return objs, maps, isFexit, nil
+}
+
+func resolvePayload(arch string, kv semver.Version, isMariner bool) (interface{}, *kprobeMaps, bool) {
+	minVersionAmd64, _ := versioncheck.Version(MinAmdVersionNum)
+	minVersionArm64, _ := versioncheck.Version(MinArmVersionNum)
+
+	isFexit := (arch == "amd64" && kv.GTE(minVersionAmd64)) ||
```
anubhabMajumdar, line 86: Question - no chance of case mismatch right?

pkg/plugin/dropreason/ebpfsetup_linux.go
```diff
@@ -0,0 +1,204 @@
...
+		(arch == "arm64" && kv.GTE(minVersionArm64))
+
+	var objs interface{}
+	var maps *kprobeMaps
+
+	switch {
+	case !isFexit:
+		objs = &kprobeObjectsOld{} //nolint:typecheck // this is a generated struct
```
anubhabMajumdar, line 94: Can you rename `kprobeObjectsOld` to something more verbose and meaningful?

pkg/plugin/dropreason/ebpfsetup_linux.go
```diff
@@ -0,0 +1,204 @@
...
+	minVersionArm64, _ := versioncheck.Version(MinArmVersionNum)
+
+	isFexit := (arch == "amd64" && kv.GTE(minVersionAmd64)) ||
+		(arch == "arm64" && kv.GTE(minVersionArm64))
+
+	var objs interface{}
+	var maps *kprobeMaps
+
+	switch {
+	case !isFexit:
```
anubhabMajumdar, line 93: Can this be refactored to be positive check? Ex:
```
switch {
case fexit:
objs = &fexitObjects{}
...
case Mariner:
objs = &marinerObjects{}
... 
}
```

I think this is more readable.

pkg/plugin/dropreason/dropreason_linux.go
```diff
@@ -184,21 +170,7 @@ func (dr *dropReason) Init() error {
...
-			if err != nil {
-				return fmt.Errorf("Failed to get kernel version: %w", err) //nolint:goerr113 //wrapping error from external module
-			}
-		}
-		dr.l.Info("Detected kernel >= ", zap.String("version", kv.String()))
-
-		minVersionAmd64, _ := versioncheck.Version("5.5")
-		minVersionArm64, _ := versioncheck.Version("6.0")
-		if (runtime.GOARCH == "amd64" && kv.GTE(minVersionAmd64)) || runtime.GOARCH == "arm64" && kv.GTE(minVersionArm64) {
+		if isFexit {
```
anubhabMajumdar, line 173: Make it `else if`?

pkg/plugin/dropreason/dropreason_linux.go
```diff
@@ -143,14 +143,34 @@ func (dr *dropReason) Init() error {
...
+	kv, err = version.GetKernelVersion()
+	if err != nil {
+		dr.l.Warn("Failed to get kernel version", zap.Error(err))
+
+		kv, err = plugincommon.GetKernelVersionMajMin()
+		if err != nil {
+			return fmt.Errorf("Failed to get kernel version: %w", err) //nolint:goerr113 //wrapping error from external module
+		}
+	}
+	dr.l.Info("Detected kernel >= ", zap.String("version", kv.String()))
```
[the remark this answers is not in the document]
rectified95: the fallback is for MS distros that have a 4-piece version string that trips up Cilium's parsing lib.
In my function, I only care about the Maj.Min and skip all the weird cases past minor, so technically we detect to this approximation

pkg/plugin/dropreason/ebpfsetup_linux.go
rectified95: If you mean eg. 'AMD64', I think these are standardized. Our build would break too, because we rely on this string

pkg/plugin/dropreason/ebpfsetup_linux.go
rectified95: removed

pkg/plugin/dropreason/ebpfsetup_linux.go
rectified95: renamed, and restructured the other types

pkg/plugin/dropreason/ebpfsetup_linux.go
rectified95: done

pkg/plugin/dropreason/dropreason_linux.go
rectified95: made into else if

pkg/plugin/dropreason/ebpfsetup_linux.go
```diff
@@ -61,47 +61,49 @@ module funcs:
...
 
 	switch {
-	case !isFexit:
-		objs = &kprobeObjectsOld{} //nolint:typecheck // this is a generated struct
-		maps = &objs.(*kprobeObjectsOld).kprobeMaps
-	case !isMariner:
-		objs = &fexitObjects{} //nolint:typecheck // this is a generated struct
-		maps = &objs.(*fexitObjects).kprobeMaps
-	case isMariner:
+	case isPodLevel:
```
anubhabMajumdar, line 92: Add a TODO about why we are excluding Pod level from running fexit.

anubhabMajumdar (approved):

anubhabMajumdar (approved):

PR #1614, opened 2025-05-16
fix: e2e job skipped due to yaml syntax error

rectified95:
Our E2e job is getting skipped in the merge queue, because of YAML parsing errors.
We have invalid syntax in the if-statements checking for success of prerequisite jobs that need to run for `manifests` and `E2E` to run. 
Looks like this used to be ignored and those jobs just ran, while recently Github started enforcing stricter YAML syntax validation which is causing these jobs to get skipped! 
It started about a week ago, example errors at the bottom of the page:
https://github.com/microsoft/retina/actions/runs/14842770818
image

### DIFF
--- a/.github/workflows/images.yaml
+++ b/.github/workflows/images.yaml
@@ -264,8 +264,8 @@ jobs:
           IS_MERGE_GROUP: ${{ github.event_name == 'merge_group' }}
 
   manifests:
-    if: ${{ github.event_name == 'merge_group' && success('retina-images') && success('retina-win-images') && success('operator-images') && success('retina-shell-images')}}
     name: Generate Manifests
+    if: ${{ github.event_name == 'merge_group' }}
     runs-on: ubuntu-latest
     needs:
       [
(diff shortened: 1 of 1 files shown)


.github/workflows/images.yaml
```diff
@@ -264,8 +264,8 @@ jobs:
           IS_MERGE_GROUP: ${{ github.event_name == 'merge_group' }}
 
   manifests:
-    if: ${{ github.event_name == 'merge_group' && success('retina-images') && success('retina-win-images') && success('operator-images') && success('retina-shell-images')}}
     name: Generate Manifests
+    if: ${{ github.event_name == 'merge_group' }}
```
nddq, line 268: does this mean that we will run this step regardless of whether the images were successfully built or not?

rectified95: Nope, per GH docs, the `needs` clause means a dependency must have succeeded, our success checks have always been redundant.
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

rectified95: This PR does not change the issue that GH skips all downstream jobs if a dependency fails or is skipped, and re-run do not re-evaluate jobs with if-conditions. 
This has come up recently when WIndows MCR endpoints have been unstable.
I'm planning to an unconditional guard job to check if E2E has run.

nddq (approved):

PR #1628, opened 2025-05-27
feat(packetparser): Allow sampling of packets

mmckeen:
# Description

For nodes with lots of forwarded packets `packetparser` can easily get overwhelmed and utilize a large amount of resources.

This allows for optional sampling of `packetparser` packets with best-effort weighting of metrics based on observed packets.  This only applies to `dataAggregationLevel: high`.

There's only a minor change in the existing behavior when the default sampling rate of 1 is applied, this is that we only look at the current packet for deciding when to report for important TCP flags.  Previously we would report if any important flag had been seen on the connection.  In practice this would mean we'd always report packets since `SYN` was considered an important flag.  Since we also take into account whether or not any new flags are seen on the connection the number of packets now being reported at the reporting interval versus always as before should be minimal.

## Checklist

At sampling rate of `100`, forwarded bytes and packets remain similar to those reported by cadvisor.

Screenshot 2025-05-27 at 1 52 55 PM
Screenshot 2025-05-27 at 1 49 37 PM

This correlation gets stronger the lower the sampling rate (below at `2`).

Screenshot 2025-05-27 at 2 24 54 PM
Screenshot 2025-05-27 at 2 25 14 PM

## Additional Notes

I was unsure whether or not reporting packets for important TCP flags across the entire lifetime of the connection was intentional or not.  Without looking only at the current packet sampling doesn't work.  I can change the logic to keep the existing behavior for non-sampled packets if it makes sense.

---


SRodi (commented):
Thanks so much for this contribution and the detailed context, @mmckeen! The sampling feature is a great idea for improving resource utilization on nodes with high traffic. I noticed there are also some improvements to existing packet reporting logic bundled with the new implementation. To keep reviews focused and changes easier to track, would you be open to splitting the new sampling functionality and the general code improvements into two separate PRs? This would help us ensure each change gets the attention it deserves.

Really appreciate your work and your openness to feedback. Let us know how we can support you with the split—happy to help however we can! cc @nddq

SRodi:
> I can also split the metric weighting and counting of packets between reports into that PR so it's isolated, let me know what makes sense 🙇.

Yes, that makes sense. Thanks. cc @mmckeen

PR #1665, opened 2025-06-06
fix(packetparser): Fix under reporting of TCP flags and packet metrics, improve scalability

mmckeen:
# Description

Previously `packetparser` in `high` `dataAggregationLevel` would report (mostly) every single packet since important flags were observed over the lifetime of the connection.

This changes that behavior to only observe the important flags on individual packets and report when necessary.

This will mean less packets are reported. However, it also adds back weighting for bytes, packets, and TCP flags so that metrics remain accurate versus before.

I also noticed the current docs for the TCP flags metrics are inaccurate, we only report a subset of the supported flags.  Not sure if this is intentional, however supporting more flags will put more memory pressure on both conntrack as well as performance pressure on packet reporting.  With sampling in place, this should be more than worth it but there may be repercussions for the performance of `low` `dataAggregationLevel`.

## Checklist

eBPF objects compile and load as expected.

# `main` Branch

tcpflags main
prometheus packets retina main
prometheus bytes retina main

# This Branch

tcpflags patched
prometheus packets retina patched
prometheus bytes retina patched

## Additional Notes

https://github.com/microsoft/retina/pull/1628 will be a follow-up to this to add additional sampling functionality.

---


pkg/plugin/conntrack/_cprog/conntrack.h
```diff
@@ -26,7 +26,7 @@
 // Define how long a TCP connection should be kept alive after receiving the first SYN
 #define CT_SYN_TIMEOUT 60
 // Define the interval at which a packet should be sent to the userspace
-#define CT_REPORT_INTERVAL 30
+#define CT_REPORT_INTERVAL 1
```
mmckeen, line 29: At least in my testing this didn't make much of a difference.  Seems there are lots of short-lived connections that rely on reporting on connection teardown.

mmckeen:
@nddq @SRodi this is ready for review 🙇

nddq:
@mmckeen sorry for the delay, I just got back from a break. I’ve gone through your proposed change a couple of times, and it looks solid to me. In fact, it addresses something we initially overlooked when conntrack was introduced. First, a few points to make sure we're aligned:

> Previously packetparser in high dataAggregationLevel would report (mostly) every single packet since important flags were observed over the lifetime of the connection.

That’s correct. For any given packet, we currently:
- Report it if it contains a flag set we haven’t seen before for this specific connection
- Report it if a certain amount of time has passed since the last reported packet for this connection (default: 30 seconds; applies to both dataAggregation levels)
- Otherwise, we skip it

In a typical TCP connection, the reported events would likely look like:
SYN, SYN-ACK, ACK, PSH, PSH-ACK, (30 secs), PSH, PSH-ACK, ... FIN, FIN-ACK, FIN-ACK

As a result, we ignore all packets during those 30-second windows, which skews the reported packet, byte, and TCP flag counts from the actual values.

That said:

> This changes that behavior to only observe the important flags on individual packets and report when necessary.

Could you clarify this part? From what I see, conntrack already behaves this way today — so I’m not sure this change introduces new behavior?

> This will mean less packets are reported. However, it also adds back weighting for bytes, packets, and TCP flags so that metrics remain accurate versus before.

This is great — it addresses the gap I mentioned earlier around ignored packets. So in that sense, this feels more like a bug fix than a new feature 🙂

mmckeen:
@nddq @srodi any further thoughts on this?

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -31,23 +31,50 @@ struct conntrackmetadata {
     __u32 packets_rx_count;
 };
 
+/**
+ *  The structure representing the count of observed TCP flags.
+ *  To observe new flags, they should be added to this structure and _ct_record_tcp_flags updated.
+ */
+struct tcpflagscount
```
nddq, line 38: Should we cover all of the other TCP flags, even though that they are uncommon?

nddq (approved):
Overall lgtm, beside a minor comment. May I suggest naming this PR to something along the lines of `fix underreporting of TCP metrics due to conntrack reporting logic`? @SRodi for another pair of eyes

pkg/plugin/conntrack/_cprog/conntrack.c
mmckeen: I can add all of the ones we support in the docs.

mmckeen:
@nddq added the remaining flags to metrics, appreciate one more review when you get the chance 🙇

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -31,23 +31,53 @@ struct conntrackmetadata {
...
+    __u32 dst_ip;
+    __u16 src_port;
+    __u16 dst_port;
+    struct tcpmetadata tcp_metadata; // TCP metadata
+    __u8 observation_point;
+    __u8 traffic_direction;
+    __u8 proto;
+    __u16 flags; // For TCP packets, this is the TCP flags. For UDP packets, this is will always be 1 for conntrack purposes.
+    bool is_reply;
+    __u32 previously_observed_packets; // When sampling, the is the number of observed packets since the last report.
```
SRodi:
```suggestion
    __u32 previously_observed_packets; // When sampling, this is the number of observed packets since the last report.
```

pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -31,23 +31,53 @@ struct conntrackmetadata {
...
+    __u32 previously_observed_bytes; // When sampling, the is the number of observed bytes since the last report.
```
SRodi:
```suggestion
    __u32 previously_observed_bytes; // When sampling, this is the number of observed bytes since the last report.
```

SRodi (commented):
@mmckeen just a couple of very minor nit. - one question, would it be worth to add this to our documentation?

mmckeen:
> @mmckeen just a couple of very minor nit. - one question, would it be worth to add this to our documentation?

Potentially, happy to follow up on that with a future PR 🙇

SRodi (approved):
LGTM @mmckeen - please update the branch prior merge

nddq (approved):

PR #1684, opened 2025-06-14
Added handling for CiliumOnWindows Registry Values in Hns and EbpfWin…

kumarvin123:
# Description
Added registry check for CiliumOnWindows in Hns and EbpfWindows plugins  to support CoExistence in Mixed Mode clusters

Added a common function to check for the Registry value.

## Related Issue

## Checklist

I have manually deployed the image on AKS cluster  with   --set enabledPlugin_win="[ebpfwindows\,hnsstats\]" `
and verified through logs either ebpfwindows or hnsttats are running based on CiliumOnWindows Registry value

## Additional Notes

---

### DIFF
--- a/pkg/plugin/common/common_windows.go
+++ b/pkg/plugin/common/common_windows.go
@@ -8,13 +8,19 @@ import (
 	"golang.org/x/sys/windows/registry"
 )
 
+const (
+	// KeyPath is the registry key path where the CiliumOnWindows value is stored.
+	// This key is used to determine if Cilium is enabled on Windows.
+	KeyPath = `SYSTEM\CurrentControlSet\Services\hns\State`
+	// CiliumOnWindows is the registry value name that indicates if Cilium is enabled on Windows.
+	// If this value is set to 1, Cilium is enabled on Windows. If this value is not set or set to 0, Cilium is not enabled.
+	ValueName = "CiliumOnWindows"
+)
+
 // IsCiliumOnWindowsEnabled checks if the CiliumOnWindows registry value is set to 1.
 // Returns (true, nil) if set to 1, (false, nil) if not set or not exist, (false, err) for other errors.
 func IsCiliumOnWindowsEnabled() (bool, error) {
-	keyPath := `SYSTEM\CurrentControlSet\Services\hns\State`
-	valueName := "CiliumOnWindows"
-
-	k, err := registry.OpenKey(registry.LOCAL_MACHINE, keyPath, registry.QUERY_VALUE)
+	k, err := registry.OpenKey(registry.LOCAL_MACHINE, KeyPath, registry.QUERY_VALUE)
 	if err != nil {
 		if err == registry.ErrNotExist {
 			return false, nil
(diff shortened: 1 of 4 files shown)


pkg/plugin/common/common_windows.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"golang.org/x/sys/windows/registry"
+)
+
+// IsCiliumOnWindowsEnabled checks if the CiliumOnWindows registry value is set to 1.
+// Returns (true, nil) if set to 1, (false, nil) if not set or not exist, (false, err) for other errors.
+func IsCiliumOnWindowsEnabled() (bool, error) {
+	keyPath := `SYSTEM\CurrentControlSet\Services\hns\State`
+	valueName := "CiliumOnWindows"
```
anubhabMajumdar: Make these constants and add a comment.

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -80,9 +82,27 @@ func (p *Plugin) Name() string {
...
+	if ciliumEnabled {
+		p.l.Info("Cilium is enabled on Windows, proceeding with ebpfWindows plugin initialization")
+		p.pullMetricsAndEvents(ctx)
+		p.l.Info("Complete ebpfWindows plugin...")
+		return nil
+	}
+
+	if err != nil {
+		p.l.Error("Error while checking if Cilium is enabled on Windows", zap.Error(err))
+	}
```
anubhabMajumdar: We should check for error first. What should be expected behavior in case of error? I would imagine plugin would return the error without running the plugin.

pkg/plugin/hnsstats/hnsstats_windows.go
```diff
@@ -210,8 +211,26 @@ func notifyHnsStats(h *hnsstats, stats *HnsStatsData) {
...
+
+	ciliumEnabled, err := plugincommon.IsCiliumOnWindowsEnabled()
+
+	if !ciliumEnabled {
+		return pullHnsStats(ctx, h)
+	}
+
+	if err != nil {
+		h.l.Error("Error while checking if Cilium is enabled on Windows", zap.Error(err))
+	}
```
anubhabMajumdar: Same as the other plugin, handle the error first.

pkg/plugin/ebpfwindows/ebpf_windows.go
kumarvin123: If we encounter an error while accessing the registry key (We proceed as if the registry check does not exist). Its that or we fail the plugin to come up. Either way there is no ideal solution. I would prefer to leave it like this

kumarvin123: Let me know if we have to bail out here in case of an error

anubhabMajumdar: This behavior is suspect. If `err != nil`, then  `ciliumEnabled` is `false`. So retina would report plugin `ebpfwindows` running successfully, but that's not the case. The plugin is just spinning without generating metrics.

I think we should edit as follows:

```
ciliumEnabled, err := plugincommon.IsCiliumOnWindowsEnabled()
if err != nil { return err } // This will crash retina and signal something went fatally wrong
if !ciliumEnabled { 
// log Cilium not enabled
return nil
}

// Run plugin to generate metrics/traces.
...
```
Refer to https://github.com/microsoft/retina/blob/main/pkg/plugin/packetparser/packetparser_linux.go#L234-L237 which does something similar (check is different, but it returns nil.

PR #1767, opened 2025-07-22
feat(packetparser): Allow sampling of packets

mmckeen:
# Description

## Related Issue

## Checklist

## Additional Notes

---


pkg/plugin/conntrack/_cprog/conntrack.c
```diff
@@ -513,21 +530,27 @@ static __always_inline struct packetreport _ct_should_report_packet(struct ct_v4
...
+        }
+    }
+
     // Report if:
     // 1. We already decided to report based on protocol-specific rules, or
-    // 2. New flags have appeared, or
+    // 2. New flags have appeared and the packet has been sampled, or
     // 3. Reporting interval has elapsed
-    if (should_report || flags != seen_flags || now - last_report >= CT_REPORT_INTERVAL) {
+    if (should_report || (sampled && flags != seen_flags) || now - last_report >= CT_REPORT_INTERVAL) {
```
mmckeen, line 545: This might be worth relaxing so we don't set `should_report` for `TCP_SYN | TCP_URG | TCP_ECE | TCP_CWR` so we apply sampling to those flags as well.

`TCP_SYN` is fairly common so it's likely reducing how often we can sample.

mmckeen:
@nddq @srodi any thoughts on this PR?

nddq:
@mmckeen I will try to make a pass this week

pkg/config/config.go
```diff
@@ -24,8 +24,9 @@ const (
 )
 
 var (
-	ErrorTelemetryIntervalTooSmall = fmt.Errorf("telemetryInterval smaller than %v is not allowed", MinTelemetryInterval)
-	DefaultTelemetryInterval       = 15 * time.Minute
+	ErrorTelemetryIntervalTooSmall        = fmt.Errorf("telemetryInterval smaller than %v is not allowed", MinTelemetryInterval)
+	DefaultTelemetryInterval              = 15 * time.Minute
+	DefaultSamplingRate            uint32 = 1
```
nddq, line 29: might be a good idea to have this in our official docs, explaining the sampling rate

pkg/plugin/packetparser/_cprog/packetparser.c
```diff
@@ -208,11 +208,24 @@ static void parse(struct __sk_buff *skb, __u8 obs)
...
 
+    #ifdef DATA_AGGREGATION_LEVEL
+
+	// Calculate sampling
+	bool sampled __attribute__((unused));
+	sampled = true;
+	
+	#ifdef DATA_SAMPLING_RATE
+	    u32 rand __attribute__((unused));
+		rand = bpf_get_prandom_u32();
```
nddq, line 219: is this function guaranteed to exist from the kernel version that Retina supported and onward?

pkg/plugin/packetparser/_cprog/packetparser.c
```diff
@@ -208,11 +208,24 @@ static void parse(struct __sk_buff *skb, __u8 obs)
...
+		if (rand >= UINT32_MAX / DATA_SAMPLING_RATE) {
```
nddq, line 220: Should this check be `rand < UINT32_MAX / DATA_SAMPLING_RATE` instead? For example, with DATA_SAMPLING_RATE = 1000, this gives a ~1/1000 chance of sampling each packet.

nddq (changes requested):
I think we should revisit the syntax of `DataSamplingRate`. For example, since I know the code context, I understand that setting `DataSamplingRate` to 1 means every packet will be sampled. But for someone new to Retina, this behavior may not be obvious.

nddq:
Also, I'm all for ditching the reporting period entirely if we can make packet sampling works as it is a lot more flexible than the currently hardcoding value that we are using for reporting period 🙂 cc @anubhabMajumdar @SRodi So what I'm thinking of is either we report a packet because it carries a new flag or because it was randomly sampled

pkg/plugin/packetparser/_cprog/packetparser.c
mmckeen: Looks like this is available since 4.17 https://github.com/torvalds/linux/commit/df4a37d8b53f9fb9af722b056da5edbd9a531768

pkg/plugin/packetparser/_cprog/packetparser.c
mmckeen: The logic is a bit confusing here since this conditional turns off sampling for the packet, does that make sense?

nddq: Yes, the same constant is currently being used both for enablement and for the sampling value. It would be clearer to separate these into two different constants. Also, my earlier point about `rand < UINT32_MAX / DATA_SAMPLING_RATE` still applies. For example, if I set `DATA_SAMPLING_RATE` to `3`, I’d expect 1 out of every 3 packets to be sampled. But with the current check `rand >= UINT32_MAX / DATA_SAMPLING_RATE`, the opposite behavior occurs, far more packets get sampled than intended. Is that correct?

mmckeen: The logic is as follows:

1) by default `sampled = true` meaning all packets are sampled
2) if `rand >= UINT32_MAX / DATA_SAMPLING_RATE` then we set `sampled = false` and don't sample the packet

So for example, with `DATA_SAMPLING_RATE = 3` there is a 2/3rds chance for the packet to pass the conditional and not be sampled, leaving a 1/3rds chance to be sampled.

mmckeen:
@nddq updated this with docs, the sampling calculation still looks accurate to me but happy for feedback 🙇

mmckeen:
> hey @mmckeen , I just merged some changes for conntrack, can you rebase your PR? Thanks!

Done!

PR #1853, opened 2025-08-29
feat: Retina capture download on Windows

kamilprz:
# Description

Retina captures were using a Linux container for the downloading, and had no counterpart for Windows based clusters. This PR adds in support for Windows and its different flavours of ltsc.

## Related Issue

https://github.com/microsoft/retina/issues/1834

## Checklist

- [x] I have read the [contributing documentation](https://retina.sh/docs/Contributing/overview).
- [x] I signed and signed-off the commits (`git commit -S -s ...`). See [this documentation](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) on signing commits.
- [x] I have correctly attributed the author(s) of the code.
- [x] I have tested the changes locally.
- [x] I have followed the project's style guidelines.
- [ ] I have updated the documentation, if necessary.
- [ ] I have added tests, if applicable.

## Screenshots (if applicable) or Testing Completed

image

---

Please refer to the [CONTRIBUTING.md](../CONTRIBUTING.md) file for more information on how to contribute to this project.

Issue #1834, opened 2025-08-25
Capture download support for Windows


kamilprz:
Currently, retina capture download only works on Linux clusters as it creates a Linux download pod.

There is demand for matching this functionality on Windows clusters as well, and given that Retina promises to be OS agnostic, it is something we should support.

PR #1905, opened 2025-10-14
fix: download windows

carlotaarvela:
# Description

Retina captures were using a Linux container for the downloading, and had no counterpart for Windows based clusters. This PR adds in support for Windows and its different flavours of ltsc.
## Related Issue

https://github.com/microsoft/retina/issues/1834

## Relate PR
https://github.com/microsoft/retina/pull/1853

## Checklist

## Additional Notes

---

### DIFF
--- a/cli/cmd/capture/download_test.go
+++ b/cli/cmd/capture/download_test.go
@@ -182,26 +182,17 @@ func TestDownloadFromCluster(t *testing.T) {
 			// on testing the service methods and individual functions
 			pods, err := getCapturePods(ctx, kubeClient, captureName, tc.namespace)
 
-			if tc.wantErr {
-				if err == nil {
-					t.Errorf("Expected error for test case %s, but got none", tc.name)
-					return
-				}
-				if tc.expectedError != "" && !strings.Contains(err.Error(), tc.expectedError) {
-					t.Errorf("Expected error message to contain %q, but got %q", tc.expectedError, err.Error())
-				}
-				return
+			if tc.wantErr && err == nil {
+				t.Errorf("Expected error for test case %s, but got none", tc.name)
 			}
-
-			if err != nil {
+			if tc.wantErr && err != nil && tc.expectedError != "" && !strings.Contains(err.Error(), tc.expectedError) {
+				t.Errorf("Expected error message to contain %q, but got %q", tc.expectedError, err.Error())
+			}
+			if !tc.wantErr && err != nil {
 				t.Errorf("Unexpected error for test case %s: %v", tc.name, err)
-				return
 			}
-
-			// For successful cases, validate the pods
-			if pods == nil || len(pods.Items) == 0 {
+			if !tc.wantErr && (pods == nil || len(pods.Items) == 0) {
 				t.Errorf("Expected to find capture pods for %s, but got none", tc.name)
-				return
 			}
 
 			// Validate pod properties
@@ -268,18 +259,12 @@ func TestGetNodeOS(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			result, err := getNodeOS(tc.node)
 
-			if tc.wantErr {
-				if err == nil {
-					t.Errorf("Expected error for %s, but got none", tc.name)
-					return
-				}
-			} else {
-				if err != nil {
-					t.Errorf("Unexpected error for %s: %v", tc.name, err)
-					return
-				}
+			if tc.wantErr && err == nil {
+				t.Errorf("Expected error for %s, but got none", tc.name)
 			}
-
+			if !tc.wantErr && err != nil {
+				t.Errorf("Unexpected error for %s: %v", tc.name, err)
+			}
 			if result != tc.expected {
 				t.Errorf("Expected %v, got %v for %s", tc.expected, result, tc.name)
 			}
@@ -505,18 +490,12 @@ func TestGetCapturePods(t *testing.T) {
 			ctx := context.Background()
 			pods, err := getCapturePods(ctx, kubeClient, tc.captureName, tc.namespace)
 
-			if tc.wantErr {
-				if err == nil {
-					t.Errorf("Expected error for %s, but got none", tc.name)
-					return
-				}
-			} else {
-				if err != nil {
-					t.Errorf("Unexpected error for %s: %v", tc.name, err)
-					return
-				}
+			if tc.wantErr && err == nil {
+				t.Errorf("Expected error for %s, but got none", tc.name)
 			}
-
+			if !tc.wantErr && err != nil {
+				t.Errorf("Unexpected error for %s: %v", tc.name, err)
+			}
 			if !tc.wantErr && len(pods.Items) != tc.expectedCount {
 				t.Errorf("Expected %d pods, got %d", tc.expectedCount, len(pods.Items))
 			}
@@ -553,18 +532,17 @@ func TestDownloadFromBlobValidation(t *testing.T) {
 
 			err := downloadFromBlob()
 
-			if tc.wantErr {
-				if err == nil {
-					t.Errorf("Expected error for %s, but got none", tc.name)
-				} else {
-					t.Logf("Expected error occurred for %s: %v", tc.name, err)
-				}
-			} else {
-				if err != nil {
-					t.Errorf("Unexpected error for %s: %v", tc.name, err)
-				} else {
-					t.Logf("Successfully completed %s", tc.name)
-				}
+			if tc.wantErr && err == nil {
+				t.Errorf("Expected error for %s, but got none", tc.name)
+			}
+			if !tc.wantErr && err != nil {
+				t.Errorf("Unexpected error for %s: %v", tc.name, err)
+			}
+			if tc.wantErr && err != nil {
+				t.Logf("Expected error occurred for %s: %v", tc.name, err)
+			}
+			if !tc.wantErr && err == nil {
+				t.Logf("Successfully completed %s", tc.name)
 			}
 		})
 	}


cli/cmd/capture/download.go
```diff
@@ -35,14 +36,34 @@ import (
...
 const (
 	DefaultOutputPath = "./"
 )
 
 var (
-	blobURL     string
-	captureName string
-	outputPath  string
+	blobURL string
+	// Error variables for lint compliance (err113)
```
alexcastilio: this comment can be deleted

cli/cmd/capture/download.go
```diff
@@ -69,19 +197,22 @@ var downloadExample = templates.Examples(i18n.T(`
...
+
 	pods, err := getCapturePods(ctx, kubeClient, captureName, namespace)
 	if err != nil {
 		return errors.Wrap(err, "failed to obtain capture pod")
 	}
 
 	err = os.MkdirAll(filepath.Join(outputPath, captureName), 0o775)
 	if err != nil {
-		return fmt.Errorf("failed to create directory: %w", err)
+		return errors.Wrap(err, ErrCreateDirectory.Error())
```
alexcastilio: I see you are just reusing this as it's already a dependency in the file, but I would suggest replacing the current `github.com/pkg/errors` dependency by `errors` from std library. The former is archived and std library should be enough for this. Then return would be `return errors.Join(ErrCreateDirectory, err)` (This is a recommendation, not a blocking comment)

cli/cmd/capture/download.go
```diff
@@ -52,8 +73,115 @@ var (
...
+		ctx:        ctx,
+	}
+}
+
+func getDownloadCmd(node *corev1.Node, hostPath, fileName string) *DownloadCmd {
+	nodeOS, err := getNodeOS(node)
+	if err != nil {
+		retinacmd.Logger.Error("Failed to detect node OS", zap.String("node", node.Name), zap.Error(err))
+		return nil
+	}
```
alexcastilio, line 118: isn't it better to return the error and leave to the caller to decide what to do with it? Or if you don't want an empty output to be considered an error, change getNodeOS to return only string/NodeOS as getDownloadCmd is doing and the caller will decide if that's acceptable or not. It does not look consistent to me as it is now because getNodeOS outputs an error -> getDownloadCmd does not consider that error and returns empty string -> DownloadFile creates an error again based on empty output.

cli/cmd/capture/download.go
```diff
@@ -197,23 +355,28 @@ func createDownloadPod(ctx context.Context, kubeClient *kubernetes.Clientset, na
 	}
 
 	fmt.Printf("Creating download pod: %s\n", podName)
-	_, err := kubeClient.CoreV1().Pods(namespace).Create(ctx, podSpec, metav1.CreateOptions{})
+	_, err := ds.kubeClient.CoreV1().Pods(ds.namespace).Create(ds.ctx, podSpec, metav1.CreateOptions{})
```
alexcastilio: Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it: https://go.dev/blog/context-and-structs

cli/cmd/capture/download.go
```diff
@@ -52,8 +73,115 @@ var (
...
+	FileCheckCommand []string
+	FileReadCommand  []string
+}
+
+// DownloadService encapsulates the download functionality and shared dependencies
+type DownloadService struct {
+	kubeClient kubernetes.Interface
+	config     *rest.Config
+	namespace  string
+	ctx        context.Context
```
alexcastilio: Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it: https://go.dev/blog/context-and-structs

cli/cmd/capture/download.go
```diff
@@ -52,8 +73,115 @@ var (
...
+		retinacmd.Logger.Info("Detected node OS: Windows", zap.String("node", node.Name), zap.String("os", node.Status.NodeInfo.OperatingSystem))
+		return Windows, nil
+	}
+
+	if strings.Contains(nodeOS, "linux") {
+		retinacmd.Logger.Info("Detected node OS: Linux", zap.String("node", node.Name), zap.String("os", node.Status.NodeInfo.OperatingSystem))
+		return Linux, nil
+	}
+
+	return Linux, errors.Wrap(ErrEmptyDownloadOutput, "unsupported operating system: "+node.Status.NodeInfo.OperatingSystem)
```
alexcastilio: I think this should now return Linux by default. It's confusing. Return empty nodeOS or change return type to pointer and return nil instead

cli/cmd/capture/download_test.go
```diff
@@ -0,0 +1,788 @@
...
+			name: "Unknown OS node",
+			node: &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{Name: "unknown-test"},
+				Status: corev1.NodeStatus{
+					NodeInfo: corev1.NodeSystemInfo{
+						OperatingSystem: "darwin",
+					},
+				},
+			},
+			expected: Linux, // Default to Linux for unknown OS
```
alexcastilio: I think this is not the behavior that should be expected. If NodeOS is unknown, than it should return nil, error or an UnknownOS value

cli/cmd/capture/download_test.go
```diff
@@ -0,0 +1,788 @@
...
+				}
+				if !strings.Contains(cmd.ContainerImage, "busybox") {
+					t.Errorf("Expected Linux container image to contain 'busybox', got %s", cmd.ContainerImage)
+				}
+				if !strings.Contains(cmd.SrcFilePath, "/host/tmp/captures/"+testCapture+".tar.gz") {
+					t.Errorf("Expected Linux source file path to match pattern, got %s", cmd.SrcFilePath)
+				}
+				if len(cmd.FileReadCommand) == 0 || cmd.FileReadCommand[0] != "cat" {
+					t.Errorf("Expected Linux file read command to start with 'cat', got %v", cmd.FileReadCommand)
+				}
```
alexcastilio: It's missing validation of MountPath, KeepAliveCommand and FileCheckCommand

cli/cmd/capture/download_test.go
```diff
@@ -0,0 +1,788 @@
...
+			blobURL = tc.blobURL
+			defer func() { blobURL = originalBlobURL }()
+
+			err := downloadFromBlob()
+
+			if tc.wantErr && err == nil {
+				t.Errorf("Expected error for %s, but got none", tc.name)
+			}
+
+			if !tc.wantErr && err == nil {
```
alexcastilio: what happens when !tc.wantError && err != nil? This is not tested

pkg/capture/constants/job_specification.go
```diff
@@ -20,8 +20,10 @@ const (
 	CaptureContainerEntrypoint    string = "./retina/captureworkload"
 	CaptureContainerEntrypointWin string = "captureworkload.exe"
 
-	CaptureAppname       string = "capture"
-	CaptureContainername string = "capture"
+	CaptureAppname        string = "capture"
+	CaptureContainername  string = "capture"
+	DownloadAppname       string = "download"
+	DownloadContainername string = "download"
```
alexcastilio: nit: name -> Name

cli/cmd/capture/download.go
carlotaarvela: I have done the replacement and updated the errors

cli/cmd/capture/download.go
carlotaarvela: I agree. I have updated this to return a nil and the error

cli/cmd/capture/download.go
carlotaarvela: Done

cli/cmd/capture/download.go
carlotaarvela: Done

cli/cmd/capture/download.go
carlotaarvela: I have updated this to return nil and the unsupported os error

cli/cmd/capture/download_test.go
carlotaarvela: Updated

cli/cmd/capture/download_test.go
carlotaarvela: Added

pkg/capture/constants/job_specification.go
carlotaarvela: Done

cli/cmd/capture/download_test.go
carlotaarvela: Added

cli/cmd/capture/download_test.go
```diff
@@ -0,0 +1,843 @@
...
+			} else {
+				if err != nil {
+					t.Errorf("Unexpected error for %s: %v", tc.name, err)
+					return
+				}
+			}
+
+			if !tc.wantErr && len(pods.Items) != tc.expectedCount {
+				t.Errorf("Expected %d pods, got %d", tc.expectedCount, len(pods.Items))
+			}
```
alexcastilio: I would suggest to avoid nesting code blocks whenever possible. it makes code easier to read and understand.
As we are running tests, there are some possible choices about the test function behavior with available methods of testing.T, without need to call return (e.g. if you want test execution to stop, call t.Fatalf() which is equivalent to t.Logf + t.FailNow; if you want to mark test as failed but continue other checks, t.Errorf() can be used as now).
I suggest the snipped below, with less indentations levels and no return calls (you can replace Errorf by Fatalf if you want to stop execution of following checks):
```suggestion
			if tc.wantErr && err == nil {
				t.Errorf("Expected error for %s, but got none", tc.name)
			}
			if !tc.wantErr && err != nil {
				t.Errorf("Unexpected error for %s: %v", tc.name, err)
			}
			if len(pods.Items) != tc.expectedCount {
				t.Errorf("Expected %d pods, got %d", tc.expectedCount, len(pods.Items))
			}
```

alexcastilio (dismissed):

alexcastilio (approved):

Issue #1692, opened 2025-06-17
Expire stale exported metrics


mmckeen:
**Is your feature request related to a problem? Please describe.**

For advanced metrics with high-cardinality (pod-level, port, IP, etc..) many metrics can become stale and not change for a long period of time.  These metrics end up remaining exported to Prometheus and take up a significant amount of resources being scraped. 

We should remove these metrics from the exported set if they have not changed in a significant period of time to save significant resource overhead on serving stale metrics.

**Describe the solution you'd like**

Stale metrics after some configurable time period (say 15 minutes by default) are no longer exported to Prometheus for advanced metrics.

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

PR #1921, opened 2025-10-28
feat: expire stale advanced metrics after not being updated for some time

mmckeen:
# Description

With advanced metrics, high-cardinality labels can cause the metrics export to bloat leading to unbounded memory and resource usage.

This PR sets up an optional TTL for advanced metrics defined in the `MetricsConfiguration` CRD.  By default, the TTL is infinite and cleanup is not tracked or done.

When defined, on a period equal to the TTL metrics which have not been updated within the last TTL duration will be removed from the metrics export.

For counters and gauges that look like counters, this will be treated by Prometheus similar to any other missing metric (for example from an application restart).  As long as functions like `rate` or `increase` are used, calculations will remain accurate.

This also includes a refactor of the base metrics object as an interface, allowing mocking in tests and a clean API contract with child objects that compose it.

## Related Issue

https://github.com/microsoft/retina/issues/1692

## Checklist

Deployed, modified CRD to enable/disable and change the TTL.

Metrics are re-initialized as expected.

Made sure that CRD validation rejects invalid TTL values.

Also, added comprehensive tests for the TTL-based cleanup logic.

---


nddq:
@mmckeen removed the stale tag

pkg/module/metrics/basemetricsobject.go
```diff
@@ -3,25 +3,119 @@
...
 		advEnable:   ctxOptions.IsAdvanced(),
 		ctxOptions:  ctxOptions,
 		l:           fl,
 		contextMode: isLocalContext,
+		lastUpdated: make(map[string]updated),
+		expireFn:    expireOrInfiniteTTL,
+	}
+
+	if expireOrInfiniteTTL != nil {
+		go func() {
```
nddq, line 163: I'm concern about the potential goroutine leaks here since there is no proper cancellation. If the `MetricsConfiguration` CRD is reconciled, a new goroutine will spawn and the old one won't get clean up.

pkg/module/metrics/basemetricsobject.go
```diff
@@ -3,25 +3,119 @@
...
+	var bf bytes.Buffer
+
+	for _, l := range lbs {
+		bf.WriteString(l)
+	}
+
+	h := sha256.New()
+	h.Write(bf.Bytes())
+
+	s := string(h.Sum(nil))
```
nddq: Can't we just concatenate all labels with a unique separator to use it as the key instead? Using sha256 here is kinda overkill imo

pkg/module/metrics/basemetricsobject.go
```diff
@@ -3,25 +3,119 @@
...
+		go func() {
+			for {
+				b.l.Debug(fmt.Sprintf("Expiring metrics: %s", ctxOptions.MetricName))
+				n := b.expire(ttl)
+				b.l.Debug(
+					fmt.Sprintf("Metric expiration finished: %s", ctxOptions.MetricName),
+					zap.Time("next_expiration", time.Now().Add(ttl)),
+					zap.Int("expired", n),
+				)
+				time.Sleep(ttl)
```
nddq: using `time.Ticker` here might be better here for cleanup purposes

nddq (changes requested):
Sorry for taking a while to respond 😅 Gave this some thoughts

pkg/module/metrics/basemetricsobject.go
mmckeen: Fair, let me wrap this in a `sync.Once` or similar

deploy/standard/manifests/controller/helm/retina/crds/retina.sh_captures.yaml
```diff
@@ -165,6 +173,14 @@ spec:
                             type: object
                         type: object
                         x-kubernetes-map-type: atomic
+                      podNames:
```
mmckeen, line 176: Looks like these were missed in previous PRs and this pulled them in when re-generating the CRD manifests, I can manually remove them or put up a second PR to reconcile them.  Let me know 🙇

nddq: let's put this into a different PR if possible

mmckeen: Created https://github.com/microsoft/retina/pull/2026

mmckeen:
@nddq appreciate a review on this when you get the chance :)

It's the last PR away from us being able to run the upstream version of Retina rather than our fork.

pkg/module/metrics/basemetricsobject.go
```diff
@@ -3,29 +3,183 @@
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+
 	b := baseMetricObject{
+		RWMutex:     &sync.RWMutex{},
 		advEnable:   ctxOptions.IsAdvanced(),
 		ctxOptions:  ctxOptions,
 		l:           fl,
 		contextMode: isLocalContext,
+		lastUpdated: make(map[string]updated),
```
nddq: this map will always get allocated even if `ttl < 0`, the same thing with the `ctx` and the mutex, we might be able to refactor this out so that these only get allocated if `ttl > 0` (expiration enabled)

pkg/module/metrics/basemetricsobject.go
```diff
@@ -3,29 +3,183 @@
...
+		ctx:         ctx,
+	}
+
+	if expireOrInfiniteTTL != nil {
+		b.l.Info(
+			"Starting metric expiration routine: "+ctxOptions.MetricName,
+			zap.Duration("ttl", ttl),
+		)
+		go func() {
+			defer cancel()
```
nddq: nit: this might be unnecessary here since the goroutine will only exit if `<-ctx.Done()` is closed, meaning that `cancel()` is already called.

crd/api/v1alpha1/validations/validate_metricconfiguration_test.go
nddq, line 1: nit: add a test for edge cases? 0 or negative values for ttl

pkg/module/metrics/dns_test.go
```diff
@@ -202,114 +204,134 @@ func TestProcessLocalCtx(t *testing.T) {
...
-			tt.d.srcCtx = m
-			tt.d.processLocalCtxFlow(tt.input)
+			ctxOptions := &v1alpha1.MetricsContextOptions{
+				MetricName: utils.DNSResponseCounterName,
+				SourceLabels: []string{
+					podCtxOption,
+					namespaceCtxOption,
+				},
+			}
+			d := NewDNSMetrics(ctxOptions, l, localContext, time.Duration(0))
```
nddq: i think `time.Duration(0)` can be just 0?

pkg/module/metrics/drops_test.go
```diff
@@ -288,11 +298,40 @@ func TestNewDrop(t *testing.T) {
...
 			f.ProcessFlow(tc.f)
+
+			// There should be no tracked metrics when TTL is infinite
+			assert.Equal(t, 0, len(f.trackedMetricLabels()), "there should be no tracked metrics when TTL is infinite Test Name: %s", tc.name)
+
+			// Test TTL based expiration
+			metricsinit.InitializeMetrics()
+
+			// Set the TTL to something high to ensure that our call to expire is the only one that expires the metrics
+			f = NewDropCountMetrics(tc.opts, log.Logger(), tc.localContext, time.Duration(time.Minute))
```
nddq: nit: `time.Minute` is already a `Duration` type, so no need to cast it here

nddq (approved):
lgtm, thanks! Just realized that markdown lint is not required to merge PRs

PR #1972, opened 2025-12-09
fix: add logging to K8sErrorHandler when error is nil

### LINKED ISSUE #1951: k8sWatcherError: potential crash
**Describe the bug**

From https://github.com/kubernetes/kubernetes/issues/135349:

In https://grep.app/search?q=runtime.ErrorHandlers one can see that some downstreams adapted to the [API change in 1.31](https://github.com/kubernetes/kubernetes/commit/5a130d2b71e5d70cfff15087f4d521c6b68fb01e) by simply ignoring all the new parameters (message string, key/value pairs) and only using the error, like they did before. Apparently someone took the path of least resistance when updating their Kubernetes dependencies, without looking at the upstream change.

Not only is this capturing the problem incompletely, it will also crash for those HandleErrorWithContext/Logger calls which pass a nil error.

**To Reproduce**

Check https://github.com/microsoft/retina/pull/1183/files#r2569342078

**Expected behavior**

Maybe use the new helper code from https://github.com/kubernetes/kubernetes/pull/135419

alexcastilio:
# Description

Add logging to K8sErrorHandler when error is nil
closes #1951 

## Related Issue

## Checklist

## Additional Notes

---

### DIFF
--- a/pkg/k8s/watcher_linux.go
+++ b/pkg/k8s/watcher_linux.go
@@ -43,7 +43,15 @@ func Start(ctx context.Context, k *watchers.K8sWatcher) {
 // retinaK8sErrorHandler is a custom error handler for the watcher
 // that logs the error and tags the error to easily identify
 func k8sWatcherErrorHandler(c context.Context, e error, s string, i ...interface{}) {
+	if e == nil {
+		// TODO: handle key/values pairs in a better way
+		// current example output: time="2009-11-10T23:00:00Z" level=error msg="msg: Some error message -- key/values: [int 1 str world]"
+		logger.WithContext(c).Errorf("msg: %s -- key/values: %+v", s, i)
+		return
+	}
+
 	errStr := e.Error()
+
 	logError := func(er, r string) {
 		logger.WithFields(logrus.Fields{
 			"underlyingError": er,


pkg/k8s/watcher_linux.go
```diff
@@ -43,7 +43,13 @@ func Start(ctx context.Context, k *watchers.K8sWatcher) {
 // retinaK8sErrorHandler is a custom error handler for the watcher
 // that logs the error and tags the error to easily identify
 func k8sWatcherErrorHandler(c context.Context, e error, s string, i ...interface{}) {
+	if e == nil {
+		logger.WithContext(c).Error(s, i)
```
pohly: Where is this `Error` defined? You are passing it an single array. Did you perhaps mean `Error(s, i...)`?

pohly: Also, why not used `k8s.K8sErrorHandler(c, nil, s, i...)`? That would be consistent with the `default` branch below.

alexcastilio: `Error(args ...interface{})` comes from logrus: https://github.com/sirupsen/logrus/blob/b61f268f75b6ff134a62cd62aee1095fa12e8d2e/entry.go#L335
You are right, but it was not accepting arguments `s, i...`, so I continued like that. Just pushed a fix to that

I'm not using the same as `default` because that is imported from cilium repository and it's not yet fixed to take nil error as argument. I raised a PR there suggesting similar fix: https://github.com/cilium/cilium/pull/43216

pohly: You cannot pass message and key/value pairs to logrus. It doesn't format them properly:

https://go.dev/play/p/ec5r0jo36di
```
logrus.Error("hello", "int", 1, "str", "world")
```

=>

```
time="2009-11-10T23:00:00Z" level=error msg=helloint1strworld
```

You can format as a string using the helper code from https://github.com/kubernetes/kubernetes/pull/135419 and then pass the resulting string to logrus.

pkg/k8s/watcher_linux.go
```diff
@@ -43,7 +43,15 @@ func Start(ctx context.Context, k *watchers.K8sWatcher) {
 // retinaK8sErrorHandler is a custom error handler for the watcher
 // that logs the error and tags the error to easily identify
 func k8sWatcherErrorHandler(c context.Context, e error, s string, i ...interface{}) {
+	if e == nil {
+		args := []any{s}
+		args = append(args, i...)
+		logger.WithContext(c).Error(args...)
```
pohly: https://github.com/microsoft/retina/pull/1972#discussion_r2605633892

pkg/k8s/watcher_linux.go
alexcastilio: I have changed it to improve a bit the formatting:
https://go.dev/play/p/Kh3-f88C9PI

It's not ideal yet but as that helper code is still being reviewed, we can return to this later. This is ok for now.

SRodi (approved):
LGTM - we can improve the formatting if/when the helper code will be available on upstream kubernetes

Issue #655, opened 2024-08-26
Degraded node network throughput when retin is installed


grzesuav:
**Describe the bug**
Along with 1.29 AKS upgrade, retina agens was installed on our nodes, which resulted in degraded network throughput, details in https://github.com/Azure/AKS/issues/4508

**To Reproduce**
See related issue - https://github.com/Azure/AKS/issues/4508
**Expected behavior**
Network throughput not impacted by retina

**Screenshots**
See related issue - https://github.com/Azure/AKS/issues/4508

**Platform (please complete the following information):**

- OS: [AKSUbuntu]
- Kubernetes Version: [1.29/1.30]
- Host: AKS
- Retina Version:

**Additional context**

vakalapa:
@grzesuav  thanks for reporting the issue, can you give me some more information on if the perf degrade is within same Node, or  with traffic between different nodes? We know that intra node has some affect with eBPF programs as there is no noise and it even a small ebpf prog can affect the line rate. 

If the perf degrade you saw was in INTER node communication, we can run some tests based on criteria you provide. We can then root cause it to which one or more eBPF progs could be causing this issue.

grzesuav:
@vakalapa as far I can tell it was between inter node and Azure (blob storage) - however I cannot say fo 100%. We have some internal S3 like app which is using azure blob as backend. The graphs from https://github.com/Azure/AKS/issues/4508 shows network throughput of this app - not sure what more I can provide - as you can see there  the overall throughput is very high

grzesuav:
hi, is there any update on this ?

vakalapa:
We are working on the performance pipeline for public test results. We were still unable to repro the issue, @ritwikranjan plz tie this issue to your performance pipeline work ?

PR #1981, opened 2025-12-15
feat(packetparser): add configurable BPF Ring Buffer support

### LINKED ISSUE #1966: offer configurable BPF ring buffers (`BPF_MAP_TYPE_RINGBUF`) impl in packetparser

related to https://github.com/microsoft/retina/issues/655 

On high-core machines (32+ CPUs), Retina experiences significant performance degradation due to its use of BPF perf event arrays. The per-CPU buffer architecture causes non-linear degradation as core count increases, especially with multi-core traffic distribution (typical with RSS/multi-queue NICs) where each active CPU's buffer must be polled independently.

The current implementation uses hardcoded per-CPU buffer sizes (`perCPUBuffer = 32` for packetparser, `perCPUBuffer = 16` for dropreason) that were tuned for modest systems but are deployed on increasingly powerful hardware. Users have no way to configure buffer behavior to match their workload.

**Describe the solution you'd like**

Implement BPF ring buffers (`BPF_MAP_TYPE_RINGBUF`) with configurable buffer size as an alternative to perf event arrays for a more consistent scaling. 

**Describe alternatives you've considered**

**Additional context**

Some analysis for a particular traffic pattern is available here: https://blog.zmalik.dev/p/who-will-observe-the-observability

Retina already uses `github.com/cilium/ebpf v0.18.0` which includes the `ringbuf` package - no new dependencies required. Ring buffers require kernel 5.8+; Retina currently supports 5.4+, so a feature flag with automatic fallback would be great.

SRodi:
# Description

This PR introduces support for `BPF_MAP_TYPE_RINGBUF` in the `packetparser` plugin as a configurable alternative to `BPF_MAP_TYPE_PERF_EVENT_ARRAY`. Ring Buffers offer better performance and memory efficiency, especially on high-core systems, by using a shared buffer across CPUs rather than per-CPU buffers.

### Changes
- **Configuration**: Added `packetParserRingBuffer` (enum: `enabled`/`disabled`; `auto` reserved) and `packetParserRingBufferSize` (uint32) to the Retina configuration and Helm charts.
- **BPF**: Updated `packetparser.c` to conditionally compile with `BPF_MAP_TYPE_RINGBUF` when enabled.
- **Userspace**: Updated `packetparser_linux.go` to:
    - Pass `-DUSE_RING_BUFFER` and `-DRING_BUFFER_SIZE` flags during BPF compilation.
    - Implement a `ringBufReaderWrapper` to adapt the `cilium/ebpf/ringbuf` reader to the existing reader interface.
    - Add logging to indicate which buffer type is active.
- **Testing**: Updated unit tests to mock metrics correctly and added a new test case to verify compilation with Ring Buffer enabled.
- **Userspace (follow-up)**:
    - Decoupled the reader interface from `cilium/ebpf/perf.Record` by introducing a custom `perfRecord` struct to support both perf array and ring buffer paths cleanly.
    - Added a `perfReaderWrapper` and updated `ringBufReaderWrapper` to convert records into the new `perfRecord` type.
- **Kernel gating**:
    - Added shared kernel version parsing/comparison helpers and reusable kernel release retrieval utils.
    - Enforced a minimum kernel version when ring buffer mode is enabled.
    - Updated telemetry to reuse the shared kernel release helper.
- **Docs/config**: Updated config docs and Helm `values.yaml` to reflect `packetParserRingBuffer`.
- **Tests**:
    - Added unit tests for kernel release parsing and version comparison helpers.
    - Limited the ring buffer config test to Linux builds.
    - Adjusted packetparser tests to align with the new `perfRecord` reader interface.
- **Chore/lint**: Addressed linting by using safe page-size casts, wrapping perf errors, and splitting long lines.

### Verification
- Verified unit tests pass: `go test -v ./pkg/plugin/packetparser/...`
- Manual verification on Kind:
    - Built image with `enablePacketParserRingBuffer: true`.
    - Deployed to Kind cluster.
    - Verified logs show "Initializing Ring Buffer reader".
    - Verified BPF map type is `ringbuf` using `bpftool`.

## Related Issues/PRs

- #655 
- fixes #1966 
- #1965 

## Checklist

image

Screenshot 2025-12-15 093850

Screenshot 2025-12-15 093819

## Additional Notes

Once this PR gets reviewed and merged, I will update the site docs accordingly. See issue #2016

In addition, I created an issue to implement an auto option to let retina decide at runtime whether to use `BPF_MAP_TYPE_RINGBUF` or not. See #2017 

I stuck with `BPF_MAP_TYPE_PERF_EVENT_ARRAY` (Perf Buffers) as the default implementation because we still have a requirement to support Linux Kernel 5.4+, as noted in our [intro docs](https://retina.sh/docs/Introduction/intro). Since Ring Buffers (`BPF_MAP_TYPE_RINGBUF`) weren't introduced until kernel 5.8, making them the default would break Retina on older LTS environments (like Ubuntu 20.04) that we're currently committed to supporting.

We also haven't had a chance to benchmark the performance difference on high-core clusters yet. The [upstream documentation](https://docs.ebpf.io/linux/map-type/BPF_MAP_TYPE_RINGBUF/) is clear that while `BPF_MAP_TYPE_RINGBUF` is generally more efficient, it uses a single shared ring buffer whereas `BPF_MAP_TYPE_PERF_EVENT_ARRAY` uses a per-CPU approach, which has different scalability characteristics.

---

### DIFF
--- a/pkg/plugin/packetparser/packetparser_linux.go
+++ b/pkg/plugin/packetparser/packetparser_linux.go
@@ -52,8 +52,14 @@ import (
 )
 
 //go:generate go run github.com/cilium/ebpf/cmd/bpf2go@master -cflags "-g -O2 -Wall -D__TARGET_ARCH_${GOARCH} -Wall" -target ${GOARCH} -type packet packetparser ./_cprog/packetparser.c -- -I../lib/_${GOARCH} -I../lib/common/libbpf/_src -I../lib/common/libbpf/_include/linux -I../lib/common/libbpf/_include/uapi/linux -I../lib/common/libbpf/_include/asm -I../filter/_cprog/ -I../conntrack/_cprog/
-var errNoOutgoingLinks = errors.New("could not determine any outgoing links")
-var errRingBufKernelTooOld = errors.New("ring buffer requires newer kernel")
+var (
+	errNoOutgoingLinks          = errors.New("could not determine any outgoing links")
+	errRingBufKernelTooOld      = errors.New("ring buffer requires newer kernel")
+	errRingBufSizeUnset         = errors.New("ring buffer size must be set when ring buffers are enabled")
+	errRingBufSizeTooSmall      = errors.New("ring buffer size is smaller than the kernel page size")
+	errRingBufSizeTooLarge      = errors.New("ring buffer size is larger than the allowed maximum")
+	errRingBufSizeNotPowerOfTwo = errors.New("ring buffer size is not a power of 2")
+)
 
 func init() {
 	registry.Add(name, New)
@@ -146,25 +152,17 @@ func validateRingBufferSize(size uint32) error {
 	pageSize := uint32(intPageSize)
 
 	if size == 0 {
-		return fmt.Errorf("ring buffer size must be set when ring buffers are enabled")
+		return errRingBufSizeUnset
 	}
 	if size < pageSize {
-		return fmt.Errorf(
-			"ring buffer size (%d) is smaller than the kernel page size (%d)",
-			size,
-			pageSize,
-		)
+		return fmt.Errorf("%w: size=%d page_size=%d", errRingBufSizeTooSmall, size, pageSize)
 	}
 	if size > maxSize {
-		return fmt.Errorf(
-			"ring buffer size (%d) is larger than the allowed maximum (%d)",
-			size,
-			maxSize,
-		)
+		return fmt.Errorf("%w: size=%d max_size=%d", errRingBufSizeTooLarge, size, maxSize)
 	}
 	// Check if size is a power of 2.
 	if (size & (size - 1)) != 0 {
-		return fmt.Errorf("ring buffer size (%d) is not a power of 2", size)
+		return fmt.Errorf("%w: size=%d", errRingBufSizeNotPowerOfTwo, size)
 	}
 
 	return nil
@@ -204,8 +202,8 @@ func (p *packetParser) Compile(ctx context.Context) error {
 	}
 
 	if p.cfg.PacketParserRingBuffer.IsEnabled() {
-		if err := validateRingBufferSize(p.cfg.PacketParserRingBufferSize); err != nil {
-			return err
+		if validateErr := validateRingBufferSize(p.cfg.PacketParserRingBufferSize); validateErr != nil {
+			return validateErr
 		}
 
 		p.l.Info("Compiling with Ring Buffer enabled", zap.Uint32("size", p.cfg.PacketParserRingBufferSize))
--- a/pkg/plugin/packetparser/packetparser_config_test.go
+++ b/pkg/plugin/packetparser/packetparser_config_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestValidateRingBufferSize(t *testing.T) {
@@ -76,7 +77,7 @@ func TestValidateRingBufferSize(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			err := validateRingBufferSize(tt.inputSize)
 			if tt.expectedErr {
-				assert.Error(t, err)
+				require.Error(t, err)
 				if tt.expectedMsg != "" {
 					assert.Contains(t, err.Error(), tt.expectedMsg)
 				}


anubhabMajumdar:
What are the cons for switching over to ring buffer as default? IMO, either make this default or make this decision during runtime (packetparser makes the decision based on node specifications). This seems like a node specific decision and maybe too complex for user to determine.

rbtr (dismissed):
i like it being configurable at runtime. can you update the docs to explain when this should be changed?

pkg/config/config.go
```diff
@@ -65,18 +65,20 @@ type Config struct {
...
+	EnablePodLevel               bool          `yaml:"enablePodLevel"`
+	EnableConntrackMetrics       bool          `yaml:"enableConntrackMetrics"`
+	RemoteContext                bool          `yaml:"remoteContext"`
+	EnableAnnotations            bool          `yaml:"enableAnnotations"`
+	BypassLookupIPOfInterest     bool          `yaml:"bypassLookupIPOfInterest"`
+	DataAggregationLevel         Level         `yaml:"dataAggregationLevel"`
+	MonitorSockPath              string        `yaml:"monitorSockPath"`
+	TelemetryInterval            time.Duration `yaml:"telemetryInterval"`
+	DataSamplingRate             uint32        `yaml:"dataSamplingRate"`
+	EnablePacketParserRingBuffer bool          `yaml:"enablePacketParserRingBuffer"`
```
anubhabMajumdar: I think this should have two options - `auto` and `false` (default `auto`).
Packetparser should make runtime decision on whether to use ring buffer or not .

anubhabMajumdar: It's ok to implement `auto` as a follow up change. Let's not make it a boolean though. Here's a recomendation
```
packetParserRingBuffer: auto    # Let Retina decide
packetParserRingBuffer: enabled  # Always use ring buffer
packetParserRingBuffer: disabled # Never use ring buffer
```

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -152,15 +153,26 @@ func (p *packetParser) Compile(ctx context.Context) error {
...
 		libbpfIncludeAsmDir,
 		libbpfIncludeLinuxDir,
 		libbpfIncludeUapiLinuxDir,
 		filterDir,
 		conntrackDir,
-	)
+	}
+
+	if p.cfg.EnablePacketParserRingBuffer {
+		p.l.Info("Compiling with Ring Buffer enabled", zap.Uint32("size", p.cfg.PacketParserRingBufferSize))
```
anubhabMajumdar, line 209: This log is erroneous. If `PacketParserRingBufferSize` is zero, what's the default size of this ring buffer?

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -152,15 +153,26 @@ func (p *packetParser) Compile(ctx context.Context) error {
...
+		cflags = append(cflags, "-DUSE_RING_BUFFER")
+		if p.cfg.PacketParserRingBufferSize > 0 {
+			cflags = append(cflags, fmt.Sprintf("-DRING_BUFFER_SIZE=%d", p.cfg.PacketParserRingBufferSize))
```
anubhabMajumdar: It seems dangerous to allow user to pass in any value here.

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -222,10 +234,22 @@ func (p *packetParser) Init() error {
...
-		return err
+	if p.cfg.EnablePacketParserRingBuffer {
+		p.l.Info("Initializing Ring Buffer reader")
+		var rb *ringbuf.Reader
+		rb, err = ringbuf.NewReader(objs.RetinaPacketparserEvents)
+		if err != nil {
+			p.l.Error("Error NewReader ringbuf", zap.Error(err))
+			return fmt.Errorf("failed to create ringbuf reader: %w", err)
+		}
+		p.reader = &ringBufReaderWrapper{reader: rb}
```
anubhabMajumdar, line 291: This interface needs to be updated to support both perf array and ring buffer - https://github.com/microsoft/retina/blob/f760861f20986773c1054927546d8cc74d115447/pkg/plugin/packetparser/types_linux.go#L104 .
Don't like that we are forcing the ringbuffer record in perf array record. I think long term we may default to ringbuf , so it would need change again.

anubhabMajumdar, line 291: Does this hold true for `ringbuffer` - https://github.com/microsoft/retina/blob/f760861f20986773c1054927546d8cc74d115447/pkg/plugin/packetparser/packetparser_linux.go#L695 ? If not, this loop will keep running even when there's no data.

deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -68,6 +68,8 @@ enableAnnotations: false
 bypassLookupIPOfInterest: true
 dataAggregationLevel: "high"
 monitorSockPath: "/var/run/cilium/monitor1_2.sock"
+enablePacketParserRingBuffer: false
```
anubhabMajumdar: Can you add comment about what this means, pros cons etc ? Also, an issue to update docs in future.

anubhabMajumdar: Also, based on the PR description 
```
Ring Buffers offer better performance and memory efficiency, especially on high-core systems, by using a shared buffer across CPUs rather than per-CPU buffers.
```
it seems like ring buffer is better for all scenarios, so should we make that the default?

pkg/config/config.go
SRodi: I agree with introducing an auto mode. However, since auto isn’t implemented yet, referencing it in the current PR may confuse users. I recommend we keep the current boolean for now (true/false) and introduce a new, explicit tri-state type, including auto, when the runtime decision logic is actually implemented.

pkg/plugin/packetparser/packetparser_linux.go
SRodi: Yes, the ring buffer reader also blocks on Read() until data arrives (or the reader is closed), same as perf. So the loop isn’t busy‑spinning; it will block inside Read() when there’s no data. The ringbuf docs describe this behavior as well.

The doc notes the blocking behavior in the "Userspace map reading / epoll" section: it says userspace can use epoll_* to wait for new data and that no CPU is consumed while waiting (the scheduler wakes it when data is available). That’s the mechanism behind a blocking read. [doc reference](https://docs.ebpf.io/linux/map-type/BPF_MAP_TYPE_RINGBUF/#epoll)

deploy/hubble/manifests/controller/helm/retina/values.yaml
SRodi: Created issue #2016 to update the docs. Also see the PR’s Additional Notes for why BPF_MAP_TYPE_PERF_EVENT_ARRAY remains the default.

pkg/plugin/packetparser/packetparser_linux.go
SRodi: This comment was addressed here a157bd43c64f9ff271ea9e5e6120f863506637c3

pkg/config/config.go
SRodi: Issue for the auto mode: #2017

pkg/plugin/packetparser/packetparser_linux.go
SRodi: This is now handled by `validateRingBufferSize`: if `PacketParserRingBufferSize` is `0`, it defaults to `8 MB` (matching what's defined in the BPF program `packetparser.c`). The size is normalized before this log, so the log shows the effective size.

pkg/telemetry/heartbeat_unix.go
```diff
@@ -7,17 +7,16 @@ package telemetry
...
-	"strings"
 
+	"github.com/microsoft/retina/pkg/utils"
 	"github.com/pkg/errors"
 )
 
 func KernelVersion(ctx context.Context) (string, error) {
-	cmd := exec.CommandContext(ctx, "uname", "-r")
-	output, err := cmd.CombinedOutput()
+	_ = ctx
```
rbtr: is this to silence the linter? I would have removed ctx from the args

```suggestion
func KernelVersion(context.Context) (string, error) {
```

pkg/utils/kernel_version_linux.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	major, minor, patch, err := ParseLinuxKernelRelease(release)
+	if err != nil {
+		return KernelVersion{Release: release}, err
+	}
+	return KernelVersion{Major: major, Minor: minor, Patch: patch, Release: release}, nil
+}
+
+// AtLeast reports whether the kernel version is >= the required version.
+func (v KernelVersion) AtLeast(major, minor, patch int) bool {
```
rbtr, line 47: perhaps we should lean on something like https://pkg.go.dev/golang.org/x/mod/semver

rbtr (dismissed):

pkg/config/config.go
anubhabMajumdar: Don't have to add the `auto` option in this PR, but can you not make it bool? Make it something that can be extended when support for `auto` comes in future PR.

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -52,6 +53,13 @@ import (
 
 //go:generate go run github.com/cilium/ebpf/cmd/bpf2go@master -cflags "-g -O2 -Wall -D__TARGET_ARCH_${GOARCH} -Wall" -target ${GOARCH} -type packet packetparser ./_cprog/packetparser.c -- -I../lib/_${GOARCH} -I../lib/common/libbpf/_src -I../lib/common/libbpf/_include/linux -I../lib/common/libbpf/_include/uapi/linux -I../lib/common/libbpf/_include/asm -I../filter/_cprog/ -I../conntrack/_cprog/
 var errNoOutgoingLinks = errors.New("could not determine any outgoing links")
+var errRingBufKernelTooOld = errors.New("ring buffer requires newer kernel")
+
+const (
+	ringBufMinKernelMajor = 5
+	ringBufMinKernelMinor = 8
+	ringBufMinKernelPatch = 0
+)
```
anubhabMajumdar: nit: can live in `types_linux.go`.

deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -68,6 +68,12 @@ enableAnnotations: false
 bypassLookupIPOfInterest: true
 dataAggregationLevel: "high"
 monitorSockPath: "/var/run/cilium/monitor1_2.sock"
+# Use BPF ring buffers (BPF_MAP_TYPE_RINGBUF) instead of BPF_PERF_EVENT_ARRAY.
+# Pros: typically lower per-event overhead at high event rates, simpler variable-sized records, often more consistent latency.
+# Cons: fixed-size pre-allocated (locked) memory that can be wasted on low traffic; events are dropped when the buffer fills (reserve fails).
+packetParserRingBuffer: "disabled"
```
anubhabMajumdar, line 75: Add in comments the possible values for `packetParserRingBuffer`.

deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -68,6 +68,12 @@ enableAnnotations: false
...
+# Ring buffer size in bytes (only used when packetParserRingBuffer="enabled").
```
anubhabMajumdar, line 76: You need to elaborate this more, especially the validation behavior. Current description makes it seem like whatever is added here is what the buffer size would be.

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -130,6 +132,57 @@ func (p *packetParser) Generate(ctx context.Context) error {
 	return nil
 }
 
+// validateRingBufferSize validates the ring buffer size and returns the adjusted size if necessary.
+// It also returns a reason for the adjustment, if any.
+func validateRingBufferSize(size uint32) (adjustedSize uint32, reason string) {
```
anubhabMajumdar: Couple of things:
1. This is more of a nit but would be great if we have a common validation function in the main controller. 
2. Minor adjustment is fine, but I would not set any default value or restrict to a value if size is greater than some constant. I would imagine we provide detailed description in `values.yaml` as to what the range should be, and if user sets it outside that range, we just return an error early,

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -152,15 +205,32 @@ func (p *packetParser) Compile(ctx context.Context) error {
...
 		libbpfIncludeUapiLinuxDir,
 		filterDir,
 		conntrackDir,
-	)
+	}
+
+	if p.cfg.PacketParserRingBuffer.IsEnabled() {
+		var reason string
+		p.cfg.PacketParserRingBufferSize, reason = validateRingBufferSize(p.cfg.PacketParserRingBufferSize)
+		if reason != "" {
```
anubhabMajumdar: This should bubble up to user. Otherwise, it's a black box for user who may keep increasing buffer size expecting different behavior.

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -174,6 +227,11 @@ func (p *packetParser) Init() error {
 		p.l.Warn("packet parser and latency plugin will not init because pod level is disabled")
 		return nil
 	}
+	if p.cfg.PacketParserRingBuffer.IsEnabled() {
+		if ringBufErr := ensureRingBufKernelSupported(); ringBufErr != nil {
```
anubhabMajumdar, line 231: I am guessing this will go away with `auto` mode?

SRodi: The check must remain so that users who explicitly configure `enabled` get an error if their environment is incompatible, rather than silently falling back to a mode they didn't ask for. However, a new branch of logic will be added alongside it to handle the loose requirements of `auto` mode.

Example:
```go
// Conceptual change for Auto mode implementation
shouldEnableRingBuffer := false

if p.cfg.PacketParserRingBuffer.IsAuto() {
    // In Auto mode: Check kernel support, but don't error out if unsupported.
    // Just fall back to perf buffer (shouldEnableRingBuffer stays false).
    if err := ensureRingBufKernelSupported(); err == nil {
        shouldEnableRingBuffer = true
    }
} else if p.cfg.PacketParserRingBuffer.IsEnabled() {
    // In Enabled mode: Strict check. Error out if unsupported (Current Logic).
    if ringBufErr := ensureRingBufKernelSupported(); ringBufErr != nil {
        return ringBufErr
    }
    shouldEnableRingBuffer = true
}
```

Issue #1893, opened 2025-10-06
Retina Windows Agent should (non-silently) fail if map callback registration fails


matthewige:
**Describe the bug**
Currently, only 1 consumer of the EVENTS_MAP can exist at a time on the system. If another process has already register, the callback in the retina windows agent currently silently fails. This can lead to a false-positive situation, where the pod / agent appears healthy, but it is not actually collecting the required metrics.

**To Reproduce**
Have a consume of the EVENTS_MAP prior to starting the retina windows pod

**Expected behavior**
Error message (possibly failed pod startup) occurs, to be noisy about the failure.

**Screenshots**
n/a
**Platform (please complete the following information):**

- OS: WindowsAnnual
- Kubernetes Version: [e.g. 1.22]
- Host: [e.g. AKS, KIND, self-host, etc]
- Retina Version:

**Additional context**
n/a

PR #1983, opened 2025-12-15
fix: pktmon etw silent failure

carlotaarvela:
# Description

Fix issue where the pktmon plugin silently fails to monitor packets when another Windows ETW consumer is active.
Use independent context.Background for health check and stream opening instead of inheriting parent context
Increased health check timeout from 30s to 60s to accommodate Windows pktmon startup
Added ERROR level logging for stream initialization failures

### Before:
Pod could crash with "context canceled" panic during shutdown
Stream initialization would hang if parent context was cancelled
Silent failures without visible error messages

### After:
Health check completes independently even during daemon shutdown
Stream initialization has independent timeout, unaffected by parent context
Errors properly logged at ERROR level for debugging
Pod exits gracefully instead of hanging or panicking

## Related Issue

https://github.com/microsoft/retina/issues/1893

## Checklist

1. Reproduce the Issue
Deployed Retina v1.0.0 (unfixed) with pktmon plugin on AKS Windows nodes. Pod entered Running state but silently failed to monitor packets with no error messages.

2. Apply the Fix
Built and pushed fixed image to ACR.

#### Scenario 1 - No Traffic:

Deployed pod with no network activity
Verified pod stays Running (2/2 containers)
Health check completed without timeout or crash
Pod remained stable indefinitely

#### Scenario 2 - Failure (pktmon unavailable):

Killed pktmon server process to simulate failure
Verified ERROR logs captured: "No connection could be made because the target machine actively refused it"
Pod exited gracefully instead of hanging
Kubernetes automatically restarted pod (CrashLoopBackOff)

#### Scenario 3 - Success with Traffic:

Fresh pod deployment with healthy pktmon
Verified metrics collection active: "emitting packets/bytes count metrics"
Collected metrics continuously for 70+ seconds across multiple endpoints
Pod handled graceful shutdown without panic when pktmon became unavailable

## Additional Notes

---

### DIFF
--- a/pkg/plugin/pktmon/pktmon_windows.go
+++ b/pkg/plugin/pktmon/pktmon_windows.go
@@ -31,16 +31,17 @@ var (
 	ErrUnexpectedExit          = errors.New("unexpected exit")
 	ErrNilGrpcClient           = errors.New("grpc client is nil")
 	ErrStreamHealthCheckFailed = errors.New("pktmon stream health check failed - likely another ETW consumer is active or stream is broken")
+	ErrTooManyNilFlows         = errors.New("received too many nil flows during health check - likely proto mismatch")
 
 	socket = "/temp/retina-pktmon.sock"
 )
 
 const (
-	name                       = "pktmon"
-	connectionRetryAttempts    = 5
-	eventChannelSize           = 1000
-	eventHealthCheckFirstEvent = 60 * time.Second
-	maxNilFlowsAllowed         = 5
+	name                     = "pktmon"
+	connectionRetryAttempts  = 5
+	eventChannelSize         = 1000
+	pktmonHealthCheckTimeout = 60 * time.Second
+	maxNilFlowsAllowed       = 5
 )
 
 type Plugin struct {
@@ -240,7 +241,7 @@ func (p *Plugin) verifyEventStream() error {
 	// Create an independent background context for the health check
 	// This is NOT a child of the parent ctx to avoid cancellation when the parent shuts down
 	// while the health check is still running
-	healthCtx, cancel := context.WithTimeout(context.Background(), eventHealthCheckFirstEvent)
+	healthCtx, cancel := context.WithTimeout(context.Background(), pktmonHealthCheckTimeout)
 	defer cancel()
 
 	// Create a channel to receive the result
@@ -259,7 +260,7 @@ func (p *Plugin) verifyEventStream() error {
 			if event.GetFlow() == nil {
 				nilFlowCount++
 				if nilFlowCount > maxNilFlowsAllowed {
-					resultCh <- errors.New("received too many nil flows during health check - likely proto mismatch")
+					resultCh <- fmt.Errorf("exceeded max nil flows allowed: %d: %w", maxNilFlowsAllowed, ErrTooManyNilFlows)
 					return
 				}
 				// Skip nil flows and wait for next event
@@ -278,13 +279,14 @@ func (p *Plugin) verifyEventStream() error {
 		// Timeout occurred - no events received, but stream connection is working.
 		// This could be due to low/no traffic rather than ETW registration failure.
 		p.l.Warn("no events received during health check",
-			zap.Duration("timeout", eventHealthCheckFirstEvent),
+			zap.Duration("timeout", pktmonHealthCheckTimeout),
 			zap.String("cause", "likely no network traffic or node is idle"))
 		return nil
 	case err := <-resultCh:
 		if err != nil {
 			// Stream error indicates ETW registration failure or connection issue
-			return errors.Wrapf(ErrStreamHealthCheckFailed, "%v", err)
+			// Use %w to properly wrap for ErrorIs traversal
+			return fmt.Errorf("%w: %w", ErrStreamHealthCheckFailed, err)
 		}
 		p.l.Info("pktmon event stream verified - events flowing normally")
 		return nil
--- a/pkg/plugin/pktmon/pktmon_windows_test.go
+++ b/pkg/plugin/pktmon/pktmon_windows_test.go
@@ -140,7 +140,7 @@ func Test_verifyEventStream_NoEventsTimeout(t *testing.T) {
 	elapsed := time.Since(start)
 
 	require.NoError(t, err, "verifyEventStream should return nil when no events are received (timeout due to no traffic)")
-	// Should timeout after approximately eventHealthCheckFirstEvent (60s)
+	// Should timeout after approximately pktmonHealthCheckTimeout (60s)
 	assert.GreaterOrEqual(t, elapsed, 59*time.Second, "should wait close to the 60s health check timeout")
 	assert.Less(t, elapsed, 62*time.Second, "should not wait significantly longer than the 60s timeout")
 }
@@ -191,7 +191,8 @@ func Test_verifyEventStream_TooManyNilFlows(t *testing.T) {
 	err := plugin.verifyEventStream()
 
 	require.Error(t, err, "verifyEventStream should fail when too many nil flows are received")
-	assert.ErrorContains(t, err, "too many nil flows", "error should mention too many nil flows")
+	// The error is wrapped as ErrStreamHealthCheckFailed which contains ErrTooManyNilFlows
+	assert.ErrorIs(t, err, ErrStreamHealthCheckFailed, "error should be a stream health check failure")
 }
 
 // Test_verifyEventStream_EventWithDelay tests successful reception of delayed non-nil flow
@@ -227,7 +228,7 @@ func Test_verifyEventStream_TimeoutExactlyAtLimit(t *testing.T) {
 	plugin := &Plugin{
 		l: NewTestLogger(),
 		stream: &MockGetFlowsClient{
-			blockFor: eventHealthCheckFirstEvent + 100*time.Millisecond,
+			blockFor: pktmonHealthCheckTimeout + 100*time.Millisecond,
 		},
 	}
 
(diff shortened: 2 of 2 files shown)


alexcastilio:
Could you add some tests to document the behavior and to avoid future regressions?

pkg/plugin/pktmon/pktmon_windows.go
```diff
@@ -212,6 +229,46 @@ func (p *Plugin) StartStream(ctx context.Context) error {
...
+		}
+		resultCh <- nil
+	}()
+
+	// Wait for either a successful event or timeout
+	select {
+	case <-healthCtx.Done():
+		return errors.Wrapf(ErrNoEventsReceived,
+			"no events received within %v - check if another ETW consumer is active or if there is network traffic to monitor",
+			eventHealthCheckFirstEvent)
```
alexcastilio: Why does this block return ErrNoEventsReceived wrapped with another string that mentions almost same message?
I suggest to change ErrNoEventsReceived to a complete message for that case and just return it or just return the desired error string directly.

pkg/plugin/pktmon/pktmon_windows.go
```diff
@@ -212,6 +229,46 @@ func (p *Plugin) StartStream(ctx context.Context) error {
...
+			resultCh <- errors.New("received nil flow during health check")
+			return
+		}
+		resultCh <- nil
+	}()
+
+	// Wait for either a successful event or timeout
+	select {
+	case <-healthCtx.Done():
+		return errors.Wrapf(ErrNoEventsReceived,
```
alexcastilio: If there's no traffic in node, an error will be triggered, right? Is that desired behavior? Is there any scenario where this can create an issue due to lack of traffic caused by any other reason?

pkg/plugin/pktmon/pktmon_windows.go
```diff
@@ -26,17 +27,20 @@ import (
...
 )
 
 const (
-	name                    = "pktmon"
-	connectionRetryAttempts = 5
-	eventChannelSize        = 1000
+	name                       = "pktmon"
+	connectionRetryAttempts    = 5
+	eventChannelSize           = 1000
+	eventHealthCheckTimeout    = 30 * time.Second
```
alexcastilio: Where is `eventHealthCheckTimeout` being used?

pkg/plugin/pktmon/pktmon_windows_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	start := time.Now()
+	err := plugin.verifyEventStream(ctx)
+	elapsed := time.Since(start)
+
+	require.Error(t, err, "verifyEventStream should fail when no events are received within timeout")
+	require.ErrorContains(t, err, "no events received", "error should mention no events received")
+	// Should timeout around eventHealthCheckFirstEvent (10s)
+	// But our context only allows 5s so we timeout on context
+	assert.GreaterOrEqual(t, elapsed, 4*time.Second, "should respect timeout")
```
alexcastilio: Why asserting if elapsed >= 4s?

pkg/plugin/pktmon/pktmon_windows_test.go
```diff
@@ -0,0 +1,302 @@
...
+		l: NewTestLogger(),
+		stream: &MockGetFlowsClient{
+			blockCh: make(chan struct{}), // Block forever
+		},
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+	defer cancel()
+
+	err := plugin.verifyEventStream(ctx)
```
alexcastilio: Consider adding a comment explaining that healthCtx is being canceled because it's created as a child of original ctx that gets timed out after 100ms

pkg/plugin/pktmon/pktmon_windows.go
```diff
@@ -212,6 +229,46 @@ func (p *Plugin) StartStream(ctx context.Context) error {
...
+	}()
+
+	// Wait for either a successful event or timeout
+	select {
+	case <-healthCtx.Done():
+		return errors.Wrapf(ErrNoEventsReceived,
+			"no events received within %v - check if another ETW consumer is active or if there is network traffic to monitor",
+			eventHealthCheckFirstEvent)
+	case err := <-resultCh:
+		if err != nil {
```
alexcastilio, line 286: is there any situation where nil flows could be generated that we would not want to fail? Or is it definitely always a problem?

if event.GetFlow() == nil, this fails straight away. 

Would we want to just ignore nil flows and maybe wait for a non nil flow eventually come?

pkg/plugin/pktmon/pktmon_windows_test.go
```diff
@@ -0,0 +1,302 @@
...
+	start := time.Now()
+	err := plugin.verifyEventStream(ctx)
+	elapsed := time.Since(start)
+
+	// Since blockFor causes recv to sleep, then returns EOF (no events),
+	// this will timeout after eventHealthCheckFirstEvent
+	// So we expect an error, but we're testing the timeout enforcement
+	require.Error(t, err, "should get timeout error")
+	assert.GreaterOrEqual(t, elapsed, 2*time.Second, "should wait at least for blockFor duration")
+}
```
alexcastilio: Is this testing reception of delayed nil event? Shouldn't it be testing delayed non-nil event?

pkg/plugin/pktmon/pktmon_windows_test.go
```diff
@@ -0,0 +1,302 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	err := plugin.verifyEventStream(ctx)
+
+	require.Error(t, err, "should get nil flow error")
+	assert.Equal(t, 1, plugin.stream.(*MockGetFlowsClient).index,
+		"should only consume exactly one event before returning error")
+}
```
alexcastilio, line 266: why is it important to assert that only one event is consumed? This seams to me internal implementation details that's not relevant to expected behavior, so probably it should not be tested.

pkg/plugin/pktmon/pktmon_windows_test.go
```diff
@@ -0,0 +1,302 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	err := plugin.verifyEventStream(ctx)
+
+	require.Error(t, err, "verifyEventStream should fail")
+	errMsg := err.Error()
+	assert.Contains(t, errMsg, "no events received from pktmon", "error should mention pktmon")
+	assert.Contains(t, errMsg, "ETW consumer", "error should mention ETW")
+	assert.Contains(t, errMsg, "network traffic", "error should mention network traffic")
```
alexcastilio: this seems too tightly coupled with exact text of error message. Is this important? Isn't assertion on sentinel error enough, for example?

pkg/plugin/pktmon/pktmon_windows.go
carlotaarvela: Removed

pkg/plugin/pktmon/pktmon_windows.go
carlotaarvela: You are right, updated this logic to distinguish between 2 different scenarios the gRPC stream connection failure and the no traffic scenario

pkg/plugin/pktmon/pktmon_windows.go
carlotaarvela: Have updated this to match the logic changes

pkg/plugin/pktmon/pktmon_windows.go
carlotaarvela: I have updated the function to ignore the nil flows and wait for a valid one. Also increased the timeout to 30 seconds for scenarios where the node is taking longer to finish setup

pkg/plugin/pktmon/pktmon_windows_test.go
carlotaarvela: Have updated this check to verify the timeouts

pkg/plugin/pktmon/pktmon_windows_test.go
carlotaarvela: Done

pkg/plugin/pktmon/pktmon_windows_test.go
carlotaarvela: Update the test to verify the health check can handle delated successful flows

pkg/plugin/pktmon/pktmon_windows_test.go
carlotaarvela: Update this to be more of a sanity check to make sure it succeeds when received a vali flow

pkg/plugin/pktmon/pktmon_windows_test.go
carlotaarvela: Update this to check the scenario where there is no traffic but the connection is successful

kamilprz (approved):
Seems ok to me, but lets wait for @alexcastilio approval

pkg/plugin/pktmon/pktmon_windows.go
alexcastilio: do you mean 60s? (I see `eventHealthCheckFirstEvent = 60 * time.Second` on top of file)

pkg/plugin/pktmon/pktmon_windows.go
```diff
@@ -26,17 +27,20 @@ import (
...
 )
 
 const (
-	name                    = "pktmon"
-	connectionRetryAttempts = 5
-	eventChannelSize        = 1000
+	name                       = "pktmon"
+	connectionRetryAttempts    = 5
+	eventChannelSize           = 1000
+	eventHealthCheckFirstEvent = 60 * time.Second
```
alexcastilio: eventHealthCheckFirstEvent this is the timeout for the context, right? Id so, I think we need a more meaningful name.

pkg/plugin/pktmon/pktmon_windows.go
```diff
@@ -26,17 +27,20 @@ import (
...
+	maxNilFlowsAllowed         = 5
```
alexcastilio: Is 5 an arbitrary choice?

pkg/plugin/pktmon/pktmon_windows_test.go
```diff
@@ -0,0 +1,303 @@
...
+		stream: &MockGetFlowsClient{
+			// More nil flows than allowed
+			responses: make([]interface{}, maxNilFlowsAllowed+2),
+		},
+	}
+
+	err := plugin.verifyEventStream()
+
+	require.Error(t, err, "verifyEventStream should fail when too many nil flows are received")
+	assert.ErrorContains(t, err, "too many nil flows", "error should mention too many nil flows")
```
alexcastilio: Instead of check the error string, I suggest using assertion on sentinel error

alexcastilio (commented):
Added some comments

pkg/plugin/pktmon/pktmon_windows.go
carlotaarvela: Yes, I did test the workflow and it seems long enough to reduce the false positives. We can update this later if needed

PR #1928, opened 2025-11-04
chore(ebpf): update bpf header files

nddq:
This patch updates the bpf header files from upstream libbpf, as well as introduce Make commands to help updating these files in the future. Also update the Dockerfiles for building the agent and generating Go codes.

# Description

Please provide a brief description of the changes made in this pull request.

## Related Issue

If this pull request is related to any issue, please mention it here. Additionally, make sure that the issue is assigned to you before submitting this pull request.

## Checklist

- [ ] I have read the [contributing documentation](https://retina.sh/docs/Contributing/overview).
- [ ] I signed and signed-off the commits (`git commit -S -s ...`). See [this documentation](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) on signing commits.
- [ ] I have correctly attributed the author(s) of the code.
- [ ] I have tested the changes locally.
- [ ] I have followed the project's style guidelines.
- [ ] I have updated the documentation, if necessary.
- [ ] I have added tests, if applicable.

## Screenshots (if applicable) or Testing Completed

Please add any relevant screenshots or GIFs to showcase the changes made.

## Additional Notes

Add any additional notes or context about the pull request here.

---

Please refer to the [CONTRIBUTING.md](../CONTRIBUTING.md) file for more information on how to contribute to this project.

PR #1984, opened 2025-12-19
feat(ebpf): generate vmlinux.h in init container

### LINKED ISSUE #1777: compile eBPF against a runtime-generated vmlinux.h
Currently Retina compiles its eBPF programs during pod startup, however it links with a static `vmlinux.h` file that is checked in the repo.
This is wrong because it risks reading garbage data at incorrect kernel structure offsets.

One correct way is to generate the header at compile time to match the runtime environment's kernel:
`bpftool btf dump file /sys/kernel/btf/vmlinux format c > vmlinux.,h`

SRodi:
## Description
This PR finalizes Retina's runtime `vmlinux.h` workflow for CO-RE-safe eBPF compilation while keeping the main agent runtime image minimal.

`vmlinux.h` is generated from host BTF via `bpftool` in the **init container**, persisted under `/tmp/retina/include`, and consumed by plugin compile paths. The loader now caches generation metadata and skips regeneration when the cached header matches the current kernel release, regenerating automatically on kernel drift.

This avoids static-header CO-RE relocation issues and removes the need for `bpftool` in the main agent runtime image.

## What Changed
- Added runtime `vmlinux.h` support in `pkg/loader`:
  - `GenerateVmlinuxH(ctx, outputDir)`
  - runtime path constants/helpers (`DefaultVmlinuxHeaderDir`, `VmlinuxHeaderDir`, `VmlinuxHeaderPath`)
  - `PrepareVmlinuxH(ctx)` for ensure-or-generate behavior
- Added cache drift detection in `pkg/loader`:
  - stores kernel release metadata in `vmlinux.kernel.release`
  - reuses cached `vmlinux.h` only when current kernel release matches cached value
  - regenerates `vmlinux.h` on mismatch/missing cache metadata
- Added generation duration logging in loader (`Generated vmlinux.h` with elapsed time) for startup observability.
- Added loader unit test coverage for runtime header generation (`pkg/loader/vmlinux_test.go`).
- Added init-container generation call in `init/retina/main_linux.go` so header creation happens before the agent starts.
- Removed plugin-manager-side `vmlinux.h` generation from agent startup.
- Updated plugin compile paths to include runtime header directory (dropreason, packetforward, packetparser) without per-plugin `GenerateVmlinuxH` calls.
- Updated Helm daemonsets (standard + hubble) to mount shared `/tmp` into both init and agent containers so generated headers are available to the agent.
- Updated container packaging so `bpftool` is available to init runtime but not bundled into the agent runtime bin set:
  - init image receives `bpftool`
  - agent `/bin` bundle excludes `bpftool`

## Operational Notes
- Init logs now include runtime `vmlinux.h` preparation and generation duration.
- Agent compiles plugins using `-I/tmp/retina/include` (or `RETINA_VMLINUX_HEADER_DIR` override) without requiring `bpftool` in agent runtime.
- Cache behavior:
  - same kernel release: skip regeneration
  - kernel release drift: regenerate and refresh metadata

## Commit Summary (PR)
- fe478053 feat(bpf): generate vmlinux.h at runtime
- a79b499f fix linting
- 3bf8411b refactor(ebpf): centralize vmlinux.h generation and share runtime include path
- 79d851fe generate vmlinux.h in init and remove bpftool from agent runtime
- 65127ed3 cache vmlinux.h by kernel release and log generation duration
- 0cead17c return accurate vmlinux generation error and defer logging to caller

Fixes #1777

## Related Issue

## Checklist

Container retina-init logs + curl to show retina metrics are generated:

image

Check `vmlinux.h` exist under `/tmp/retina/include` within the retina-agent Pod:

image

Debug logs to show plugins compile using header location `/tmp/retina/include`:

Screenshot 2026-04-02 165513

## Additional Notes

---

### DIFF
--- a/pkg/loader/vmlinux.go
+++ b/pkg/loader/vmlinux.go
@@ -26,6 +26,11 @@ const (
 	kernelReleasePath            = "/proc/sys/kernel/osrelease"
 )
 
+var (
+	errKernelReleaseEmpty       = errors.New("kernel release is empty")
+	errCachedKernelReleaseEmpty = errors.New("cached kernel release is empty")
+)
+
 // VmlinuxHeaderDir returns the runtime directory where vmlinux.h is expected.
 func VmlinuxHeaderDir() string {
 	dir := strings.TrimSpace(os.Getenv(VmlinuxHeaderDirEnv))
@@ -111,21 +116,22 @@ func currentKernelRelease() (string, error) {
 
 	release := strings.TrimSpace(string(b))
 	if release == "" {
-		return "", fmt.Errorf("kernel release from %s is empty", kernelReleasePath)
+		return "", fmt.Errorf("%w: %s", errKernelReleaseEmpty, kernelReleasePath)
 	}
 
 	return release, nil
 }
 
 func readCachedKernelRelease(outputDir string) (string, error) {
-	b, err := os.ReadFile(filepath.Join(outputDir, VmlinuxKernelReleaseFileName))
+	metaPath := filepath.Join(outputDir, VmlinuxKernelReleaseFileName)
+	b, err := os.ReadFile(metaPath)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("failed to read cached kernel release %s: %w", metaPath, err)
 	}
 
 	release := strings.TrimSpace(string(b))
 	if release == "" {
-		return "", errors.New("cached kernel release is empty")
+		return "", errCachedKernelReleaseEmpty
 	}
 
 	return release, nil
@@ -133,7 +139,7 @@ func readCachedKernelRelease(outputDir string) (string, error) {
 
 func writeCachedKernelRelease(outputDir, kernelRelease string) error {
 	metaPath := filepath.Join(outputDir, VmlinuxKernelReleaseFileName)
-	if err := os.WriteFile(metaPath, []byte(kernelRelease+"\n"), 0o644); err != nil {
+	if err := os.WriteFile(metaPath, []byte(kernelRelease+"\n"), 0o600); err != nil {
 		return fmt.Errorf("failed to write cached kernel release %s: %w", metaPath, err)
 	}
 
--- a/pkg/plugin/dropreason/dropreason_linux.go
+++ b/pkg/plugin/dropreason/dropreason_linux.go
@@ -121,7 +121,11 @@ func (dr *dropReason) Compile(ctx context.Context) error {
 	runtimeIncludeDir := "-I" + loader.VmlinuxHeaderDir()
 
 	// Keep target as bpf, otherwise clang compilation yields bpf object that elf reader cannot load.
-	err = loader.CompileEbpf(ctx, "-target", "bpf", "-Wall", targetArch, "-g", "-O2", "-c", bpfSourceFile, "-o", bpfOutputFile, runtimeIncludeDir, includeDir, libbpfDir, filterDir)
+	err = loader.CompileEbpf(
+		ctx,
+		"-target", "bpf", "-Wall", targetArch, "-g", "-O2", "-c", bpfSourceFile,
+		"-o", bpfOutputFile, runtimeIncludeDir, includeDir, libbpfDir, filterDir,
+	)
 	if err != nil {
 		return errors.Wrap(err, "unable to compile eBPF code")
 	}
--- a/init/retina/main_linux.go
+++ b/init/retina/main_linux.go
@@ -67,7 +67,11 @@ func run(args ...string) error {
 
 	runtimeHeaderDir, headerErr := loader.PrepareVmlinuxH(context.Background())
 	if headerErr != nil {
-		l.Warn("failed to prepare runtime vmlinux.h in init container, falling back to static headers", zap.String("path", runtimeHeaderDir), zap.Error(headerErr))
+		l.Warn(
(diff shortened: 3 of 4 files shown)


pkg/loader/vmlinux.go
```diff
@@ -0,0 +1,50 @@
...
+	vmlinuxPath := filepath.Join(outputDir, "vmlinux.h")
+
+	if err := os.MkdirAll(outputDir, 0o755); err != nil {
+		return fmt.Errorf("failed to create output directory: %w", err)
+	}
+
+	// Check if vmlinux.h already exists to avoid regenerating it unnecessarily?
+	// However, if the pod restarts on a different node (unlikely for same pod instance but possible if volume persisted?),
+	// or if we want to be sure.
+	// Given the startup time is not critical and it's fast, let's generate it.
```
rbtr: how fast?

rbtr: i'm thinking we should write it to disk (/tmp/retina like you've already referenced below) and skip regenerating it. but what happens if the kernel is updated? how can we tell when the kernel has drifted from our cached vmlinux.h? if there was an easy way to do that, i would say fully persist it to disk and only regenerate when that happens...but i'm not sure we can do that.

rbtr (approved):
I think this is a good idea already, but there's same ambiguity to me around when/if the header could be regenerated instead of always. Kernel drift underneath us is a concern currently, since we can live-upgrade kernels.

pkg/plugin/dropreason/dropreason_linux.go
```diff
@@ -117,8 +117,16 @@ func (dr *dropReason) Compile(ctx context.Context) error {
 	if arch == "arm64" {
 		targetArch = "-D__TARGET_ARCH_arm64"
 	}
+
+	// Generate vmlinux.h
+	runtimeHeaderDir := "/tmp/retina/include"
+	if err = loader.GenerateVmlinuxH(ctx, runtimeHeaderDir); err != nil {
+		dr.l.Warn("Failed to generate vmlinux.h, falling back to static headers", zap.Error(err))
+	}
```
anubhabMajumdar: I don't think we should invoke it from every plugin. Seems like `GenerateVmlinuxH` should be invoked before plugins are started, and the path should be a constant plugin can refer to.

pkg/loader/vmlinux.go
```diff
@@ -0,0 +1,50 @@
...
+	cmd.Stdout = outfile
+	// Capture stderr for debugging
+	cmd.Stderr = os.Stderr
+
+	l.Info("Generating vmlinux.h", zap.String("path", vmlinuxPath))
+	if err := cmd.Run(); err != nil {
+		l.Error("Failed to generate vmlinux.h", zap.Error(err))
+		// If bpftool fails (e.g. /sys/kernel/btf/vmlinux doesn't exist), we might want to fallback or error out.
+		// If it fails, the compilation will likely fail later if we rely on this header.
+		return fmt.Errorf("failed to run bpftool: %w", err)
```
anubhabMajumdar: I like this error message better - `l.Error("Failed to generate vmlinux.h", zap.Error(err))` . Seems more accurate as bpftool can run and fail for myriad reasons.
Also, no need to log it here as caller would probably do that.

pkg/plugin/dropreason/dropreason_linux.go
```diff
@@ -117,8 +117,16 @@ func (dr *dropReason) Compile(ctx context.Context) error {
 	if arch == "arm64" {
 		targetArch = "-D__TARGET_ARCH_arm64"
 	}
+
+	// Generate vmlinux.h
+	runtimeHeaderDir := "/tmp/retina/include"
+	if err = loader.GenerateVmlinuxH(ctx, runtimeHeaderDir); err != nil {
+		dr.l.Warn("Failed to generate vmlinux.h, falling back to static headers", zap.Error(err))
```
anubhabMajumdar: I am curious about the consequence of this compilation failure. Seems like subsequent `datapath` errors can be attributed to this failure, but we won't have any idea if the warning log rotates after some time.

controller/Dockerfile
```diff
@@ -90,7 +91,7 @@ RUN tdnf install -y \
     ca-certificates \
     tar
 RUN mkdir -p /tmp/bin
-RUN arr="clang tcpdump ip ss iptables-legacy iptables-legacy-save iptables-nft iptables-nft-save cp uname" ;\
+RUN arr="clang bpftool tcpdump ip ss iptables-legacy iptables-legacy-save iptables-nft iptables-nft-save cp uname" ;\
```
anubhabMajumdar: I would rather have an init image with bpftool that can compile the vmlinux, and we can mount the directory to the agent image (optional). This way agent image only has the essentials.

invidian:
Any updates on this PR since #1928 got closed?

pkg/loader/vmlinux.go
SRodi: For "how fast?":
I added timing logs around bpftool execution, so startup logs now report actual vmlinux.h generation duration per node instead of assumptions. (see screenshot, it's ~121ms)

For cache/drift question:
We now persist the kernel release alongside vmlinux.h and reuse cached headers only when it matches the current kernel release. If it differs, vmlinux.h is regenerated automatically on the next init/startup path (for example after pod restart/recreate).

Issue #1958, opened 2025-12-02
Packetparser metrics missing when Cilium uses TCX (TC/TCX attachment conflict)


aanchal22:
**Describe the bug**

Retina's packetparser plugin is unable to collect metrics (e.g., `networkobservability_adv_forward_bytes`) on clusters where Cilium has TCX enabled. The packetparser plugin relies on traditional TC (Traffic Control) with clsact qdiscs for BPF program attachment, which conflicts with Cilium's TCX-based attachment on the same network interfaces.

Retina was working correctly before TCX was enabled on the cluster. No explicit errors appear in the logs - the failure is silent.

**To Reproduce**

1. Deploy a Kubernetes cluster with kernel >= 6.6 (e.g., 6.8.0-1039-aws on Ubuntu 22.04)
2. Install Cilium 1.16.x (which uses TCX by default on supported kernels)
3. Install Retina 0.0.36 with packetparser plugin enabled
4. Query for packetparser metrics (e.g., `networkobservability_adv_forward_bytes`)
5. Observe that the metrics are not being collected/scraped

**Expected behavior**

Retina should successfully collect packetparser metrics (`networkobservability_adv_forward_bytes`, etc.) regardless of whether Cilium is using TC or TCX for its BPF programs.

**Screenshots**

N/A - No visual errors. The issue manifests as missing metrics in Prometheus/Grafana.

**Platform (please complete the following information):**

- OS: Ubuntu 22.04.5 LTS
- Kernel: 6.8.0-1039-aws
- Kubernetes Version: 1.32
- Host: AWS EKS
- Retina Version: 0.0.36
- Cilium Version: 1.16.10

**Additional context**

**Root Cause Analysis:**
- Cilium 1.16.x uses TCX (TC eXpress) by default for BPF program attachment on kernels >= 6.6
- Retina's packetparser currently only supports traditional TC with clsact qdiscs (`pkg/plugin/packetparser/packetparser_linux.go`)
- When Cilium attaches its BPF programs via TCX to network interfaces, Retina's TC-based attachment either fails silently or doesn't receive packet events due to some conflicts.
- Retina does not currently have TCX support to coexist with Cilium's TCX programs

**Relevant Code:**
The packetparser attaches BPF programs using TC clsact qdiscs at `pkg/plugin/packetparser/packetparser_linux.go`. It creates a clsact qdisc and attaches ingress/egress BPF filters, which may not work correctly when Cilium is using TCX on the same interfaces.

[...]

PR #1994, opened 2026-01-09
feat: Add TCX support to the packetparser plugin

bugie:
# Description

<html><head></head><body>
<h2>Add TCX support to packetparser plugin</h2>
<p>Adds support for attaching BPF programs via <strong>TCX (TC eXpress)</strong>, the newer kernel hook available in Linux 6.6+, as an alternative to the traditional <code>clsact</code> qdisc approach. TCX provides lower overhead and better multi-program chaining semantics.</p>
<h3>Approach</h3>
<p>Rather than a separate plugin, TCX attachment lives inside the existing <code>packetparser</code> plugin. The existing C programs (<code>SEC("classifier_...")</code>) require <strong>no changes</strong> — they are <code>BPF_PROG_TYPE_SCHED_CLS</code> programs and can be attached via either mechanism.</p>
<h3>Configuration</h3>
<p>A new <code>enableTCX</code> config field gates the feature:</p>

Value | Behaviour
-- | --
off | Always use traditional TC (default-safe)
auto | Probe kernel support; use TCX if available, fall back to TC

<p>Defaults to <code>auto</code> when unset.</p>
<h3>Implementation details</h3>
<ul>
<li><strong>Kernel probe</strong> — <code>isTCXSupported()</code> creates a minimal throwaway <code>SchedCLS</code> program and attempts a TCX attach to loopback to detect kernel support without relying on the plugin's own loaded programs.</li>
<li><strong>Unified attachment map</strong> — <code>tcMap</code>/<code>tcValue</code> replaced with <code>attachmentMap</code>/<code>attachmentValue</code>, which tracks whether each interface was attached via TC or TCX and cleans up correctly on removal.</li>
<li><strong><code>createQdiscAndAttach</code></strong> delegates to <code>attachViaTCX</code> or <code>attachViaTC</code> depending on <code>p.tcxSupported</code>, set once during <code>Init()</code>.</li>
</ul></body></html>

## Related Issue

https://github.com/microsoft/retina/issues/1958

## Checklist

## Additional Notes

---

### DIFF
--- a/deploy/hubble/manifests/controller/helm/retina/values.yaml
+++ b/deploy/hubble/manifests/controller/helm/retina/values.yaml
@@ -101,6 +101,11 @@ logLevel: info
 enabledPlugin_linux: '["linuxutil","packetforward","packetparser","dns", "dropreason"]'
 enabledPlugin_win: '["hnsstats"]'
 
+# TCX Support
+# Supported Values: "auto", "off"
+# Default: auto
+enableTCX: "auto"
+
 enableTelemetry: false
 
 # Interval, in duration, to scrape/publish metrics.
--- a/deploy/standard/manifests/controller/helm/retina/templates/configmap.yaml
+++ b/deploy/standard/manifests/controller/helm/retina/templates/configmap.yaml
@@ -17,6 +17,7 @@ data:
     enabledPlugin: {{ .Values.enabledPlugin_linux }}
     metricsInterval: {{ .Values.metricsInterval }}
     metricsIntervalDuration: {{ .Values.metricsIntervalDuration }}
+    enableTCX: {{ .Values.enableTCX }}
     enableTelemetry: {{ .Values.enableTelemetry }}
     enablePodLevel: {{ .Values.enablePodLevel }}
     enableConntrackMetrics: {{ .Values.enableConntrackMetrics }}


pkg/config/config.go
```diff
@@ -16,6 +16,18 @@ import (
...
 type Level int
 
+// TCXMode controls whether TCX (TC eXpress) attachment is used for packetparser.
+type TCXMode string
+
+const (
+	// TCXModeAuto detects kernel support and uses TCX if available, falling back to TC.
+	TCXModeAuto TCXMode = "auto"
+	// TCXModeAlways requires TCX and fails if not supported.
+	TCXModeAlways TCXMode = "always"
```
nddq: what is the benefit of an `always` mode here?

bugie: Just a way to force TCX. Maybe not needed but just in case there is a need to fail instead of falling back. This incorporated the feedback from @anubhabMajumdar

anubhabMajumdar:
> > I think we can simplify this significantly. `SEC("tcx/ingress")` and `SEC("classifier")` both compile to `BPF_PROG_TYPE_SCHED_CLS`, the section name is just ELF metadata, not bytecode. `link.AttachTCX()` only needs a `SchedCLS` program FD and the caller-supplied attach type. The kernel TCX patch itself says programs can migrate from `cls_bpf` to TCX ["without needing to change the program's source code, just the BPF loader mechanics for attaching."](https://lore.kernel.org/bpf/20230707172455.7634-3-daniel@iogearbox.net/)
> > So, we should be able to reuse the existing `packetparser` BPF program as-is and just branch the Go-side attachment logic to call link.AttachTCX() instead of the `netlink/qdisc` path when TCX is selected. That would let us avoid the new `/packetparsertcx` directory and the duplicated bpf2go codegen entirely and cut down a lot of the duplicated Go code.
> > For reference, this is exactly what we're doing in the experimental Rust rewrite, single SchedCLS program, try TCX first, fall back to legacy TC on the Go/loader side: https://github.com/microsoft/retina/blob/main/experimental/plugins/packetparser/userspace/src/loader.rs#L39-L67
> 
> @nddq , Thank you for the review and feedback. I am all for simplification. I feel that my initial PR with [commit](https://github.com/bugie/retina/commit/cb952a3770a46185a57a26bebb435fe9c8b61924) was doing what you are suggesting but the feedback I received had me go in this direction, to break it out into a separate plugin. Can you have a look at the previous commit?

@bugie Reading Quang's comments, it seems like you do not need any C code change at all, the existing C code can simply attach to TCX. If that's the case, then maybe we can keep it in the existing `Packetparser` module. 

- we keep the config to gatekeep the feature
- depending on config, load accordingly

pkg/config/config.go
anubhabMajumdar: I think `auto` and `off` would suffice?

bugie: Satisfies our use case, I can remove `always` if there is not a need to force TCX.

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -300,9 +302,23 @@ func (p *packetParser) Init() error {
...
-	p.tcMap = &sync.Map{}
+	p.attachmentMap = &sync.Map{}
 	p.interfaceLockMap = &sync.Map{}
 
+	// Resolve TCX support based on config.
+	switch p.cfg.EnableTCX {
+	case kcfg.TCXModeOff:
+		p.tcxSupported = false
+		p.l.Info("EnableTCX=off: will use traditional TC attachment")
+	case kcfg.TCXModeAuto:
```
nddq, line 319: missing a `default` case here

pkg/config/config.go
```diff
@@ -175,6 +186,11 @@ func GetConfig(cfgFilename string) (*Config, error) {
 		config.DataSamplingRate = DefaultSamplingRate
 	}
 
+	// Default EnableTCX to "auto" if unset.
+	if config.EnableTCX == "" {
+		config.EnableTCX = TCXModeAuto
+	}
+
```
nddq: I think we need some config validation here as well to fail fast on invalid values

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -448,6 +467,66 @@ func (p *packetParser) clean(rtnl nltc, qdisc *tc.Object) {
...
+	var loopback netlink.Link
+	for _, l := range links {
+		if l.Attrs().Name == "lo" {
+			loopback = l
+			break
+		}
+	}
+	if loopback == nil {
+		return false
+	}
```
nddq: Can we use https://pkg.go.dev/github.com/vishvananda/netlink#Handle.LinkByName here instead of iterating through all interfaces?

pkg/plugin/packetparser/_cprog/packetparser.h
```diff
@@ -6,6 +6,8 @@
 #define MAX_TCP_OPTIONS_LEN 40
 // tc-bpf return code to execute the next tc-bpf program.
 #define TC_ACT_UNSPEC   (-1)
+// tcx return code to pass to the next program in the chain (kernel 6.6+).
+#define TCX_NEXT        (-1)
```
nddq: Are we using this value anywhere?

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/cilium/cilium/api/v1/flow"
 	v1 "github.com/cilium/cilium/pkg/hubble/api/v1"
 	"github.com/cilium/ebpf"
+	asm "github.com/cilium/ebpf/asm"
```
nddq: might be a formatter thing, idk why it has to have this alias when the package name  is already asm

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -484,9 +567,70 @@ func (p *packetParser) endpointWatcherCallbackFn(obj interface{}) {
...
+	default:
+		p.l.Error("Unknown interface type for TCX", zap.String("interface type", string(ifaceType)))
+		return
+	}
+
+	ingressLink, err := link.AttachTCX(link.TCXOptions{
+		Program:   ingressProgram,
+		Attach:    ebpf.AttachTCXIngress,
+		Interface: iface.Index,
+		Anchor:    link.Head(),
```
nddq, line 606: we should document somewhere that we want Retina's tcx program to be first in the chain so that we can see all packets + we are not making any packet decision, so it is safe to use with other programs i.e. Cilium that might enforce some kind of traffic policy.

pkg/plugin/packetparser/packetparser_linux.go
bugie: Default case added.

pkg/config/config.go
bugie: Added fail fast and a test.

pkg/plugin/packetparser/packetparser_linux.go
bugie: Done

pkg/plugin/packetparser/_cprog/packetparser.h
bugie: Good catch, old code. Removed.

pkg/plugin/packetparser/packetparser_linux.go
bugie: yup

pkg/plugin/packetparser/packetparser_linux.go
bugie: Comments added above the assignment

pkg/config/config_test.go
```diff
@@ -70,6 +71,65 @@ func TestDecodeLevelHook(t *testing.T) {
 	}
 }
 
+func TestGetConfig_EnableTCX(t *testing.T) {
```
nddq, line 73: This test is not actually testing the switching logic in `GetConfig()` but rather just duplicating the logic from there. I suggest you look at other tests in this file and their patterns (write a YAML file with specific values, call `GetConfig()`, assert on the result)

bugie: Fair point. I wanted something quick and lightweight but it is a bit too light.

bugie: @nddq I have updated the test with yaml files. Let me know if there is anything else.

PR #1997, opened 2026-01-13
fix: CVE-2013-3900 - Updates to Windows image

kamilprz:
# Description

This PR tackles CVE-2013-3900 which is present on the Windows images. The Windows registry must be updated to mitigate the CVE, thus we had to separate the our Windows/Linux build process, so that Windows runs on a Windows host VM.

The Windows image build step is now separate into two steps. The binaries are built as part of a separate Makefile target. The GitHub Action runs this on an Ubuntu host (intentionally - its faster than on Windows). The Windows 2019/2022 image build actions then use those binaries to complete the build on their respective Windows based hosts.

## Related Issue

https://msrc.microsoft.com/update-guide/vulnerability/CVE-2013-3900

## Checklist

- [] I have updated the documentation, if necessary.
- [] I have added tests, if applicable.

Vulnerability scan on pre-fix Windows 2022. `ghcr.io/microsoft/retina/retina-agent:4895b43-windows-ltsc2022-amd64`
image

Vulnerability scan after the fix. Doesn't detect the same CVE. `ghcr.io/kamilprz/retina/retina-agent:357857e-windows-ltsc2022-amd64`
image

## Additional Notes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -281,7 +281,7 @@ retina-image: ## build the retina linux container image.
 				TARGET=$$target; \
 	done
 
-retina-image-win: ## build the retina Windows container image.
+retina-image-win: build-windows-binaries ## build the retina Windows container image.
 # The Windows images are built on a corresponding Windows host without buildx.
 # This is done to mitigate CVE-2013-3900.
 	for year in $(WINDOWS_YEARS); do \


.github/workflows/images.yaml
```diff
@@ -62,27 +62,59 @@ jobs:
...
+          path: output/windows_amd64/
+          retention-days: 1
+
+  retina-image-win-2019:
+    name: Build Agent Image - Windows 2019 
+    needs: build-windows-binaries
+    runs-on: windows-2019
 
     strategy:
       matrix:
```
apontejaj, line 104: The matrix strategy is used when you want to reuse code for different configurations.
In this case there is only one config `windows/amd64/2019`, so there is very little point having this here.
I would suggest to remove the strategy and replace the values in the corresponding placeholders.

.github/workflows/images.yaml
```diff
@@ -91,34 +123,89 @@ jobs:
...
+        env:
+          IS_MERGE_GROUP: ${{ github.event_name == 'merge_group' }}
+
+
+  retina-image-win-2022:
+    name: Build Agent Image - Windows 2022 
+    needs: build-windows-binaries
+    runs-on: windows-2022
+    
+    strategy:
```
apontejaj: The matrix strategy is used when you want to reuse code for different configurations.
In this case there is only one config `windows/amd64/2020`, so there is very little point having this here.
I would suggest to remove the strategy and replace the values in the corresponding placeholders.

.github/workflows/images.yaml
kamilprz: Replaced with environment variables for the job

.github/workflows/images.yaml
kamilprz: Replaced with environment variables for the job

.github/workflows/images.yaml
```diff
@@ -91,36 +117,91 @@ jobs:
...
+              REPO_PATH=.
+          fi
+        env:
+          PLATFORM: windows
+          ARCH: amd64
+          YEAR: "2019"
+          IS_MERGE_GROUP: ${{ github.event_name == 'merge_group' }}
+
+
+  retina-image-win-2022:
```
carlotaarvela: This logic seems similar to the 2019. Could we this be refactored to reduce the logic duplication?

Makefile
```diff
@@ -262,19 +282,23 @@ retina-image: ## build the retina linux container image.
 	done
 
 retina-image-win: ## build the retina Windows container image.
```
carlotaarvela: When building a windows image on my local machine, do I know always have to run `make build-windows-binaries` before this? If so, it should be written somewhere or we should include the build-windows-binaries task here

controller/Dockerfile.windows-retina-oss-build
```diff
@@ -19,8 +19,11 @@ ENV GOOS=${GOOS}
 ENV OS_VERSION=${OS_VERSION}
 ENV BINARIES_PATH=${BINARIES_PATH}
 ENV REPO_PATH=${REPO_PATH}
-COPY ${REPO_PATH}/retina/windows/kubeconfigtemplate.yaml kubeconfigtemplate.yaml
-COPY ${REPO_PATH}/retina/windows/setkubeconfigpath.ps1 setkubeconfigpath.ps1
+# CVE-2013-3900 Mitigation
+RUN reg add "HKEY_LOCAL_MACHINE\Software\Microsoft\Cryptography\Wintrust\Config" /v "EnableCertPaddingCheck" /t REG_DWORD /d "1" /f
+RUN reg add "HKEY_LOCAL_MACHINE\Software\Wow6432Node\Microsoft\Cryptography\Wintrust\Config" /v "EnableCertPaddingCheck" /t REG_DWORD /d "1" /f
+COPY ${REPO_PATH}/windows/kubeconfigtemplate.yaml kubeconfigtemplate.yaml
+COPY ${REPO_PATH}/windows/setkubeconfigpath.ps1 setkubeconfigpath.ps1
```
carlotaarvela, line 26: Have you confirmed that the captureworkload.exe and controller.exe binaries are still signed?

.github/workflows/images.yaml
alexcastilio: cant `matrix` be brought back in?

kamilprz: yes the logic is the same - the main reason for the separation was that 2019 needs to run on 2019 host and likewise 2022 on 2022

let me see if i can parameterize this to make it cleaner

Makefile
kamilprz: good point, you're right

.github/workflows/images.yaml
kamilprz: done - see below for an example of execution (hope you can open it).  

https://github.com/kamilprz/retina/actions/runs/21366138281

commit - https://github.com/microsoft/retina/pull/1997/commits/ab09c073126bf8ebba553c786dd9ff6376af9855

Makefile
kamilprz: done - https://github.com/microsoft/retina/pull/1997/commits/613cc95500e417dd46ba5ca19190c2e89af9db36

controller/Dockerfile.windows-retina-oss-build
kamilprz: Yes :)

PR #2096, opened 2026-03-05
fix(ci): build kapinger arm64 natively instead of cross-compiling

nddq:
# Description

Currently the kapinger workflow cross-compiles all architectures (linux/amd64, linux/arm64, windows/amd64) from a single `--platform=linux/amd64` builder. The arm64 binary is built with `GOOS=linux GOARCH=arm64 go build`, which doesn't work reliably for CGO-dependent code and prevents using platform-specific optimizations.

This PR restructures the workflow and Dockerfile for native builds:

- **Dockerfile**: Split into a native Linux builder (runs on target arch) and a cross-compiling Windows builder (`--platform=$BUILDPLATFORM` with `GOOS=windows`). Fixed missing `/lib64/` copy that caused "no such file or directory" at runtime.
- **Workflow**: Use a matrix strategy with native runners (`ubuntu-latest` for amd64, `ubuntu-24.04-arm` for arm64). Split into separate `build-linux`, `build-windows`, `manifest`, and `build-toolbox` jobs.
- **Smoke test**: Each Linux build loads the image and verifies the binary starts correctly via `docker run`.
- **Manifest**: After all builds pass, create a multi-arch manifest list (`docker buildx imagetools create`) combining linux-amd64, linux-arm64, and windows.
- **Toolbox**: Add GHCR login and enable push on merge to main (previously build-only).
- **CI hardening**: Add path filters, PR/merge_group triggers, `permissions`, GHCR login, and pin all action SHAs.

## Related Issue

N/A

## Checklist

## Additional Notes

- Images are pushed only on merge to main, not on PRs.

### DIFF
--- a/.github/workflows/kapinger.yaml
+++ b/.github/workflows/kapinger.yaml
@@ -6,7 +6,68 @@ on:
       - main
 
 jobs:
-  build:
+  build-linux:
+    name: Build Linux Kapinger Image (${{ matrix.arch }})
+    runs-on: ${{ matrix.runner }}
+
+    strategy:
+      matrix:
+        arch: ["amd64", "arm64"]
+        include:
+          - arch: amd64
+            runner: ubuntu-latest
+          - arch: arm64
+            runner: ubuntu-24.04-arm
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Set outputs
+        id: vars
+        run: echo "sha_short=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
+
+      - name: Log in to GHCR
+        if: github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v'))
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
+
+      - name: Build Linux Kapinger Image
+        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6.19.2
+        with:
+          context: hack/tools/kapinger
+          file: hack/tools/kapinger/Dockerfile
+          platforms: linux/${{ matrix.arch }}
+          target: linux
+          load: true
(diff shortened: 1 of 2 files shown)


.github/workflows/kapinger.yaml
```diff
@@ -16,50 +40,68 @@ jobs:
...
-        uses: docker/build-push-action@v6
+        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6.19.2
         with:
           context: hack/tools
           file: hack/tools/toolbox/Dockerfile
           platforms: linux/amd64
-          push: false
+          push: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
           provenance: false
           tags: ghcr.io/${{ github.repository }}/toolbox:${{ steps.vars.outputs.sha_short }}-linux
```
alexcastilio, line 159: missing target for linux

alexcastilio (dismissed):

.github/workflows/kapinger.yaml
```diff
@@ -16,50 +43,150 @@ jobs:
...
-        run: echo ${{ steps.vars.outputs.sha_short }}
 
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v4
+      - name: Log in to GHCR
+        if: github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v'))
+        uses: docker/login-action@74a5d142397b4f367a81961eba4e8cd7edddf772 # v3.4.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
```
alexcastilio: will `github.actor` with github token work? E.g. I create a PR, how can it login with my username and `secrets.GITHUB_TOKEN`?

.github/workflows/kapinger.yaml
```diff
@@ -4,10 +4,37 @@ on:
...
+    tags:
+      - "v*"
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "hack/tools/kapinger/**"
+      - ".github/workflows/kapinger.yaml"
+  merge_group:
+    types: [checks_requested]
```
alexcastilio: I think we can change the triggers in a different PR

.github/workflows/kapinger.yaml
nddq: the username is not actually required here as docker login only concerns about the token, i'll change this to match with what we are doing in `release-images.yaml`

.github/workflows/kapinger.yaml
nddq: sure, let me split this out

PR #2322, opened 2026-05-12
fix(capture-cve): Retina capture tcpdump sanitization

### LINKED ISSUE #2298: vuln(retina-capture): cmdline Injection - Validate tcpdump args created via CR
```
}
```

S-03:

```
    "cmd", "/C",
)
if len(filter) != 0 {
}
```

PoC (CRD):
```
kind: Capture
spec:
  captureConfiguration:
    captureTarget:
```

**Fix:**

mereta:
# Security Enhancement - Command Injection Fix & API Modernization
### Original Issue
The `tcpdumpFilter` field in Capture CRD allowed command injection through malicious filter values:
- **Attack Vector**: Attackers with `captures.retina.sh` create permission could escalate to node-root/SYSTEM access
- **Example Exploit**: `tcpdumpFilter: "-z /evil.sh"` would execute arbitrary scripts with root privileges
- **Platforms Affected**: Both Linux (tcpdump) and Windows (netsh) implementations

### Attack Surface
1. **Linux tcpdump**: Flag injection via `-z`, `-w`, `-r`, `-G` and other dangerous flags
2. **Windows netsh**: Shell metacharacter injection via `&`, `|`, `;`, `$()`, etc.
3. **Filter bypass**: Whitespace-only or empty filters to bypass validation

---

## Overview

Security fix with API modernization - deprecated the vulnerable `tcpdumpFilter` field in favor of type-safe structured options under `captureOption`.

**Note:** Despite the field name, Windows nodes **do not use tcpdump**. The filter fields are interpreted as:
- **Unix/Linux**: BPF filter syntax for tcpdump
- **Windows**: netsh filter syntax for netsh trace

**Key Changes:**
1. **New Structured API**: Introduced `captureOption.pcapFilter` field with enum-based flags for type-safe capture configuration
2. **NEW: 3 Enum Fields** for tcpdump display options (Linux/Unix only):
   - `verbosity`: `""` (normal), `"verbose"` (tcpdump -v), `"extra"` (tcpdump -vv), `"max"` (tcpdump -vvv)
   - `printDataFormat`: `""` (none), `"hex"` (tcpdump -x), `"hex-with-link"` (tcpdump -xx), `"ascii"` (tcpdump -A), `"ascii-with-link"` (tcpdump -AA)
   - `timestampFormat`: `""` (default), `"none"` (tcpdump -t), `"unformatted"` (tcpdump -tt), `"delta"` (tcpdump -ttt), `"date"` (tcpdump -tttt), `"delta-since-first"` (tcpdump -ttttt)
3. **CLI Enhancement**: Added enum-based flags to `kubectl retina capture create` (`--verbosity`, `--print-data`, `--timestamp-format`)
4. **Unix/Linux - BPF Syntax Validation**: Uses `tcpdump -d` to validate filter syntax before capture execution
5. **Windows - Regex Validation**: Blocks shell metacharacters in netsh trace filters (enum flags ignored on Windows)
6. **Backward Compatibility**: Deprecated `tcpdumpFilter` still works with validation, but `pcapFilter` takes precedence
7. **Defense in Depth**: Multiple independent validation layers block dangerous content and malformed filters
8. **Clear Error Messages**: Users receive specific feedback about validation failures

---

### New Structured API

**CRD Changes** (`crd/api/v1alpha1/capture_types.go`)
- **New Field**: `captureOption.pcapFilter` (string) - BPF/netsh filter syntax (replaces `tcpdumpFilter`)
- **New Enum Fields** (**Linux/Unix only**) - Type-safe options for tcpdump display formatting:
  
  **`verbosity` (string enum)** - Controls tcpdump verbosity level (mutually exclusive):
  - `""` → Normal output (no -v flags)
  - `"verbose"` → `-v` (verbose output)
  - `"extra"` → `-vv` (extra verbose output)
  - `"max"` → `-vvv` (maximum verbose output)
  - **kubebuilder validation**: `+kubebuilder:validation:Enum="";verbose;extra;max`
  
  **`printDataFormat` (string enum)** - Controls packet data printing format (mutually exclusive):
  - `""` → No packet data printing
  - `"hex"` → `-x` (hex packet data)
  - `"hex-with-link"` → `-xx` (hex with link-level headers)
  - `"ascii"` → `-A` (ASCII packet data)
  - `"ascii-with-link"` → `-AA` (ASCII with link-level headers)
  - **kubebuilder validation**: `+kubebuilder:validation:Enum="";hex;hex-with-link;ascii;ascii-with-link`
  
  **`timestampFormat` (string enum)** - Controls timestamp display format (mutually exclusive):
  - `""` → Default timestamp format
  - `"none"` → `-t` (no timestamps)
  - `"unformatted"` → `-tt` (Unix timestamp)
  - `"delta"` → `-ttt` (time since previous packet)
  - `"date"` → `-tttt` (date and time)
  - `"delta-since-first"` → `-ttttt` (time since first packet)
  - **kubebuilder validation**: `+kubebuilder:validation:Enum="";none;unformatted;delta;date;delta-since-first`

- **Remaining Boolean Flags** (non-mutually-exclusive options):
  - `noPromiscuous` → `-p` (don't capture in promiscuous mode)
  - `packetBuffered` → `-U` (packet-buffered output)
  - `immediateMode` → `--immediate-mode` (immediate packet delivery)
  - `noResolveDNS` → `-n` (don't resolve addresses)
  - `noResolvePort` → `-nn` (don't resolve addresses or ports)
  - `printLinkHeader` → `-e` (print link-level header)
  - `quietOutput` → `-q` (quick/quiet output)
  - `absoluteSeq` → `-S` (absolute TCP sequence numbers)
  - `dontVerifyChecksum` → `-K` (don't verify checksums)
  - **Note**: All flags are ignored on Windows (netsh trace has no equivalent formatting options)

- **Deprecated Field**: `tcpdumpFilter` (still works for backward compatibility)
- **Precedence**: `pcapFilter` takes precedence over `tcpdumpFilter` when both are set
- **MaxLength**: 1024 characters on filter fields prevents resource exhaustion

### Three Validation Layers

**Layer 1: Type Safety** (CRD Level)
- **Enum validation**: Kubebuilder annotations enforce valid enum values at API server level
- **Mutually exclusive**: Single enum field prevents conflicting options (e.g., can't set both "verbose" and "max")
- **String-based enums**: Direct mapping to tcpdump flags with clear semantics
- No user-provided flags can reach the command line

**Layer 2: Filter Syntax Validation** (Unix/Linux - `network_capture_unix.go`)
- Validates BPF filter syntax using `tcpdump -d` before execution
- Catches invalid ports, unknown protocols, malformed syntax
- Prevents injection attempts disguised as filter syntax

**Layer 3: Legacy Path Validation** (Backward Compatibility)
- Deprecated `tcpdumpFilter` still validated with warning logs
- Same BPF/netsh syntax validation applied

### Windows: Filter Validation Only

**Implementation** (`network_capture_win.go`)
- **Uses netsh trace** (tcpdump is not available on Windows)
- **Enum flags not supported** - netsh trace does not have formatting options like tcpdump
  - Enum and boolean flags are silently ignored on Windows (no error, but no effect)
  - Output formatting happens post-capture using tools like Network Monitor or Wireshark
- **Filter validation only**: `pcapFilter` interpreted as netsh filter syntax (key=value format, e.g., `IPv4.Address=10.0.0.1`)
- Regex pattern validates filter contains only safe characters: `^[A-Za-z0-9.=():, ]+$`
- **Blocks shell metacharacters**: `&`, `|`, `^`, `<`, `>`, `%`, `"`, `'`, `;`, `$`, `` ` ``, `\`, newlines
- Filter tokens split on spaces and passed as separate arguments to netsh

**Security Features:**
- Uses `exec.Command` with explicit Args (no shell parsing)
- Validates filter content before adding to command
- Clear error message: "filter contains invalid characters; only alphanumeric, dots, colons, parentheses, commas, equals, and spaces are allowed"

---

## Test Results

**7 test cases covering new enum-based API, backward compatibility, and CLI:**

| Test ID | Test Case | Configuration | Expected | Result |
|---------|-----------|---------------|----------|--------|
| **TC-01** | Enum - Verbose + Hex + Date | `verbosity: "verbose"` + `printDataFormat: "hex"` + `timestampFormat: "date"` + `pcapFilter: "tcp port 443"` | `-v -x -tttt tcp port 443` | ✅ PASS |
| **TC-02** | Enum - Max Verbosity + ASCII + Delta | `verbosity: "max"` + `printDataFormat: "ascii"` + `timestampFormat: "delta"` + other flags | `-vvv -A -ttt tcp port 443` | ✅ PASS |
| **TC-03** | Enum - Extra Verbose + Hex with Link + No Timestamps | `verbosity: "extra"` + `printDataFormat: "hex-with-link"` + `timestampFormat: "none"` | `-vv -xx -t tcp` | ✅ PASS |
| **TC-04** | Enum - Default Values (Empty Strings) | `verbosity: ""` + `printDataFormat: ""` + `timestampFormat: ""` + boolean flags | No enum flags, only boolean flags applied | ✅ PASS |
| **TC-05** | CLI - Enum Flags | `kubectl retina capture create --verbosity verbose --print-data hex --timestamp-format date` | CRD created with enum values set correctly | ✅ PASS |
| **TC-06** | Deprecated Field Still Works | `tcpdumpFilter: "icmp"` (no enums) | Warning logged + `icmp` filter applied | ✅ PASS |
| **TC-07** | Precedence Test | `tcpdumpFilter: "tcp port 8080"` + `pcapFilter: "tcp port 80"` | `pcapFilter` used: `tcp port 80` | ✅ PASS |
| **TC-08** | '-' rejection in pcapFilter & tcpdumpFilter | `tcpdumpFilter: "-n tcp port 8080"` + `pcapFilter: "-n tcp port 80"` | Validation fail | ✅ PASS |

**Overall Results:**
- **New Enum-Based API**: 4/4 tests passed (TC-01, TC-02, TC-03, TC-04)
- **CLI with Enums**: 1/1 test passed (TC-05)
- **Backward Compatibility**: 1/1 test passed (TC-06)
- **Precedence Logic**: 1/1 test passed (TC-07)
- **Overall**: 7/7 tests passed (100%)

CLI command:
```
/usr/local/bin/kubectl-retina capture create \
  --name test-enum-cli-v3 \
  --duration 3m \
  --verbosity verbose \
  --print-data hex \
  --timestamp-format date \
  --no-resolve-dns \
  --no-resolve-port \
  --host-path "/tmp/retina/cli-test" \
  --node-selectors "kubernetes.io/os=linux"
```

One of CRD's with filters:
image

**Screenshots:**
TC-05: CLI with Enum Flags:
image

TC-06: Deprecated Filter:
image

TC-07: Filter Precedence:
image

TC-01:  Filter Flags:
image

TC-08: Disallow '-' flags
tcpdumpfilter:
image

pcapfilter:
image

CRD:
image

**Test Files for Validation:**
- `test-enum-capture.yaml` - Comprehensive example with all three enum fields set
- `test-enum-examples.yaml` - Five different examples showing various enum combinations
- `TESTING-ENUMS.md` - Complete testing guide with enum→tcpdump flag mapping table and validation tests

---

### Issue
Old code concatenated user and system filters without `and` operator, creating invalid BPF syntax.
Only true for when `tcpdumpfilter` and `podselector` or `include/exclude` combination was present.

**Old (INVALID):**
```bash
tcpdump -i any tcp port 443 ((host 192.168.0.1) or (host 10.244.0.5))
```

**New (VALID):**
```bash
tcpdump -i any "tcp port 443 and ((host 192.168.0.1) or (host 10.244.0.5))"
```

### Impact
- User filters now work with system filters
- Proper BPF combination using `and` operator
- Both new `PcapFilter` and deprecated `TcpdumpFilter` functional
---

### Why Enum-Based Fields for Display Options?

Enums provide a cleaner API for mutually exclusive tcpdump options:
- **Mutually Exclusive by Design**: Single field prevents conflicting options (e.g., can't set both "verbose" and "max")
- **API Validation**: Kubebuilder enum validation enforces valid values at Kubernetes API server level  
- **Clear Intent**: `verbosity: "max"` is self-documenting
- **Type Safety**: Enum values constrained to predefined safe strings, preventing command injection

---

## Impact

**No Breaking Changes** - New enum fields added alongside existing options

**What's New:**
- 3 enum fields for tcpdump display options: `verbosity`, `printDataFormat`, `timestampFormat`
- CLI flags: `--verbosity`, `--print-data`, `--timestamp-format`
- Kubebuilder enum validation at API level
- Prevents setting conflicting options (e.g., both verbose and max verbosity)

**Security:**
- Type-safe enum values prevent command injection
- Filter syntax validation using `tcpdump -d` (Linux) or regex (Windows)
- No user-provided flags reach command line


alexcastilio:
I think the field name `tcpdumpFilter` states that it's meant for filtering not for providing other flags. If providing other flags is desirable, then that should be handled separately.

mereta:
> I think the field name `tcpdumpFilter` states that it's meant for filtering not for providing other flags. If providing other flags is desirable, then that should be handled separately.

It would make sense since the -i 'interface' flag is extracted as an option.

SRodi (commented):
Why not remove the `tcpdumpFilter` entirely, and add fields only for what we want to allow, as opposed to maintain a whitelist of tcpdump flags?

cli/cmd/capture/capture.go
```diff
@@ -38,7 +38,30 @@ type Opts struct {
...
+	tcpdumpFilter        string
+	pcapFilter           string
+	noPromiscuous        bool
+	packetBuffered       bool
+	immediateMode        bool
+	noResolveDNS         bool
+	noResolvePort        bool
+	verbose              bool
+	extraVerbose         bool
+	maxVerbose           bool
```
SRodi: Can we group fields into enums, especially if fields are mutually exclusive? verbosity is an example: 

```go
type VerbosityLevel int

const (
    VerbosityNormal VerbosityLevel = iota
    VerbosityVerbose
    VerbosityExtra
    VerbosityMax
)
```

mereta: That's a good idea! We can do this with a few others also! 👌🏼

pkg/capture/provider/network_capture_unix.go
```diff
@@ -122,21 +144,76 @@ func (ncp *NetworkCaptureProvider) CaptureNetworkPacket(ctx context.Context, fil
...
+
+	if trimmedUserInput == "" && (pcapFilter != "" || tcpdumpRawFilter != "") {
+		return errTcpdumpFilterEmptyOrWhitespace
+	}
+
+	// Combine BPF filters using 'and' operator if both are present
+	switch {
+	case trimmedUserInput != "" && filter != "":
+		// User filter + system filter: "tcp port 80 and (host 10.0.0.1)"
+		combinedFilter = fmt.Sprintf("%s and %s", trimmedUserInput, filter)
```
alexcastilio:
```suggestion
		combinedFilter = fmt.Sprintf("(%s) and %s", trimmedUserInput, filter)
```
if user inputs `"tcp port 80 or udp port 53"`  -> `combindedFilter =  "tcp port 80 or udp port 53 and host XX"`. If OR takes precedence over AND (I'm not sure about this), this could be logically same as `tcp port 80 or (udp port 53 and host XX)`, which is undesirable. Adding the parenthesis will give more safety on what is AND'ed.

crd/api/v1alpha1/capture_types.go
```diff
@@ -67,10 +67,128 @@ type CaptureOption struct {
...
+
+	// PrintDataHex prints packet data in hex and ASCII (equivalent to tcpdump -X flag).
+	// Shows packet payload in both hexadecimal and ASCII format.
+	// +optional
+	PrintDataHex *bool `json:"printDataHex,omitempty"`
+
+	// PrintDataHexLink prints packet data including link-level header in hex and ASCII (equivalent to tcpdump -XX flag).
+	// Like PrintDataHex but also includes the link-layer (Ethernet) header.
+	// +optional
+	PrintDataHexLink *bool `json:"printDataHexLink,omitempty"`
```
alexcastilio: In someplaces, it's -X and -XX and in others it's -x and -xx.

pkg/capture/crd_to_job.go
```diff
@@ -1013,10 +1041,26 @@ func (translator *CaptureToPodTranslator) ObtainCaptureJobPodEnv(capture retinav
 		jobPodEnv[captureConstants.PacketSizeEnvKey] = strconv.Itoa(*capture.Spec.CaptureConfiguration.CaptureOption.PacketSize)
 	}
 
+	if capture.Spec.CaptureConfiguration.CaptureOption.PcapFilter != nil {
+		jobPodEnv[captureConstants.PcapFilterEnvKey] = *capture.Spec.CaptureConfiguration.CaptureOption.PcapFilter
+	}
```
alexcastilio, line 1034: Is there any validation happening to the PcapFilter when running from the CLI? What happens if a user tries to inject some malicious argument here?

pkg/capture/provider/network_capture_test.go
```diff
@@ -123,42 +143,555 @@ func TestTcpdumpSpecificInterfaces(t *testing.T) {
...
+			err := ncp.CaptureNetworkPacket(context.Background(), "", 1, 0)
+
+			if tt.shouldError {
+				if err == nil {
+					t.Errorf("Expected error containing '%s', but got no error", tt.errorMsg)
+				} else if !strings.Contains(err.Error(), tt.errorMsg) {
+					t.Errorf("Expected error containing '%s', but got: %v", tt.errorMsg, err)
+				}
+			} else {
+				if err != nil && !strings.Contains(err.Error(), "tcpdump") {
```
alexcastilio: if host does not have tcpdump installed, the error message might contain "tcpdump" in it (e.g. "tcpdump command not found"), and it will be ignored even though test was not run.

pkg/capture/provider/network_capture_unix.go
```diff
@@ -122,21 +144,76 @@ func (ncp *NetworkCaptureProvider) CaptureNetworkPacket(ctx context.Context, fil
...
+	// PcapFilter takes precedence over deprecated TcpdumpFilter if both are provided
+	var combinedFilter string
 
-	if filter != "" {
-		captureStartCmd.Args = append(
-			captureStartCmd.Args,
-			filter,
-		)
+	// Check for PcapFilter first (new, preferred)
+	pcapFilter := os.Getenv(captureConstants.PcapFilterEnvKey)
```
alexcastilio: why are you reading this env var here instead of using function argument `filter` as before? 

check:
https://github.com/microsoft/retina/blob/1f9ebdbeba756defa6407a46659bf696df7c56fb/pkg/capture/capture_manager.go#L57-L69
https://github.com/microsoft/retina/blob/1f9ebdbeba756defa6407a46659bf696df7c56fb/pkg/capture/capture_manager.go#L116-L122

alexcastilio (changes requested):

crd/api/v1alpha1/capture_types.go
mereta: Corrected the doc.

pkg/capture/provider/network_capture_unix.go
mereta: I will wrap both params in brackets.

pkg/capture/crd_to_job.go
mereta: Here we just set up for the job and validation happens in `network_capture_unix.go`.
I will add a check for early validation in `create.go` if filters contain any illegal prefixes job wont even start.

pkg/capture/provider/network_capture_test.go
mereta: Added a prereq `requireTcpdump`

pkg/capture/provider/network_capture_unix.go
mereta: The filters you reference are preprocessed filters from include/exclude filters they are not the Tcpdumprawfilter.
Its a bit confusing.

but netsh get set here `getNetshFilterWithPodIPAddress`
and tcpdumpfilter (include/exclude) here - `updateTcpdumpFilterWithPodIPAddress` 

So the logic we have reading from env var is corretc.

mereta: I renamed the `filter` to `includeExcludeFilter`

deploy/standard/manifests/controller/helm/retina/crds/retina.sh_captures.yaml
```diff
@@ -263,7 +357,12 @@ spec:
                       - Network statistics information
                     type: boolean
                   tcpdumpFilter:
-                    description: TcpdumpFilter is a raw tcpdump filter string.
+                    description: |-
+                      TcpdumpFilter accepts BPF filter expressions only (no flags).
+
+                      Deprecated: Use captureOption.pcapFilter for BPF expressions and captureOption boolean flags for display options.
+                    maxLength: 1024
+                    pattern: ^[^-]*$
```
carlotaarvela, line 366: We still support some tcpdumpfilter params so stating that is deprecated is a bit confusing. Maybe we should be more descriptive here about what is valid and invalid? Same for the docs

pkg/capture/provider/network_capture_unix.go
```diff
@@ -112,7 +205,7 @@ func (ncp *NetworkCaptureProvider) Setup(filename file.CaptureFilename) (string,
 	return ncp.TmpCaptureDir, nil
 }
 
-func (ncp *NetworkCaptureProvider) CaptureNetworkPacket(ctx context.Context, filter string, duration, maxSizeMB int) error {
+func (ncp *NetworkCaptureProvider) CaptureNetworkPacket(ctx context.Context, includeExcludeFilter string, duration, maxSizeMB int) error {
```
carlotaarvela, line 208: Where is this function being called? I can't find where it was called, as the params pass need to now include the  `includeExcludeFilter` but I can't find that change

mereta: This is just a rename from `filter` to `includeExcludeFilter` to deine it more explicitly. Its still just a string param. No other change needed.

deploy/standard/manifests/controller/helm/retina/crds/retina.sh_captures.yaml
mereta: I didn't want to remove it completely due to Geneva Action changes, hence the deprecated comment.
I can make the comment more descriptive though.

mereta: Updated the descriptions.

carlotaarvela (approved):

alexcastilio (approved):

SRodi (approved):

PR #2328, opened 2026-05-13
fix(clang-injection): Validate Retina dir header

### LINKED ISSUE #2298: vuln(retina-capture): cmdline Injection - Validate tcpdump args created via CR

mereta:
# Security Fix: Prevent Clang Argument Injection via RETINA_VMLINUX_HEADER_DIR

## Summary
Fixes a **High severity** information disclosure vulnerability where the `RETINA_VMLINUX_HEADER_DIR` environment variable could be exploited to inject clang arguments, enabling arbitrary file reads through clang's `--config` flag.

This fix implements:
1. **Input Validation**: Block dangerous characters at the source
2. **Safe Default**: Always fall back to known-good value
3. **Security Logging**: Warn when suspicious input is detected
4. **No Breaking Changes**: Default behavior unchanged

## Vulnerability Details

**Attack Vector**: Argument injection during eBPF compilation  
**Impact**: Information disclosure - leak sensitive files (passwords, tokens, keys)  
**Prerequisites**: Ability to modify Retina agent DaemonSet configuration  

The `RETINA_VMLINUX_HEADER_DIR` environment variable is read without validation and directly incorporated into clang command-line arguments via the `-I` flag:

```go
// Before (vulnerable)
runtimeIncludeDir := "-I" + loader.VmlinuxHeaderDir()
// If RETINA_VMLINUX_HEADER_DIR="/tmp/fake --config /etc/shadow"
// becomes: -I/tmp/fake --config /etc/shadow
```

Clang interprets this as two separate arguments:
1. `-I/tmp/fake` (include directory)
2. `--config /etc/shadow` (read config from file)

File contents leak via clang error messages in pod logs.

**Example Attack**:
```yaml
# Malicious DaemonSet modification
env:
  - name: RETINA_VMLINUX_HEADER_DIR
    value: "/tmp/fake --config /etc/shadow"
```

**Files at Risk**:
- `/etc/shadow` - Password hashes
- `/etc/passwd` - User information
- `/var/run/secrets/kubernetes.io/serviceaccount/token` - K8s service account token
- `/root/.ssh/id_rsa` - SSH private keys
- Application secrets, certificates, configuration files

### Validation Rules

- Reject spaces (prevents multi-argument injection)
- Reject tabs/newlines (prevents hidden arguments)  
- Reject dashes (prevents flag injection like `--config`)
- Fall back to safe default `/tmp/retina/include`
- Log warning for security monitoring

### Unit Tests

Added 9 test cases in `pkg/loader/vmlinux_test.go`:
**All tests passing**: `go test -v ./pkg/loader/... -run TestVmlinux`

### Manual Validation
```
# Deploy with malicious env var
kubectl set env daemonset/retina-agent -n kube-system \
  RETINA_VMLINUX_HEADER_DIR="/tmp/headers --config /etc/shadow"
```

Pre Fix:
image

Post Fix:
image

### No Impact on Existing Behavior

- Environment variable is not set  in deployments 
- Default path `/tmp/retina/include` is always used
- Variable only configurable via Helm values in standard deployment:
  ```yaml
  daemonset:
    container:
      retina:
        env: []  # Empty by default
  ```
- Hubble deployment has no provision for custom env vars

### DIFF
--- a/pkg/loader/vmlinux.go
+++ b/pkg/loader/vmlinux.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
+	"unicode"
 
 	"github.com/microsoft/retina/pkg/log"
 	"go.uber.org/zap"
@@ -35,6 +36,28 @@ var (
 func VmlinuxHeaderDir() string {
 	dir := strings.TrimSpace(os.Getenv(VmlinuxHeaderDirEnv))
 	if dir == "" {
+		return DefaultVmlinuxHeaderDir
+	}
+
+	// Security: Prevent argument injection via environment variable
+	// Block any path containing whitespace that could be used to inject
+	// additional clang arguments (e.g., "--config /etc/passwd")
+	// Also block paths starting with dash (would be interpreted as a flag)
+	for _, r := range dir {
+		if unicode.IsSpace(r) {
+			log.Logger().Named("vmlinux").Warn(
+				"RETINA_VMLINUX_HEADER_DIR contains whitespace, using default",
+				zap.String("rejected_value", dir),
+			)
+			return DefaultVmlinuxHeaderDir
+		}
+	}
+
+	if strings.HasPrefix(dir, "-") {
+		log.Logger().Named("vmlinux").Warn(
+			"RETINA_VMLINUX_HEADER_DIR starts with dash (potential flag injection), using default",
+			zap.String("rejected_value", dir),
+		)
 		return DefaultVmlinuxHeaderDir
 	}
 
--- a/pkg/loader/vmlinux_test.go
+++ b/pkg/loader/vmlinux_test.go
@@ -36,3 +36,120 @@ func TestGenerateVmlinuxH(t *testing.T) {
 	require.NoError(t, err)
 	require.Contains(t, string(content), "typedef")
 }
+
+func TestVmlinuxHeaderDir_Security(t *testing.T) {
+	// Initialize logger for test
+	_, err := log.SetupZapLogger(log.GetDefaultLogOpts())
+	require.NoError(t, err)
+
+	tests := []struct {
+		name     string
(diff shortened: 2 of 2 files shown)


SRodi (dismissed):
LGTM

pkg/loader/vmlinux_test.go
```diff
@@ -36,3 +36,108 @@ func TestGenerateVmlinuxH(t *testing.T) {
...
+		},
+		{
+			name:     "block space injection",
+			envValue: "/tmp/fake --config /etc/passwd",
+			expected: DefaultVmlinuxHeaderDir,
+			desc:     "Should reject path with spaces (argument injection attempt)",
+		},
+		{
+			name:     "block dash injection",
+			envValue: "/tmp/fake-config",
```
alexcastilio: isn't this a valid path? Isn't it overly aggressive and potentially breaking legitimate paths?

pkg/loader/vmlinux.go
```diff
@@ -38,6 +38,17 @@ func VmlinuxHeaderDir() string {
 		return DefaultVmlinuxHeaderDir
 	}
 
+	// Security: Prevent argument injection via environment variable
+	// Block any path containing whitespace or dash characters that could
+	// be interpreted as additional clang arguments (e.g., "--config /etc/passwd")
+	if strings.ContainsAny(dir, " \t\n\r-") {
```
alexcastilio: to check all different possible space characters `unicode.IsSpace` can be used (https://pkg.go.dev/unicode#IsSpace)

alexcastilio (approved):

PR #2335, opened 2026-05-13
fix(capture): make HostPath a subpath under a fixed base dir

### LINKED ISSUE #2298: vuln(retina-capture): cmdline Injection - Validate tcpdump args created via CR

SRodi:
## Problem
`Capture.spec.outputConfiguration.hostPath` was passed unchecked into `corev1.HostPathVolumeSource.Path`. Any user able to create a `Capture` CR could mount an arbitrary host directory RW into a privileged pod and write artifacts anywhere on the node.

## Fix
`HostPath` is now a **relative subpath name**. The operator mounts `${baseDir}/${hostPath}`, where `baseDir` is set by the cluster operator via Helm value `capture.hostPathBaseDir` (default `/var/log/retina/captures`). CR authors cannot influence `baseDir`.

Validation rejects (loudly, no silent rewrite):
- empty / `.`
- absolute paths (POSIX `/`, Windows `C:\`, leading `\`)
- any `..` segment, before or after `filepath.Clean`
- defense-in-depth: cleaned join must stay under `baseDir`

The resolved path is used for `HostPathVolumeSource.Path`, the container `VolumeMount.MountPath`, and the `CAPTURE_OUTPUT_LOCATION_HOST_PATH` env var.

## Key changes
- New `pkg/capture/hostpath_validation.go` + tests (`validateHostPath`, sentinel errors, `DefaultHostPathBaseDir`).
- `pkg/capture/crd_to_job.go`: `validateCapture` rejects early; `initJobTemplate` and `obtainCaptureOutputEnv` use the resolved path.
- `pkg/config/capture.go` + `operator/config/config.go`: new `CaptureHostPathBaseDir` (defaulted, cleaned, must be absolute).
- Helm: `capture.hostPathBaseDir` in `values.yaml` and `operator-configmap.yaml`.
- CLI: `--host-path` default is now `retina`; new `--host-path-base-dir` flag; auto-allow hack removed.
- CRD godoc + manifest regenerated; samples and docs updated.

## Breaking change
Capture CRs with an absolute `hostPath` (e.g. `/tmp/retina`) are rejected. Migration: use a bare name (`retina`); the mount becomes `/var/log/retina/captures/retina`. Operators needing a different root set `capture.hostPathBaseDir`.

## Test
- `go test ./pkg/capture/ ./pkg/config/ ./operator/config/ ./cli/...` — pass
- `go vet  ./pkg/capture/ ./pkg/config/ ./operator/config/ ./cli/...` — clean
- Unit cases: empty, `.`, POSIX/Windows absolute, pre-clean `..`, post-clean `foo/../bar`, escape, accepted nested subpaths.
- Manual kind run: absolute / `..` / `foo/../bar` rejected with no Job created and a single `CaptureError` condition; bare subpath mounts at `${baseDir}/${name}`; overriding `capture.hostPathBaseDir` relocates the mount without changing the CR.

## Related Issue

## Checklist

### Negative — absolute HostPath is rejected loudly

Screenshot 2026-05-18 164606

### Negative — parent segment is rejected

Screenshot 2026-05-18 164729

### Negative — post-clean traversal is rejected

Screenshot 2026-05-18 164804

### Positive — bare subpath is accepted

Screenshot 2026-05-18 165003

### Override the base directory

Screenshot 2026-05-18 165500

## Additional Notes

---


mereta (dismissed):
LGTM

cli/cmd/capture/create.go
```diff
@@ -437,12 +437,19 @@ func createCaptureF(ctx context.Context, kubeClient kubernetes.Interface) (*reti
 }
 
 func getCLICaptureConfig() config.CaptureConfig {
```
alexcastilio, line 449: Please add a unit test validating this behavior change and to be safe against future regressions

operator/config/config.go
```diff
@@ -53,5 +53,12 @@ func GetConfig(cfgFileName string) (*OperatorConfig, error) {
 		return nil, ErrorTelemetryIntervalTooSmall
 	}
 
+	// If unset, default the HostPath allowlist so that Capture CRs cannot mount arbitrary
+	// host directories into the privileged capture pod.
+	if len(cfg.CaptureHostPathAllowedPrefixes) == 0 {
+		log.Printf("captureHostPathAllowedPrefixes is not set, defaulting to [/var/log/retina/captures]")
+		cfg.CaptureHostPathAllowedPrefixes = []string{"/var/log/retina/captures"}
+	}
```
alexcastilio: Isn't it better to leave it empty? If you have a field to allow something, leave it empty, and it is set in the background to some value is not a good experience. It's also a way to not allow captures in a cluster.

alexcastilio (commented):
I sent some comments on the current approach. I'd like to suggest a different approach:

Instead of allowing the user to set any folder in the host, use the value provided by the user in `outputConfiguration.hostPath`, to create that as a subpath inside a default folder. E.g. defaultOutputDir = `/captures` and user sets `outputConfiguration.hostPath=my_capture123`, then it will be saved to `/captures/my_capture123`

Keep the validation against using `..` to go outside of that folder, and we don't need to create a breaking change which is to create a new list of allowedPrefixes.

operator/config/config.go
mereta: Actually we need to document this. 
- CRD - https://github.com/microsoft/retina/blob/2323dfbdd4605474dd69c6eb8e4568df4b4410dc/docs/05-Concepts/CRDs/Capture.md?plain=1#L34
- CLI - https://github.com/microsoft/retina/blob/2323dfbdd4605474dd69c6eb8e4568df4b4410dc/docs/04-Captures/02-cli.md?plain=1#L66

SRodi:
> I sent some comments on the current approach. I'd like to suggest a different approach:
> 
> Instead of allowing the user to set any folder in the host, use the value provided by the user in `outputConfiguration.hostPath`, to create that as a subpath inside a default folder. E.g. defaultOutputDir = `/captures` and user sets `outputConfiguration.hostPath=my_capture123`, then it will be saved to `/captures/my_capture123`
> 
> Keep the validation against using `..` to go outside of that folder, and we don't need to create a breaking change which is to create a new list of allowedPrefixes.

Thanks @alexcastillo Agreed. I implemented exactly that (user value is treated as a subpath under an operator-configured base dir, default /var/log/retina/captures); pushed in the last commit. The only remaining breaking change is that absolute hostPath values are now rejected loudly instead of being silently relocated, to avoid surprising on-disk moves on upgrade.

mereta (dismissed):
LGTM

alexcastilio:
@SRodi have you reviewed the download capture flow? Will they download from right location with your change? I noticed that there's no change in `cli/cmd/capture/{download,download_test}.go` files.

SRodi:
> @SRodi have you reviewed the download capture flow? Will they download from right location with your change? I noticed that there's no change in `cli/cmd/capture/{download,download_test}.go` files.

good catch! thanks @alexcastilio! We need to automate capture and download tests to avoid manual testing for each capture change.

Here is a manual test for now

image

cli/cmd/capture/create_test.go
```diff
@@ -340,28 +342,33 @@ func JobsCreatedCorrectly(t *testing.T, kubeClient *fake.Clientset, tc testcase)
...
+	opts.Name = &name
+
+	capture, err := createCaptureF(context.Background(), fake.NewClientset())
+	require.NoError(t, err)
+
+	require.Equal(t, []string{"nonexistent-pod"}, capture.Spec.CaptureConfiguration.CaptureTarget.PodNames)
+	require.Nil(t, capture.Spec.CaptureConfiguration.CaptureTarget.NodeSelector)
 }
 
 func TestCreateCaptureWithPodNames_CRDStructure(t *testing.T) {
```
alexcastilio, line 374: add a test with `--host-path=/tmp/foo` and `wantErr: true` to make sure the CLI will fail as expected.

cli/cmd/capture/create_test.go
```diff
@@ -340,28 +342,33 @@ func JobsCreatedCorrectly(t *testing.T, kubeClient *fake.Clientset, tc testcase)
...
+	opts.nodeSelectors = DefaultNodeSelectors
+	opts.podNames = "nonexistent-pod"
+	opts.Namespace = &namespace
+	opts.Name = &name
+
+	capture, err := createCaptureF(context.Background(), fake.NewClientset())
+	require.NoError(t, err)
+
+	require.Equal(t, []string{"nonexistent-pod"}, capture.Spec.CaptureConfiguration.CaptureTarget.PodNames)
+	require.Nil(t, capture.Spec.CaptureConfiguration.CaptureTarget.NodeSelector)
```
alexcastilio, line 371: Is this test change related with the goal of this PR?

pkg/capture/utils/annotations.go
```diff
@@ -9,14 +9,22 @@ import (
 	"github.com/microsoft/retina/pkg/capture/file"
 )
 
-func GetPodAnnotationsFromCapture(capture *retinav1alpha1.Capture) map[string]string {
+// GetPodAnnotationsFromCapture builds the capture pod annotations. resolvedHostPath,
+// when non-empty, is written as the CaptureHostPathAnnotationKey value (the on-node
+// path actually mounted into the capture pod). When empty, the raw value from the
+// Capture spec is used as a fallback (e.g. when called from contexts where the
+// host path has not been resolved yet).
+func GetPodAnnotationsFromCapture(capture *retinav1alpha1.Capture, resolvedHostPath string) map[string]string {
```
alexcastilio, line 17: Add some tests regarding the behavior of this function. Probably this would have flagged the issue with the download when the create logic changed.

cli/cmd/capture/create_test.go
SRodi: The original `TestCreateCaptureCommand_PodNamesWithNodeSelector_ShouldFail` was broken by this PR — it passed `--host-path=/tmp/capture` (now rejected as absolute) and then asserted on a `"not found"` error string, which masked the real failure. I replaced it with `TestCreateCaptureCommand_PodNamesClearsDefaultNodeSelector`, which verifies the actual intent (pod-names clears the default node selector) without depending on HostPath. Happy to revert to the minimal change (just `--host-path=capture`) if you'd prefer to keep the diff strictly scoped.

pkg/controllers/operator/capture/controller.go
```diff
@@ -226,7 +226,6 @@ func (cr *CaptureReconciler) createJobsFromCapture(ctx context.Context, capture
 
 	jobs, err := cr.captureToPodTranslator.TranslateCaptureToJobs(ctx, capture)
 	if err != nil {
-		cr.logger.Error("Failed to translate Capture to jobs", zap.Error(err), zap.String("Capture", captureRef.String()))
```
carlotaarvela, line 229: Why was this removed? Is this error being logged somewhere else?

pkg/capture/utils/annotations.go
```diff
@@ -9,14 +9,22 @@ import (
...
 		captureConstants.CaptureFilenameAnnotationKey: capture.Name,
 	}
 	if capture.Status.StartTime != nil {
 		annotations[captureConstants.CaptureTimestampAnnotationKey] = file.TimeToString(capture.Status.StartTime)
 	}
-	if capture.Spec.OutputConfiguration.HostPath != nil {
+	switch {
+	case resolvedHostPath != "":
+		annotations[captureConstants.CaptureHostPathAnnotationKey] = resolvedHostPath
+	case capture.Spec.OutputConfiguration.HostPath != nil:
```
mereta: `Cosmetic`: Only two cases here. could just be if else

pkg/capture/crd_to_job.go
```diff
@@ -122,6 +122,21 @@ func (translator *CaptureToPodTranslator) initJobTemplate(ctx context.Context, c
...
 	// uploaded, and this happens when the user want to stop a capture on demand by deleting the capture.
 	captureTerminationGracePeriodSeconds := int64(1800)
+
+	// Resolve the user-supplied HostPath subpath against the operator-configured base
+	// directory up front so the resolved on-node path can be stamped onto the pod
+	// annotation (used by `kubectl retina capture download`) and reused for the
+	// volume mount below.
+	var resolvedHostPath string
+	if capture.Spec.OutputConfiguration.HostPath != nil && *capture.Spec.OutputConfiguration.HostPath != "" {
+		resolved, err := validateHostPath(*capture.Spec.OutputConfiguration.HostPath, translator.config.CaptureHostPathBaseDir)
```
mereta: We do this same hostpath validation in the init, capture & output functions.
```
if capture.Spec.OutputConfiguration.HostPath != nil && *capture.Spec.OutputConfiguration.HostPath != "" {
		resolved, err := validateHostPath(*capture.Spec.OutputConfiguration.HostPath, translator.config.CaptureHostPathBaseDir)
```
Could probably simplify to validate once and use the result.

And they are all called sequentially in `TranslateCaptureToJobs`.

mereta (dismissed):
Approved with comments.

pkg/controllers/operator/capture/controller.go
SRodi: this was moved to the default branch of the switch statement https://github.com/microsoft/retina/pull/2335/changes/BASE..65af10e2f9ff06cea9320d9a78fdaa583473776a#diff-4684e057605308d5421aff31f549b9bbe173ab5771914902aed4031d4281fd63L240

mereta (dismissed):
LGTM! 👌🏼

carlotaarvela (dismissed):

mereta (approved):
Re-approve!

PR #2367, opened 2026-05-20
feat(capture): Add --cleanup-after-upload flag for automatic resource cleanup

carlotaarvela:
# Description

Adds a `--cleanup-after-upload` flag to `kubectl retina capture create` that enables automatic cleanup of capture jobs, secrets, and host-path files after successful upload to remote storage (blob, S3, or PVC).

**Key behaviors:**
- **TTL on jobs** (no-wait mode): When `--cleanup-after-upload` is combined with `--no-wait=true` and a remote destination, jobs get a `TTLSecondsAfterFinished` (5 min) so Kubernetes garbage-collects them automatically.
- **Secret ownerReferences** (always): Secrets created for blob/S3 uploads always get ownerReferences pointing to the capture job, ensuring they are cleaned up when the job is deleted—whether by TTL, manual delete, or the controller.
- **Controller auto-delete**: The operator's `CaptureReconciler` now deletes the Capture resource once all jobs succeed and `CleanUpAfterUpload` is true with remote storage configured (blob, S3, or PVC).
- **ActiveDeadlineSeconds**: Jobs now get a hard deadline (`duration + 30min`) to prevent indefinite hangs.
- **Wait timeout fix**: `waitUntilJobsComplete` deadline is now `duration + 5min` (floored at 5min), fixing premature timeouts for short captures.
- **NotFound tolerance**: `deleteSecret` and `capture delete` now tolerate `NotFound` errors caused by ownerRef garbage collection racing with explicit deletion.

## Related Issue

N/A — feature request for safer automated capture workflows.

## Checklist

### Unit Tests

All tests pass in the affected packages. The new logic is 100% covered by the new CLI and Operator controller tests added.

### Manual Testing

| # | Scenario | Command | Expected | Result |
|---|----------|---------|----------|--------|
| 1 | `--cleanup-after-upload` without remote | `./bin/kubectl-retina capture create --name test1 --node-names aks-nodepool1-... --host-path /tmp/captures --cleanup-after-upload --duration 5s` | Error: requires remote storage | **PASS** — error returned immediately |
| 2 | No-wait + blob + cleanup | `./bin/kubectl-retina capture create --name test2 --node-names aks-nodepool1-... --blob-upload "<SAS URL>" --cleanup-after-upload --no-wait=true --duration 10s` | CLI exits immediately, job has TTL, secret has ownerRef | **PASS** — verified via `kubectl get job -o yaml` (TTL=300), `kubectl get secret -o yaml` (ownerReferences present), job auto-deleted after ~5 min |
| 3 | Wait + blob + cleanup | `./bin/kubectl-retina capture create --name test3 --node-names aks-nodepool1-... --blob-upload "<SAS URL>" --cleanup-after-upload --no-wait=false --duration 10s` | CLI waits, uploads blob, then deletes jobs/secrets | **PASS** — CLI printed completion, `kubectl get jobs` showed no jobs remaining, blob appeared in storage account, secret deleted |
| 4 | Wait + blob WITHOUT cleanup | `./bin/kubectl-retina capture create --name test4 --node-names aks-nodepool1-... --blob-upload "<SAS URL>" --no-wait=false --duration 10s` | CLI waits, uploads, jobs/secrets remain | **PASS** — jobs and secrets preserved after CLI exit |
| 5 | Host-path only + no-wait (no cleanup flag) | `./bin/kubectl-retina capture create --name test5 --node-names aks-nodepool1-... --host-path /tmp/captures --no-wait=true --duration 5s` | No TTL set, jobs remain | **PASS** — `kubectl get job -o yaml` showed no TTLSecondsAfterFinished |

**Wait timeout fix verified:** Test 3 used `--duration 10s` and the CLI completed within ~75s (10s capture + upload time), confirming the new `duration + 5min` deadline works correctly for short captures (previously would timeout at 2×duration = 20s before upload finished).

**NotFound tolerance verified:** In test 2, running `kubectl retina capture delete --name test2` after TTL had already garbage-collected the job+secret returned success (no "secret not found" error).

## Additional Notes

- The `CleanUpAfterUpload` field is added to the CRD spec as optional with `+kubebuilder:default=false`. Existing Captures are unaffected.
- `JobActiveDeadlineBufferSeconds` (30 min) is generous to accommodate large captures uploading over slow links or Windows nodes with slow image pulls.
- The operator controller cleanup path only triggers when ALL jobs succeed — any failure preserves everything for debugging.
- The controller's remote storage check includes all three types: `BlobUpload`, `S3Upload`, and `PersistentVolumeClaim`.

### DIFF
--- a/cli/cmd/capture/create.go
+++ b/cli/cmd/capture/create.go
@@ -19,6 +19,7 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/kubectl/pkg/util/i18n"
@@ -63,6 +64,14 @@ const (
 	// the Job's activeDeadlineSeconds. This buffer accounts for output upload
 	// and cleanup time after the capture itself finishes.
 	JobActiveDeadlineBufferSeconds int64 = 1800 // 30 minutes
+
+	// WaitDeadlineBuffer is extra time added to the capture duration when
+	// computing the wait deadline, to account for upload and scheduling overhead.
+	WaitDeadlineBuffer time.Duration = 5 * time.Minute
+
+	// MinPollAttempts is the minimum number of polling iterations before the
+	// wait deadline expires. Used to compute a fallback poll period.
+	MinPollAttempts = 4
 )
 
 var errCleanupRequiresRemoteStorage = errors.New("--cleanup-after-upload requires remote storage (--blob-upload, --s3-bucket, or --pvc)")
@@ -666,7 +675,7 @@ func waitUntilJobsComplete(ctx context.Context, kubeClient kubernetes.Interface,
 	if opts.duration != 0 {
 		// Allow time for capture + upload + scheduling overhead (Windows
 		// nodes in particular need extra time for image pull and pod start).
-		deadline = opts.duration + 5*time.Minute
+		deadline = opts.duration + WaitDeadlineBuffer
 		if deadline < DefaultWaitTimeout {
 			deadline = DefaultWaitTimeout
 		}
@@ -679,7 +688,7 @@ func waitUntilJobsComplete(ctx context.Context, kubeClient kubernetes.Interface,
 	}
 	// Ensure poll period is less than the deadline so we get multiple checks.
 	if period >= deadline {
-		period = deadline / 4
+		period = deadline / MinPollAttempts
 	}
 	retinacmd.Logger.Info(fmt.Sprintf("Waiting timeout is set to %s", deadline))
 
@@ -756,7 +765,7 @@ func setSecretOwnerReferences(ctx context.Context, kubeClient kubernetes.Interfa
 	ownerRefs := make([]metav1.OwnerReference, 0, len(jobs))
 	for i := range jobs {
 		ownerRefs = append(ownerRefs, metav1.OwnerReference{
-			APIVersion: "batch/v1",
+			APIVersion: batchv1.SchemeGroupVersion.String(),
 			Kind:       "Job",
 			Name:       jobs[i].Name,
 			UID:        jobs[i].UID,
@@ -768,7 +777,16 @@ func setSecretOwnerReferences(ctx context.Context, kubeClient kubernetes.Interfa
 		if err != nil {
 			return fmt.Errorf("failed to get secret %s: %w", secretName, err)
 		}
-		secret.OwnerReferences = append(secret.OwnerReferences, ownerRefs...)
+		// Deduplicate owner references to avoid appending the same refs on retries.
+		existing := make(map[types.UID]struct{}, len(secret.OwnerReferences))
+		for _, ref := range secret.OwnerReferences {
+			existing[ref.UID] = struct{}{}
+		}
+		for _, ref := range ownerRefs {
+			if _, found := existing[ref.UID]; !found {
+				secret.OwnerReferences = append(secret.OwnerReferences, ref)
+			}
+		}
 		if _, err := kubeClient.CoreV1().Secrets(*opts.Namespace).Update(ctx, secret, metav1.UpdateOptions{}); err != nil {
 			return fmt.Errorf("failed to update secret %s with owner references: %w", secretName, err)
 		}
--- a/cli/cmd/capture/create.go
+++ b/cli/cmd/capture/create.go
@@ -36,9 +36,9 @@ import (
 )
 
 const (
-	DefaultCleanUpAfterUpload bool        = false
-	DefaultDebug    bool          = false
-	DefaultDuration time.Duration = 1 * time.Minute
+	DefaultCleanUpAfterUpload bool          = false
+	DefaultDebug              bool          = false
+	DefaultDuration           time.Duration = 1 * time.Minute
 	// DefaultHostPath is the default subpath (joined under DefaultHostPathBaseDir on
 	// the node) where capture artifacts are stored. It is a bare name, not an
 	// absolute path: absolute paths are rejected by the operator.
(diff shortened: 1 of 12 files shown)


cli/cmd/capture/create.go
```diff
@@ -471,14 +523,23 @@ func waitUntilJobsComplete(ctx context.Context, kubeClient kubernetes.Interface,
 	// TODO: let's make the timeout and period to wait for all job to finish configurable.
 	deadline := DefaultWaitTimeout
 	if opts.duration != 0 {
-		deadline = opts.duration * 2
+		// Allow time for capture + upload + scheduling overhead (Windows
+		// nodes in particular need extra time for image pull and pod start).
+		deadline = opts.duration + 5*time.Minute
```
SRodi: let's make the 5min delay either const or (even better) configurable

cli/cmd/capture/create.go
```diff
@@ -471,14 +523,23 @@ func waitUntilJobsComplete(ctx context.Context, kubeClient kubernetes.Interface,
...
 	}
 
 	period := DefaultWaitPeriod
 	// To print less noisy messages, we rely on duration to decide the wait period.
 	if period < opts.duration/10 {
 		period = opts.duration / 10
 	}
+	// Ensure poll period is less than the deadline so we get multiple checks.
+	if period >= deadline {
+		period = deadline / 4
```
SRodi: maybe we can avoid this magic number?

pkg/controllers/operator/capture/controller.go
```diff
@@ -215,7 +215,27 @@ func (cr *CaptureReconciler) updateCaptureStatusFromJobs(ctx context.Context, ca
...
 	}
 	capture.Status.CompletionTime = &lastCompleteTime
-	return cr.updateStatus(ctx, capture)
+	if result, err := cr.updateStatus(ctx, capture); err != nil {
+		return result, err
+	}
+
+	// If CleanUpAfterUpload is enabled and all jobs succeeded (uploaded to remote storage),
+	// automatically delete the Capture resource which triggers cleanup via the finalizer.
+	if capture.Spec.CleanUpAfterUpload && len(failedJobs) == 0 {
```
SRodi: I think we should flatten the nesting here for readability 

```go
if !capture.Spec.CleanUpAfterUpload || len(failedJobs) != 0 {
    return ctrl.Result{}, nil
}

hasRemoteStorage :=
    capture.Spec.OutputConfiguration.BlobUpload != nil ||
    capture.Spec.OutputConfiguration.S3Upload != nil ||
    capture.Spec.OutputConfiguration.PersistentVolumeClaim != nil

if !hasRemoteStorage {
    return ctrl.Result{}, nil
}
```

cli/cmd/capture/create.go
```diff
@@ -534,3 +595,42 @@ func deleteJobs(ctx context.Context, kubeClient kubernetes.Interface, jobs []bat
...
+	}
+
+	if len(secretNames) == 0 || len(jobs) == 0 {
+		return nil
+	}
+
+	ownerRefs := make([]metav1.OwnerReference, 0, len(jobs))
+	for i := range jobs {
+		ownerRefs = append(ownerRefs, metav1.OwnerReference{
+			APIVersion: "batch/v1",
```
SRodi: uses SchemeGroupVersion instead of hardcoded "batch/v1"

```suggestion
			APIVersion: "batchv1.SchemeGroupVersion.String()",
```

cli/cmd/capture/create.go
```diff
@@ -534,3 +595,42 @@ func deleteJobs(ctx context.Context, kubeClient kubernetes.Interface, jobs []bat
...
+			UID:        jobs[i].UID,
+		})
+	}
+
+	for _, secretName := range secretNames {
+		secret, err := kubeClient.CoreV1().Secrets(*opts.Namespace).Get(ctx, secretName, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get secret %s: %w", secretName, err)
+		}
+		secret.OwnerReferences = append(secret.OwnerReferences, ownerRefs...)
```
SRodi: at every reconcile we append the same ownerRefs again, this updates the object and triggers more reconciles/API calls/churn

With the current code we end up with something like this:

desired: [JobA, JobB]
actual after 10 reconciles: [JobA, JobB, JobA, JobB, JobA, JobB, ...]



```suggestion
		// Deduplicate owner references
        existing := make(map[types.UID]struct{}, len(secret.OwnerReferences))
        for _, ref := range secret.OwnerReferences {
            existing[ref.UID] = struct{}{}
        }

        for _, ref := range ownerRefs {
            if _, found := existing[ref.UID]; !found {
                secret.OwnerReferences = append(secret.OwnerReferences, ref)
            }
        }
```

cli/cmd/capture/create.go
carlotaarvela: Done, made it a const

cli/cmd/capture/create.go
carlotaarvela: Done, made in a const with a descriptive name

cli/cmd/capture/create.go
carlotaarvela: Done

cli/cmd/capture/create.go
carlotaarvela: Makes sense. Made the change and added unit tests to cover this scenario

pkg/controllers/operator/capture/controller.go
carlotaarvela: Done

crd/api/v1alpha1/capture_types.go
```diff
@@ -272,6 +272,14 @@ type CaptureSpec struct {
...
 	// +kubebuilder:validation:Required
 	OutputConfiguration OutputConfiguration `json:"outputConfiguration,omitempty"`
+	// CleanUpAfterUpload indicates whether the capture jobs and associated resources
+	// should be automatically cleaned up after a successful upload to remote storage
+	// (BlobUpload or S3Upload). When set to true, completed capture jobs, secrets,
+	// and the Capture resource itself will be deleted once all jobs have succeeded
+	// and uploads are confirmed.
+	// +optional
+	// +kubebuilder:default=false
+	CleanUpAfterUpload bool `json:"cleanUpAfterUpload,omitempty"`
```
mereta, line 282: I don't see the updated CRD definition in this PR, if you edit capture_types you should also run `make manifests` to build them.

cli/cmd/capture/create.go
```diff
@@ -114,11 +142,24 @@ func create(kubeClient kubernetes.Interface) error {
...
 	}
+
+	// Set owner references on secrets so they are garbage-collected when the
+	// jobs are removed (by TTL, explicit deletion, or manual kubectl delete).
+	if ownerRefErr := setSecretOwnerReferences(ctx, kubeClient, capture, jobsCreated); ownerRefErr != nil {
+		retinacmd.Logger.Error("Failed to set owner references on capture secrets", zap.Error(ownerRefErr))
+	}
+
 	if opts.nowait {
 		retinacmd.Logger.Info("Please manually delete all capture jobs")
```
mereta: Can `--no-wait` and `----cleanup-after-upload` are used together? Then is this message confusing?

pkg/controllers/operator/capture/cleanup_after_upload_test.go
mereta, line 1: All test cases are either `JobFailed` or `JobComplete`.
Should we have a test case that has both a failed & succeeded jobs?

crd/api/v1alpha1/capture_types.go
carlotaarvela: Done

cli/cmd/capture/create.go
carlotaarvela: I have updated these messages to be more clear. They can be use together, but it we are not cleaning up, we just set the secrets to be garbage collected once the user deletes the captures

pkg/controllers/operator/capture/cleanup_after_upload_test.go
carlotaarvela: Added test scenario

cli/cmd/capture/create.go
mereta: That's much clearer now! Thanks!

mereta (approved):
LGTM 👌🏼

SRodi (approved):
LGTM, thanks @carlotaarvela!

PR #1569, opened 2025-04-30
fix(controller): add warning for nodes with no addresses in reconciliation

selenehyun:
# Description

This PR improves the `NodeReconciler` logic in `controller.go` by adding a safeguard for nodes with no addresses.
Specifically, it ensures that if a node has an empty `Status.Addresses` field, a warning is logged, and the reconciliation process exits gracefully without further processing.

This prevents potential runtime errors when attempting to access an address that does not exist.

## Related Issue

resolve #1541

## Checklist

- [x] I have read the [contributing documentation](https://retina.sh/docs/Contributing/overview).
- [x] I signed and signed-off the commits (`git commit -S -s ...`). See [this documentation](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) on signing commits.
- [x] I have correctly attributed the author(s) of the code.
- [x] I have tested the changes locally.
- [x] I have followed the project's style guidelines.
- [ ] I have updated the documentation, if necessary.
- [ ] I have added tests, if applicable.

## Screenshots (if applicable) or Testing Completed

The following scenarios were tested:
1. **Node with no addresses**: Verified that a warning is logged, and the reconciliation exits without errors.
2. **Node with valid addresses**: Verified that the `RetinaNode` is created and updated in the cache correctly.
3. **Node being deleted**: Verified that the `RetinaNode` is removed from the cache as expected.

All tests passed successfully.

## Additional Notes

This change ensures that the `NodeReconciler` handles edge cases more robustly, improving the stability of the controller. The added safeguard prevents potential issues when interacting with nodes that lack address information.

---

Please refer to the [CONTRIBUTING.md](../CONTRIBUTING.md) file for more information on how to contribute to this project.

Issue #1541, opened 2025-04-17
Panic in Node controller when node.Status.Addresses is empty (index out of range at controller.go:78)


selenehyun:
**Describe the bug**  
Retina operator crashes with a panic due to an out-of-bounds slice access in `pkg/controllers/daemon/node/controller.go`. Specifically, the controller assumes that `node.Status.Addresses` always contains at least one entry and directly accesses `node.Status.Addresses[0].Address` without checking its length.

When the Node object has no addresses (common right after registration or in some environments), this results in a panic like:

```
[...]

PR #2389, opened 2026-05-27
fix: select node IP by address type

immanuwell:
# Description

Tiny follow-up for the node reconciler. If `status.addresses` has `Hostname` before `InternalIP`, Retina used the hostname as the cached node IP. That breaks lookups by the real node IP, kinda sneaky.

Now it picks `InternalIP` first, then `ExternalIP`, then any parseable IP. If there is no IP, it logs and skips the cache update.

## Related Issue

Related: #1541
Follow-up to #1569

## Repro

1. Reconcile a Node with addresses ordered like this:
   - `Hostname: test-node`
   - `InternalIP: 10.0.0.4`
2. Ask the cache for `10.0.0.4`.
3. Before this fix it returns nil, because the node was cached with `test-node`. not great

The new test covers this exact case.

## Checklist

- [x] I have read the contributing documentation.
- [x] I have followed the project style.
- [x] I have added tests.

## Testing Completed

`go test ./pkg/controllers/daemon/node ./pkg/controllers/cache ./pkg/controllers/daemon/service ./pkg/controllers/daemon/namespace ./pkg/controllers/daemon/metricsconfiguration`

Also checked `go test ./pkg/controllers/daemon/...`, but local envtest is missing `/usr/local/kubebuilder/bin/etcd`.

### DIFF
--- a/pkg/controllers/daemon/node/controller.go
+++ b/pkg/controllers/daemon/node/controller.go
@@ -80,13 +80,40 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{}, nil
 	}
 
-	retinaNodeCommon := retinaCommon.NewRetinaNode(node.Name, net.ParseIP(node.Status.Addresses[0].Address), node.Labels[corev1.LabelTopologyZone])
+	nodeIP := parseNodeIP(node.Status.Addresses)
+	if nodeIP == nil {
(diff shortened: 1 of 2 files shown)


immanuwell:
@microsoft-github-policy-service agree [company="Immanuel Tikhonov"]

immanuwell:

immanuwell:
@microsoft-github-policy-service agree company="Immanuel Tikhonov"

PR #2453, opened 2026-06-18
fix(chart): route hubble ui api via grpc_pass

abdullazin:
## Summary
- fix the Hubble UI `/api` path in nginx by switching from `proxy_pass` to `grpc_pass`
- preserve existing `baseUrl` rewrite logic
- keep the fix scoped to nginx template behavior (no extra backend service/routing objects)

## Why this change
- with Envoy in front of Hubble UI, the UI data stream call to `/api/ui.UI/GetEvents` was failing with HTTP `400`
- the UI surfaced this as: `Data stream has failed on the UI backend: Http response at 400 or 500 level, http status code: 400`
- the backend response included: `grpc-message: gRPC requires HTTP/2`
- this indicates the `/api` path needed to be handled as gRPC traffic at the nginx hop

## Path handling note
- a gRPC request carries its route in the `:path` pseudo-header
- `grpc_pass` does not rewrite request path based on upstream URI suffix
- therefore the conditional `grpc://127.0.0.1:8090` vs `grpc://127.0.0.1:8090/` is unnecessary
- this PR simplifies to a single `grpc_pass grpc://127.0.0.1:8090;`

## Files
- `deploy/hubble/manifests/controller/helm/retina/templates/hubble-ui/_nginx.tpl`

## Verification
- rendered chart templates locally after the nginx change
- confirmed only `_nginx.tpl` changed in this PR branch


abdullazin:
@microsoft-github-policy-service agree [company="GoTo"]

abdullazin:
@microsoft-github-policy-service agree company="GoTo"

abdullazin:
@nairashu or @alam-tahmid , can you please review my PR?

abdullazin:
Not necessary anymore: https://github.com/cilium/hubble-ui/pull/767

PR #2473, opened 2026-06-25
fix(dropreason): filter EAGAIN false drops in inet_csk_accept metrics

carlotaarvela:
# Description

Fix false/missing `TCP_ACCEPT_BASIC` drop metrics in the `dropreason` eBPF plugin. Two bugs were found:

1. **fexit path** (default mode): `inet_csk_accept_fexit` counted every NULL return from `inet_csk_accept()` as a packet drop without checking the error code. Non-blocking sockets returning EAGAIN (errno 11, "no pending connection") were incorrectly counted as drops, inflating `networkobservability_drop_count{reason="TCP_ACCEPT_BASIC"}` by orders of magnitude on busy nodes.

2. **kretprobe path** (pod-level mode): `inet_csk_accept_ret` did `int err = (int)*err_ptr` which truncates the stored **kernel pointer** to 32 bits instead of actually dereferencing it with `bpf_probe_read_kernel`. On x86_64, kernel stack addresses truncated to 32 bits are typically positive, so `err >= 0` was always true, so no drops were ever counted (not even real ones).

The fix:
- **fexit path (6.10+ kernels)**: Reads the error from `proto_accept_arg.err` via `bpf_probe_read_kernel()` and filters `err == 0` and `err == -EAGAIN`.
- **fexit path (pre-6.10 kernels)**: The BPF verifier on kernels like 5.15 rejects reading `ctx[2]` (encodes `int*` as scalar INT in BTF). The fexit `else` branch is now a no-op; instead, the Go loader attaches the kprobe/kretprobe pair for `inet_csk_accept` alongside the other fexit programs.
- **kretprobe path**: Uses `bpf_probe_read_kernel()` to properly dereference the saved kernel pointer, then filters `err >= 0 || err == -EAGAIN` (-11).

## Checklist

### Live cluster test environment
- AKS cluster
- Nodes: 3 Ubuntu 22.04 (kernel 5.15.0-1114-azure, amd64) + 1 AzureLinux 3.0 (kernel 6.6.139, amd64)
- Fixed image: `acnpublic.azurecr.io/carvela-dropreason-test:agent` / `:init`

### Test 1: Fix verification — EAGAIN no longer causes false drops
Deployed fixed image, ran non-blocking TCP accept server (tight EAGAIN loop) on Mariner node for 60 seconds:
```
Baseline TCP_ACCEPT_BASIC:  0
After 60s with EAGAIN load: 0    ← FIX CONFIRMED
```
For comparison, v1.2.0  « produced ~50,000,000 false drops in 60 seconds on the same node.

### Test 2: Programs attach correctly on all node types
Verified via pod logs on both Ubuntu 5.15 and AzureLinux 6.6:
```
Attached program: nf_hook_slow_fexit
Attached program: tcp_v4_connect_fexit
Attached kprobe: inet_csk_accept
Attached kretprobe: inet_csk_accept
```
All 4 pods (Ubuntu + Mariner) running with 0 restarts.

### Test 3: Bug reproduction (before fix, v1.2.0)
Same cluster, same non-blocking server, with the buggy v1.2.0 image:
```
Baseline: 20,076,242
After 30s: 45,743,820   (+25.7M false drops in 30s, ~856K/sec)
```

### Test 4: BPF verifier compatibility
The first iteration of the fix attempted to read `ctx[2]` in the fexit `else` branch on pre-6.10 kernels. This was rejected by the Ubuntu 5.15 BPF verifier:
```
func 'inet_csk_accept' arg2 type INT is not a struct: invalid bpf_context access off=16 size=8
```
The final fix uses the kprobe/kretprobe fallback for `inet_csk_accept` alongside fexit for other functions, avoiding this verifier limitation entirely.

### Unit tests added
- `TestProcessMapValue_TCPAcceptBasicWithError` — real error (ENOMEM) correctly reported
- `TestProcessMapValue_TCPAcceptBasicEAGAINNotInMap` — documents eBPF-level filtering
- `TestDropMetricKey_GetDirection` — direction mapping for all drop types
- `TestBuildFexitPrograms_ReturnsKprobeForAccept` — fexit objects return kprobe programs
- `TestBuildFexitPrograms_MarinerReturnsKprobeForAccept` — mariner variant
- `TestBuildFexitPrograms_KprobeObjectsReturnsNil` — kprobe-only mode unaffected
- `TestResolvePayload_FexitObjectsHaveKprobeFields` — integration of resolve + build
- `TestResolvePayload_KprobeObjectsNotAffected` — pod-level mode regression test

## Additional Notes

- The fexit `inet_csk_accept_fexit` program is still compiled into the BPF object (for 6.10+ kernels where it works via `proto_accept_arg`), but it is no longer attached as a fexit tracing program. Instead, the Go loader always attaches the kprobe/kretprobe pair for `inet_csk_accept`, which works correctly on all kernels.
- The `bpf_probe_read_kernel` call in the kretprobe is safe because the error pointer saved by the kprobe points to a stack variable in the caller (`sys_accept4`), which is still on-stack at kretprobe time.
- Affects all Retina versions v1.2.0–v1.2.2, all node OS types, on any kernel >= 5.5 with fexit support.
- Added `InetCskAccept` and `InetCskAcceptRet` fields to `allFexitPrograms` and `marinerPrograms` structs so the kprobe programs are loaded from the same BPF ELF alongside the fexit programs.

### DIFF
--- a/pkg/plugin/dropreason/_cprog/drop_reason.c
+++ b/pkg/plugin/dropreason/_cprog/drop_reason.c
@@ -32,6 +32,7 @@ char __license[] SEC("license") = "Dual MIT/GPL";
 #define ETH_P_8021Q 0x8100
 #define ETH_P_ARP 0x0806
 #define TASK_COMM_LEN 16
+#define EAGAIN_ERRNO 11
 // TODO (Vamsi): Add top 100 dropped connections with LRU map
 
 // Ref: https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/if_packet.h#L26
@@ -426,8 +427,8 @@ int BPF_KRETPROBE(inet_csk_accept_ret, struct sock *sk)
     if (bpf_probe_read_kernel(&err, sizeof(err), (void *)(unsigned long)*err_ptr) < 0)
         return 0;
 
-    // err >= 0 means no error; -EAGAIN (-11) is normal for non-blocking accept.
-    if (err >= 0 || err == -11)
+    // err >= 0 means no error; -EAGAIN is normal for non-blocking accept.
+    if (err >= 0 || err == -EAGAIN_ERRNO)
         return 0;
 
     struct packet p;
@@ -461,10 +462,11 @@ int BPF_PROG(inet_csk_accept_fexit)
         if ((struct sock *)ctx[2] == NULL) {
             // Read err from proto_accept_arg (ctx[1] is the arg pointer).
             int err = 0;
-            bpf_probe_read_kernel(&err, sizeof(err),
-                &((struct proto_accept_arg___new *)ctx[1])->err);
-            // EAGAIN (-11) is normal for non-blocking accept; not a real drop.
-            if (err != 0 && err != -11)
+            if (bpf_probe_read_kernel(&err, sizeof(err),
+                &((struct proto_accept_arg___new *)ctx[1])->err) < 0)
+                return 0;
+            // EAGAIN is normal for non-blocking accept; not a real drop.
+            if (err != 0 && err != -EAGAIN_ERRNO)
                 update_metrics_map_basic(TCP_ACCEPT_BASIC, err, 0);
         }
     } else {
--- a/pkg/plugin/dropreason/_cprog/drop_reason.c
+++ b/pkg/plugin/dropreason/_cprog/drop_reason.c
@@ -420,12 +420,9 @@ int BPF_KRETPROBE(inet_csk_accept_ret, struct sock *sk)
     if (sk != NULL)
         return 0;
 
-    // *err_ptr holds the ADDRESS of the callee's errno output (the int *err
-    // parameter pre-6.10, &arg->err on 6.10+); the errno must be read through it.
-    int err = 0;
-    if (bpf_probe_read_kernel(&err, sizeof(err), (void *)(unsigned long)*err_ptr) < 0)
-        return 0;
-    if (err >= 0)
+    int err = (int)*err_ptr;
+    // err >= 0 means no error; -EAGAIN (-11) is normal for non-blocking accept.
+    if (err >= 0 || err == -11)
         return 0;
 
     struct packet p;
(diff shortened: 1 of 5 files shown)


nddq:
@carlotaarvela There is an overlapping fix that I've done here (your first bulletpoint) https://github.com/microsoft/retina/pull/2504 specifically to address the flaky CI regarding dropreason, do you want to get that in first since it is one line to get the CI going first?

pkg/plugin/dropreason/_cprog/drop_reason.c
```diff
@@ -451,11 +458,25 @@ int BPF_PROG(inet_csk_accept_fexit)
...
+            if (err != 0 && err != -EAGAIN_ERRNO)
+                update_metrics_map_basic(TCP_ACCEPT_BASIC, err, 0);
+        }
     } else {
-        if ((struct sock *)ctx[4] == NULL)
-            update_metrics_map_basic(TCP_ACCEPT_BASIC, 0, 0);
+        // Pre-6.10: ctx = [sk, flags, err_ptr, kern, retsk]
+        // On older kernels (e.g. 5.15), the BPF verifier rejects reading ctx[2]
+        // because the BTF encodes the int* parameter as scalar type INT.
+        // Since we cannot read the error code, we skip counting entirely.
```
SRodi, line 476: This else branch is a silent no-op. Real accept errors on `5.15–6.9` kernels are now only reported by the `kprobe/kretprobe` pair. That's fine if the `kretprobe` fires reliably, but consider what happens if the `kprobe` on `inet_csk_accept`  is missed (`kprobe` recursion limit, tracer conflict, etc.): we lose ALL accept-error visibility on those kernels with no signal. I think we should log a one-time warning at attach time so operators know the `pre-6.10` `fexit` branch is intentionally silent

pkg/plugin/dropreason/dropreason_linux.go
```diff
@@ -177,10 +180,49 @@ func (dr *dropReason) Init() error {
...
 
 	if supportsFexit {
 		err = dr.attachFexitPrograms(progsFexit)
+		if err != nil {
+			return err
+		}
+		// Attach kprobe/kretprobe for inet_csk_accept alongside fexit programs.
+		// inet_csk_accept_fexit is not attached (pre-6.10 verifier limitation), so
+		// accept metrics depend on this kprobe/kretprobe pair being attached.
+		cleanupOnErr := func() {
```
SRodi: If `link.Kprobe / link.Kretprobe` for `inet_csk_accept` fails, `cleanupOnErr()` tears down every `fexit` attachment (`nf_hook_slow`, `tcp_v4_connect`, etc.) that was just successfully attached and returns an error. Previously (per `attachFexitPrograms`), individual attach failures were logged but the plugin kept running as long as `progCount >= 1` . This makes the plugin more brittle, one flaky kernel symbol takes down the whole `dropreason` plugin. I think we should log + continue (matching the existing style) instead of hard-failing.

pkg/plugin/dropreason/ebpfsetup_linux.go
```diff
@@ -185,21 +185,27 @@ func buildKprobePrograms(objs any) (progsKprobe, progsKprobeRet map[string]*ebpf
...
 }
 
-func buildFexitPrograms(objs any) map[string]*ebpf.Program {
-	progsFexit := make(map[string]*ebpf.Program)
+func buildFexitPrograms(objs any) (progsFexit map[string]*ebpf.Program, acceptKprobe, acceptKretprobe *ebpf.Program) {
+	progsFexit = make(map[string]*ebpf.Program)
 
 	switch o := objs.(type) {
 	case *allFexitObjects:
-		progsFexit[inetCskAcceptFnFexit] = o.InetCskAcceptFexit
```
SRodi, line 193: The fexit `inet_csk_accept_fexit` program is loaded but never attached, the PR unconditionally uses the `kprobe/kretprobe` fallback even on `6.10+ kernels` where `fexit` would be faster and immune to `kretprobe`-miss under load. This is acceptable as a stop-the-bleeding fix, but please file a follow-up to attach `fexit` on `6.10+` (gated by `bpf_core_type_exists(struct proto_accept_arg___new)` or a Go-side kernel version check) and only fall back to `kprobe/kretprobe` on `5.5–6.9`. Otherwise Retina permanently loses the perf advantage of `fexit` on this hot path.

SRodi (commented):
@carlotaarvela IMO We can ship this PR (the correctness fix is urgent), but file a P1 follow-up to gate the `kprobe/kretprobe` fallback on `!bpf_core_type_exists(struct proto_accept_arg___new)`, i.e. only use `kprobe/kretprobe` on `5.5–6.9` and keep `fexit` on `6.10+`. Users on modern kernels should get `fexit` back as soon as the pre-6.10 verifier issue is properly isolated.

pkg/plugin/dropreason/dropreason_linux.go
carlotaarvela: Done

pkg/plugin/dropreason/_cprog/drop_reason.c
carlotaarvela: Done

PR #2282, opened 2026-05-04
fix(helm): use native gRPC probe on relay :4222 and bump default to v1.17.15

pesarkhobeee:
Closes #2165

> @nddq external contributors can't self-assign on this repo — could you assign #2165 to me, please?

## Summary

1. **`templates/hubble-relay/deployment.yaml`** — replace the conditional `grpc_health_probe` exec / native gRPC probe branching with a single native `grpc:` probe pointed at the relay's dedicated health server on `:4222`. Works regardless of `hubble.tls.enabled` (the health listener is plaintext; only the main API on `listenPort` is TLS-wrapped). Mirrors what upstream Cilium's chart adopted after [cilium/cilium#37806](https://github.com/cilium/cilium/pull/37806).
2. **`values.yaml`** — bump default `hubble.relay.image` from `mcr.microsoft.com/oss/cilium/hubble-relay:v1.15.0` to `quay.io/cilium/hubble-relay:v1.17.15` (lowest currently-supported Cilium minor at its latest patch, smallest jump that lands on a maintained line). Refs: [v1.17.15 release notes](https://github.com/cilium/cilium/releases/tag/v1.17.15), [available tags](https://quay.io/repository/cilium/hubble-relay?tab=tags).

## Why

With the chart's default `hubble.tls.enabled: true`, the probe template selects the `exec grpc_health_probe` branch. Upstream Cilium [removed that binary from the relay image](https://github.com/cilium/cilium/pull/37806) (backported to all supported branches), so any v1.16+ tag fails the startup probe with `executable file not found in $PATH` and the rollout stalls. Reproduced against `v1.17.15`.

Probing `:4222` instead of `listenPort` (`4245`) avoids this without forcing operators to disable TLS or ship a custom relay image — the health server is always plaintext regardless of the main API's TLS state.

PR #2500, opened 2026-07-06
fix(helm): use native gRPC probe on relay :4222 and bump default to v1.19.5

slariviere:
Closes #2165

Address review comments from https://github.com/microsoft/retina/pull/2282 (closed due to inactivity)

## Summary

Fixes hubble-relay's startup/readiness/liveness probes, which fail on Cilium
v1.16+ images because the `grpc_health_probe` binary used by the legacy
exec-based probe was removed upstream ([cilium/cilium#37806](https://github.com/cilium/cilium/pull/37806)).

1. **`templates/hubble-relay/deployment.yaml`** — single native `grpc:` probe
   against the relay's dedicated health server (`hubble.relay.grpcHealthPort`,
   default `4222`), replacing the k8s-version-gated exec/native branching.
   The health listener is always plaintext regardless of `hubble.tls.enabled`
   (only the main API on `listenPort` is TLS-wrapped), so this probe works in
   both TLS and non-TLS deployments. Also renders the relay image via the
   `cilium.image` helper (previously a raw `repository:tag` string) so
   `useDigest`/`override` are actually honored.
2. **`templates/hubble-relay/configmap.yaml`** — wires `hubble.relay.grpcHealthPort`
   through to `health-listen-address`, so the relay's health server actually
   listens on the port the probes target.
3. **`templates/hubble-ui/deployment.yaml`** — frontend and backend images now
   also render via the `cilium.image` helper.
4. **`values.yaml`**:
   - `hubble.relay.image` default bumped from
     `mcr.microsoft.com/oss/cilium/hubble-relay:v1.15.0` to
     `quay.io/cilium/hubble-relay:v1.19.5` (matches the Cilium `v1.19.3`
     dependency in `go.mod`), with `useDigest: true`.
   - Added `hubble.relay.grpcHealthPort: "4222"`.
   - `hubble.ui.backend.image` and `hubble.ui.frontend.image` both moved from
     `mcr.microsoft.com` to `quay.io/cilium/hubble-ui{,-backend}:v0.13.5`
     (previously split across two registries at `v0.12.2`), matching upstream
     Cilium's `v1.19.5` chart image pins.

### DIFF
--- a/deploy/hubble/manifests/controller/helm/retina/templates/hubble-relay/deployment.yaml
+++ b/deploy/hubble/manifests/controller/helm/retina/templates/hubble-relay/deployment.yaml
@@ -55,7 +55,7 @@ spec:
           securityContext:
             {{- toYaml . | nindent 12 }}
           {{- end }}
-          image: "{{ .Values.hubble.relay.image.repository }}:{{ .Values.hubble.relay.image.tag }}"
+          image: {{ include "cilium.image" .Values.hubble.relay.image | quote }}
           imagePullPolicy: {{ .Values.hubble.relay.image.pullPolicy }}
           command:
             - hubble-relay
@@ -193,24 +193,15 @@ spec:
 {{- end }}
 
 {{- define "hubble-relay.probe" }}
-{{- /* This distinction can be removed once we drop support for k8s 1.23 */}}
-{{- if and (semverCompare ">=1.24-0" .Capabilities.KubeVersion.Version) (not .Values.hubble.tls.enabled) -}}
+{{- /*
+Probe the relay's dedicated gRPC health server on :4222. It is always plaintext
+regardless of hubble.tls.enabled (the TLS-enabled listener is the main API on
+listenPort), so the kubelet's native gRPC probe works in both TLS and non-TLS
+deployments. Upstream Cilium removed the grpc_health_probe binary from the
+hubble-relay image (cilium/cilium#37806), so the prior exec-based probe is no
+longer viable on v1.16+ images.
+*/ -}}
 grpc:
-  port: {{ .Values.hubble.relay.listenPort }}
-{{- else }}
-exec:
-  command:
-  - grpc_health_probe
-  - -addr=localhost:{{ .Values.hubble.relay.listenPort }}
-{{- if .Values.hubble.tls.enabled }}
-  - -tls
-  - -tls-ca-cert=/var/lib/hubble-relay/tls/hubble-server-ca.crt
-  - -tls-client-cert=/var/lib/hubble-relay/tls/client.crt
-  - -tls-client-key=/var/lib/hubble-relay/tls/client.key
-  - -tls-server-name=instance.hubble-relay.cilium.io
-  - -connect-timeout=5s
-  - -rpc-timeout=5s
-{{- end }}
-{{- end }}
+  port: {{ .Values.hubble.relay.grpcHealthPort }}
 timeoutSeconds: 3
 {{- end }}
--- a/deploy/hubble/manifests/controller/helm/retina/values.yaml
+++ b/deploy/hubble/manifests/controller/helm/retina/values.yaml
@@ -429,12 +429,14 @@ hubble:
     rollOutPods: false
 
     # -- Hubble-relay container image.
+    # Official Cilium registry. Pick the latest patch of a maintained Cilium
+    # minor: https://quay.io/repository/cilium/hubble-relay?tab=tags
     image:
       override: ~
-      repository: "mcr.microsoft.com/oss/cilium/hubble-relay"
-      tag: "v1.15.0"
-      digest: "sha256:19cd56e7618832257bf88b2f281287cb57f9f7fcb9e04775a6198d4bc4daffae"
-      useDigest: false
+      repository: "quay.io/cilium/hubble-relay"
+      tag: "v1.19.5"
+      digest: "sha256:24409bfa1bca075c92acb26ba4b49cd573d99d68d5370f7cc825078185222a0c"
+      useDigest: true
       pullPolicy: "Always"
 
     # -- Specifies the resources for the hubble-relay pods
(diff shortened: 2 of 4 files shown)


deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -428,12 +428,19 @@ hubble:
...
     image:
       override: ~
-      repository: "mcr.microsoft.com/oss/cilium/hubble-relay"
-      tag: "v1.15.0"
-      digest: "sha256:19cd56e7618832257bf88b2f281287cb57f9f7fcb9e04775a6198d4bc4daffae"
-      useDigest: false
+      repository: "quay.io/cilium/hubble-relay"
+      tag: "v1.17.15"
+      digest: "sha256:60dcac76e5841a14d5c4813377cb463822db78568146e8c93ffc5b5cc0e894fb"
+      useDigest: true
```
nddq, line 439: Flipping this to `true` doesn't take effect: the relay deployment renders the image as plain `repo:tag` (`templates/hubble-relay/deployment.yaml` line 58), so the digest is never emitted. We should switch the relay (and the hubble-ui frontend/backend at `templates/hubble-ui/deployment.yaml` lines 60/96) to the existing helper:
```yaml
image: {{ include "cilium.image" .Values.hubble.relay.image | quote }}
```

`cilium.image` (in `_helpers.tpl`) already handles `override` and `useDigest`.

deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -541,6 +548,9 @@ hubble:
     # -- Port to listen to.
     listenPort: "4245"
 
+    # -- Port for the relay's dedicated gRPC health server
+    grpcHealthPort: "4222"
```
nddq, line 548: As-is this value is a trap: overriding it moves the probes but the relay keeps listening on its built-in default `:4222`, because the configmap never sets `health-listen-address`. `helm template --set hubble.relay.grpcHealthPort=9999` renders all three probes on 9999 against a listener on 4222 — the pod can never become ready. Please wire it through in `templates/hubble-relay/configmap.yaml`:

```yaml
health-listen-address: ":{{ .Values.hubble.relay.grpcHealthPort }}"
```

(or drop the value and hardcode 4222 in the probe like upstream cilium does.)

deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -428,12 +428,19 @@ hubble:
...
+    # available tags.
     image:
       override: ~
-      repository: "mcr.microsoft.com/oss/cilium/hubble-relay"
-      tag: "v1.15.0"
-      digest: "sha256:19cd56e7618832257bf88b2f281287cb57f9f7fcb9e04775a6198d4bc4daffae"
-      useDigest: false
+      repository: "quay.io/cilium/hubble-relay"
+      tag: "v1.17.15"
+      digest: "sha256:60dcac76e5841a14d5c4813377cb463822db78568146e8c93ffc5b5cc0e894fb"
```
nddq: Since our `go.mod` is on cilium v1.19.3, let's land on the v1.19 line instead

```suggestion
      tag: "v1.19.5"
      digest: "sha256:24409bfa1bca075c92acb26ba4b49cd573d99d68d5370f7cc825078185222a0c"
```

deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -428,12 +428,19 @@ hubble:
     rollOutPods: false
 
     # -- Hubble-relay container image.
+    # Pinned to the official Cilium registry (quay.io/cilium/hubble-relay) and
+    # to v1.17.15 — the lowest currently-supported minor (1.17) at its latest
+    # patch. This intentionally minimises the major jump from the previous
+    # default (v1.15.0) so users have the smallest supported delta to validate.
+    # See https://github.com/cilium/cilium/releases/tag/v1.17.15 for release
+    # notes and https://quay.io/repository/cilium/hubble-relay?tab=tags for
+    # available tags.
```
nddq: This comment will be stale the next time we bump this, keep just the durable part:

```suggestion
    # Official Cilium registry. Pick the latest patch of a maintained Cilium
    # minor: https://quay.io/repository/cilium/hubble-relay?tab=tags
```

deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -674,9 +684,9 @@ hubble:
       # -- Hubble-ui backend image.
       image:
         override: ~
-        repository: "mcr.microsoft.com/oss/cilium/hubble-ui-backend"
+        repository: "quay.io/cilium/hubble-ui-backend"
```
nddq, line 683: This hubble-ui-backend registry change isn't mentioned in the PR description, we should add it. It also leaves the hubble ui front end on MCR, so the same pod pulls from two registries. Let's either move both frontend+backend to quay.io at v0.12.2, or drop the ui change and keep this PR relay-only.

deploy/hubble/manifests/controller/helm/retina/templates/hubble-relay/deployment.yaml
```diff
@@ -193,24 +193,15 @@ spec:
...
-  - -tls
-  - -tls-ca-cert=/var/lib/hubble-relay/tls/hubble-server-ca.crt
-  - -tls-client-cert=/var/lib/hubble-relay/tls/client.crt
-  - -tls-client-key=/var/lib/hubble-relay/tls/client.key
-  - -tls-server-name=instance.hubble-relay.cilium.io
-  - -connect-timeout=5s
-  - -rpc-timeout=5s
-{{- end }}
-{{- end }}
+  port: {{ .Values.hubble.relay.grpcHealthPort | default 4222 }}
```
nddq:
```suggestion
*/ -}}
grpc:
  port: {{ .Values.hubble.relay.grpcHealthPort }}
```

nddq (changes requested):

nddq (approved):
lgtm, thanks!

nddq:
@slariviere All commits need to have verified signatures

slariviere:
@nddq should be good now

nddq (approved):

PR #2503, opened 2026-07-08
feat(packetparser): report dropped events on buffer write failure

### LINKED ISSUE #2235: Monitoring for lost samples in kernel space and failed packetparser reports for ring buffers

With perf buffers, we can monitor for lost samples via https://github.com/cilium/ebpf/blob/e4d11b5080301912e47c99b84bb522681d600b62/perf/reader.go#L49.

Even then, there's no reporting from the eBPF program when writing to the buffer fails.

**Describe the solution you'd like**

Metrics for lost events should reflect both fails to write to the ring buffer/perf buffer as well as lost reports from userspace.

**Additional context**

It is difficult to tune the ring buffer size and sampling rate without knowing if any reports are being dropped.

mmckeen:
# Description

For ring buffers, we can't get dropped events from user space.  Instead, track drop metrics in eBPF that user space then reports.

Re-use this logic for perf buffers as well.

## Related Issue

https://github.com/microsoft/retina/issues/2235

## Checklist

Deployed into a test cluster with perf and ring buffers, drop metrics show up for ring buffers and perf metrics are the same.

---

### DIFF
--- a/pkg/plugin/packetparser/_cprog/packetparser.c
+++ b/pkg/plugin/packetparser/_cprog/packetparser.c
@@ -28,6 +28,31 @@ struct
 #endif
 } retina_packetparser_events SEC(".maps");
 
+// Counts events dropped in the kernel when the event buffer write fails.
+struct
+{
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, __u32);
+	__type(value, __u64);
+} retina_packetparser_metrics SEC(".maps");
+
+static __always_inline void submit_packetparser_event(struct __sk_buff *skb, struct packet *p)
+{
+	long ret;
+#ifdef USE_RING_BUFFER
+	ret = bpf_ringbuf_output(&retina_packetparser_events, p, sizeof(*p), 0);
+#else
+	ret = bpf_perf_event_output(skb, &retina_packetparser_events, BPF_F_CURRENT_CPU, p, sizeof(*p));
+#endif
+	if (ret < 0) {
+		__u32 zero = 0;
+		__u64 *dropped = bpf_map_lookup_elem(&retina_packetparser_metrics, &zero);
+		if (dropped)
+			*dropped += 1;
+	}
+}
+
 // Define const variables to avoid warnings.
 const struct packet *unused __attribute__((unused));
 
@@ -239,11 +264,7 @@ static void parse(struct __sk_buff *skb, __u8 obs)
 		p.previously_observed_packets = 0;
 		p.previously_observed_bytes = 0;
 		__builtin_memset(&p.previously_observed_flags, 0, sizeof(struct tcpflagscount));
-#ifdef USE_RING_BUFFER
-		bpf_ringbuf_output(&retina_packetparser_events, &p, sizeof(p), 0);
-#else
-		bpf_perf_event_output(skb, &retina_packetparser_events, BPF_F_CURRENT_CPU, &p, sizeof(p));
-#endif
+		submit_packetparser_event(skb, &p);
 		return;
 	// If the data aggregation level is high, only send the packet to the perf buffer if it needs to be reported.
 	#elif DATA_AGGREGATION_LEVEL == DATA_AGGREGATION_LEVEL_HIGH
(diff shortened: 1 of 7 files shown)


pkg/plugin/packetparser/packetparser_linux_test.go
nddq, line 1: Can we also add ebpf tests for the dropped event counter itself? I think we can fill up the buffer in the tests and then assert the count to be > 0

pkg/plugin/packetparser/_cprog/packetparser.c
```diff
@@ -28,6 +28,31 @@ struct
...
+	long ret;
+#ifdef USE_RING_BUFFER
+	ret = bpf_ringbuf_output(&retina_packetparser_events, p, sizeof(*p), 0);
+#else
+	ret = bpf_perf_event_output(skb, &retina_packetparser_events, BPF_F_CURRENT_CPU, p, sizeof(*p));
+#endif
+	if (ret < 0) {
+		__u32 zero = 0;
+		__u64 *dropped = bpf_map_lookup_elem(&retina_packetparser_metrics, &zero);
+		if (dropped)
```
nddq, line 53: I think we can use `__sync_fetch_and_add(dropped, 1)` here

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -942,6 +946,63 @@ func (p *packetParser) readData() {
 	}
 }
 
+// reportLostEvents reports kernel-side event drops (recorded in eBPF because ring
+// buffers surface no lost-sample signal to userspace) on the metrics interval.
+func (p *packetParser) reportLostEvents(ctx context.Context) {
```
nddq, line 967: nothing joins this goroutine, so `Stop()` can close `p.objs` while we're mid-Lookup. Since `ctx` is cancelled before `Stop()` in all shutdown paths, we can close a done channel when this exits and wait on it in `Stop()` before `p.objs.Close()`, something like:

```diff
--- a/pkg/plugin/packetparser/types_linux.go
+++ b/pkg/plugin/packetparser/types_linux.go
@@ -152,6 +152,9 @@ type packetParser struct {
 	wg                  sync.WaitGroup
 	recordsChannel      chan perfRecord
+	// reporterDone is closed when reportLostEvents exits; Stop waits on it
+	// before closing the eBPF objects the reporter reads.
+	reporterDone    chan struct{}
 	externalChannel chan *v1.Event
 	tcxSupported    bool // Whether TCX is supported on this system

--- a/pkg/plugin/packetparser/packetparser_linux.go
+++ b/pkg/plugin/packetparser/packetparser_linux.go
@@ -386,6 +386,7 @@ func (p *packetParser) Start(ctx context.Context) error {
 	// Create the channel.
 	p.recordsChannel = make(chan perfRecord, buffer)
+	p.reporterDone = make(chan struct{})
 	p.l.Debug("Created records channel")

@@ -412,6 +413,12 @@ func (p *packetParser) Stop() error {
 		p.l.Debug("Closed records channel")
 	}

+	// Wait for the drop reporter to exit before closing the map it reads.
+	// The context is cancelled before Stop in the shutdown paths, so this is bounded.
+	if p.reporterDone != nil {
+		<-p.reporterDone
+	}
+
 	// Stop map and progs.
 	if p.objs != nil {

@@ -949,6 +956,7 @@ func (p *packetParser) readData() {
 func (p *packetParser) reportLostEvents(ctx context.Context) {
+	defer close(p.reporterDone)
 	if p.objs == nil || p.objs.RetinaPacketparserMetrics == nil {
 		return
 	}
```

this builds and passes the unit tests locally. The nil guard keeps the pre-`Start` `Stop()` calls (initial `Reconcile`, pod level off) from blocking, and a second `Stop()` just re-receives from the closed channel.

pkg/plugin/packetparser/packetparser_linux.go
```diff
@@ -942,6 +946,63 @@ func (p *packetParser) readData() {
...
+	if p.objs == nil || p.objs.RetinaPacketparserMetrics == nil {
```
nddq, line 970: can this ever be nil here? `run` only executes after `Init` has set `p.objs`, and `LoadAndAssign` would have errored if the map was missing. If we are keeping this as a safety net, we should at least log here since this silently disables drop reporting

nddq (changes requested):

mmckeen:
@nddq updated 👍

nddq:
@mmckeen can you rebase the PR with latest main as well, this has drifted a lot

PR #2527, opened 2026-07-13
feat: Add rotating capture support (--file-count) for long-running packet captures

carlotaarvela:
# Description

Adds rotating capture support to Retina's packet capture system using tcpdump's `-W` (file count) and `-C` (file size) flags. This enables long-running captures that maintain a circular buffer of recent network traffic without unbounded disk growth.

When `--file-count` is specified alongside `--max-size`, tcpdump rotates through a fixed number of capture files, overwriting the oldest when the limit is reached. For example, `--file-count 3 --max-size 100` keeps approximately the last 300MB of network traffic across 3 files.

### Key changes:

- **CRD**: Added `fileCount` field to `CaptureOption` with kubebuilder validation (minimum 1)
- **CLI**: Added `--file-count` flag with client-side validation (requires `--max-size`)
- **Provider (Linux)**: When `fileCount > 0`, appends `-C` and `-W` to tcpdump and skips manual file-size polling
- **Provider (Windows)**: Signature updated; rotating capture is unsupported on Windows (netsh limitation)
- **Capture Manager**: Added `captureFileCount()` env parsing; fixed `captureDuration()` to handle empty string (for captures without duration)
- **Capture Workload**: Split context into `captureCtx` (cancelled on SIGTERM) and `outputCtx` (fresh, for blob/S3 uploads after capture stops)
- **Validation**: `fileCount` requires `maxCaptureSize`; captures with `fileCount` don't require `duration` (can be stopped manually)

## Related Issue

N/A — feature request for long-running capture support with bounded disk usage.

## Checklist

### Unit Tests

All new logic is covered:

| Test | What it verifies |
|------|-----------------|
| `TestTcpdumpRotatingCaptureArgs` | `-C`/`-W` flag construction (5 subtests: with/without fileCount and maxSize) |
| `TestCaptureFileCount` | `captureFileCount()` env parsing (empty, valid, invalid) |
| `TestCaptureDuration` | `captureDuration()` returns 0 for empty string |
| `TestCaptureNetworkWithRotatingCapture` | CaptureManager passes fileCount to provider |
| `TestCaptureNetworkWithNoDuration` | CaptureManager handles missing duration env |
| `TestCompressFolderToTarGzIncludesRotatedFiles` | Tar archive includes all `.pcap0`, `.pcap1`, etc. files |
| `TestCompressFolderToTarGzPreservesContent` | File content preserved through compression |
| `Test_CaptureToPodTranslator_ValidateCapture` | fileCount validation (3 new subtests) |
| `Test_CaptureToPodTranslator_ObtainCaptureJobPodEnv` | CAPTURE_FILE_COUNT env var passed to job |

### Live Cluster E2E (manual)

Tested on AKS cluster with custom image:

- **Rotating capture (5 min, 3 files x 1MB)**: tcpdump ran with `-C 1 -W 3`. After 5 minutes, downloaded tar.gz contained 3 pcap files (`.pcap0`, `.pcap1`, `.pcap2`). Timestamps confirmed rotation — the 5-minute capture only retained the last ~15 seconds of traffic, proving the circular buffer overwrote earlier data correctly.
- **Standard capture (backward compat)**: Capture without `--file-count` works identically to before.
- **CLI validation**: `--file-count=-1` rejected; `--file-count` without `--max-size` rejected.

## Additional Notes

- **Backward compatible**: `fileCount` defaults to 0/nil, preserving existing behavior exactly.
- **Windows**: The `fileCount` parameter is accepted in the interface but has no effect on Windows (netsh doesn't support file rotation). This matches the existing pattern where `maxCaptureSize` is also Linux-only.
- **Context fix in captureworkload/main.go**: This fixes a pre-existing latent bug where SIGTERM cancellation would also cancel blob/S3 uploads. For rotating captures this becomes critical since SIGTERM is the expected stop mechanism.


cli/cmd/capture/create.go
```diff
@@ -74,7 +74,11 @@ const (
 	MinPollAttempts = 4
 )
 
-var errCleanupRequiresRemoteStorage = errors.New("--cleanup-after-upload requires remote storage (--blob-upload, --s3-bucket, or --pvc)")
+var (
+	errCleanupRequiresRemoteStorage = errors.New("--cleanup-after-upload requires remote storage (--blob-upload, --s3-bucket, or --pvc)")
+	errFileCountTooSmall            = errors.New("--file-count must be at least 1")
+	errFileCountRequiresMaxSize     = errors.New("--file-count requires --max-size to be set as per-file size limit")
```
mereta, line 81: nit: sync naming between CLI & CRD

pkg/capture/crd_to_job.go
```diff
@@ -34,10 +34,11 @@ import (
...
 var (
-	errNoTargetsSelected    = errors.New("no targets are selected by node selector, pod selector, or pod names")
-	errNoValidSelector      = errors.New("neither NodeSelector, NamespaceSelector&PodSelector, nor PodNames is set")
-	errNodeSelectorIncompat = errors.New("NodeSelector is not compatible with NamespaceSelector&PodSelector or PodNames, please use one or the other")
-	errPodNamesIncompat     = errors.New("PodNames is not compatible with NamespaceSelector or PodSelector, please use one or the other")
+	errNoTargetsSelected       = errors.New("no targets are selected by node selector, pod selector, or pod names")
+	errNoValidSelector         = errors.New("neither NodeSelector, NamespaceSelector&PodSelector, nor PodNames is set")
+	errNodeSelectorIncompat    = errors.New("NodeSelector is not compatible with NamespaceSelector&PodSelector or PodNames, please use one or the other")
+	errPodNamesIncompat        = errors.New("PodNames is not compatible with NamespaceSelector or PodSelector, please use one or the other")
+	errFileCountRequiresMaxCap = errors.New("fileCount requires maxCaptureSize to be set as per-file size limit")
```
mereta: nit: Would rename to MaxCaptureSize or MaxSize to match the CLI naming convention.

cli/cmd/capture/create.go
```diff
@@ -256,7 +260,10 @@ func NewCreateSubCommand(kubeClient kubernetes.Interface) *cobra.Command {
 	}
 
 	createCapture.Flags().DurationVar(&opts.duration, "duration", DefaultDuration, "Duration of capturing packets")
-	createCapture.Flags().IntVar(&opts.maxSize, "max-size", DefaultMaxSize, "Limit the capture file to MB in size which works only for Linux") //nolint:gomnd // default
+	createCapture.Flags().IntVar(&opts.maxSize, "max-size", DefaultMaxSize,
+		"Limit the capture file to MB in size (per-file size when used with --file-count). Linux only") //nolint:gomnd // default
+	createCapture.Flags().IntVar(&opts.fileCount, "file-count", 0,
```
mereta: Instead of defining zero here we could follow the pattern and add a default value var on line 58.

carlotaarvela: Done

pkg/capture/crd_to_job.go
carlotaarvela: Done

cli/cmd/capture/create.go
carlotaarvela: Done

mereta (approved):
LGTM

PR #2591, opened 2026-07-28
fix(chart): default agent and operator images to ghcr.io

darkweaver87:
# Motivation
The retina-hubble chart defaults the agent and operator images to `acndev.azurecr.io`, an internal registry that isn't reachable outside Microsoft, so a stock `helm install` of the published chart lands in `ImagePullBackOff`. The `retina-init` image, and the `standard` chart's agent/operator, already default to `ghcr.io/microsoft/retina/*`; this aligns the hubble chart with them. Only the two default repositories change — rendered output is otherwise identical.


deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -17,7 +17,7 @@ os:
 operator:
   enabled: true
   replicas: 1
-  repository: acndev.azurecr.io/retina-operator
+  repository: ghcr.io/microsoft/retina/retina-operator
   pullPolicy: Always
   tag: "latest"
```
nddq: I don't think we've ever published `latest` tags for any images, we should pin this to a released version tag

nddq (changes requested):

deploy/hubble/manifests/controller/helm/retina/values.yaml
darkweaver87: Good catch — done. The three Retina image tags now default to `.Chart.Version` (the released chart version, e.g. `v1.2.3`) when left empty, instead of `latest`; `--set <image>.tag=…` still overrides. Verified with `helm package --version` + `helm template` — which is exactly what `release-charts.yaml` (L55-56) does, so the published chart's `.Chart.Version` is the release tag.

deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -17,9 +17,10 @@ os:
 operator:
   enabled: true
   replicas: 1
-  repository: acndev.azurecr.io/retina-operator
+  repository: ghcr.io/microsoft/retina/retina-operator
   pullPolicy: Always
-  tag: "latest"
+  # -- Image tag; when empty, defaults to the chart version (the released version).
+  tag: ""
```
nddq, line 23: nit: `.Chart.Version` is only the release tag for the packaged chart, since we don't update those, from a source-tree install it is still 0.0.1, so `helm install ./deploy/hubble/... renders retina-agent:0.0.1`. I'd say "the chart version" and drop the parenthetical.

deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -41,14 +42,14 @@ agent:
   leaderElection: false
   enabled: true
   name: retina-agent
-  repository: acndev.azurecr.io/retina-agent
-  tag: "latest"
+  repository: ghcr.io/microsoft/retina/retina-agent
+  tag: ""
```
nddq, line 47: we can mirror the comments for the retina-operator's tag here as well + agent-init

deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -41,14 +42,14 @@ agent:
...
   init:
     enabled: true
     name: retina-agent-init
     repository: ghcr.io/microsoft/retina/retina-init
```
nddq: drop this whitespace

nddq:
```suggestion
```

nddq (commented):
We should expand this fix to the default charts as well, since they are being hardcoded to v0.0.2.

darkweaver87:
@nddq thanks for your review. I addressed your comments :-)

deploy/standard/manifests/controller/helm/retina/values.yaml
```diff
@@ -47,8 +48,8 @@ image:
   repository: ghcr.io/microsoft/retina/retina-agent
   initRepository: ghcr.io/microsoft/retina/retina-init
   pullPolicy: Always
-  # Overrides the image tag whose default is the chart appVersion.
-  tag: "v0.0.2"
+  # -- Image tag; when empty, defaults to the chart version.
+  tag: ""
```
nddq, line 52: This regresses the source-tree install. `Chart.yaml` is still `0.0.1`, so `helm install retina ./deploy/standard/manifests/controller/helm/retina/` renders `retina-agent:0.0.1`. The registry returns `404` for that tag and `200` for `v0.0.2`. `docs/04-Captures/04-managed-storage-account.md` L59 uses that local path with no tag override, so it now ends in `ImagePullBackOff`.

Let's point it at the published chart instead:

```diff
-helm upgrade --install retina ./deploy/standard/manifests/controller/helm/retina/ \
+VERSION=$( curl -sL https://api.github.com/repos/microsoft/retina/releases/latest | jq -r .name)
+helm upgrade --install retina oci://ghcr.io/microsoft/retina/charts/retina \
+  --version $VERSION \
   --set operator.enabled=true \
```

`templates/NOTES.txt` L1 points at the same local path. That one is pre-existing, so a follow-up is fine.

deploy/hubble/manifests/controller/helm/retina/values.yaml
```diff
@@ -17,9 +17,10 @@ os:
 operator:
   enabled: true
   replicas: 1
-  repository: acndev.azurecr.io/retina-operator
+  repository: ghcr.io/microsoft/retina/retina-operator
```
nddq, line 20: The Hubble block in `docs/02-Installation/01-Setup.md` still sets three repositories that now match the defaults. I rendered the packaged chart with and without those flags. The output is identical. Let's drop L133, L136 and L139:

```diff
         --set operator.enabled=true \
-        --set operator.repository=ghcr.io/microsoft/retina/retina-operator \
         --set operator.tag=$VERSION \
         --set agent.enabled=true \
-        --set agent.repository=ghcr.io/microsoft/retina/retina-agent \
         --set agent.tag=$VERSION \
         --set agent.init.enabled=true \
-        --set agent.init.repository=ghcr.io/microsoft/retina/retina-init \
         --set agent.init.tag=$VERSION \
```

The `--set *.tag=$VERSION` lines are dead now too, in both charts' blocks. That sweep is larger, so a follow-up is fine.

nddq (changes requested):

darkweaver87:
@nddq both addressed.

One correction: `templates/NOTES.txt` isn't pre-existing — the tag was hardcoded to `v0.0.2` before this PR, so this PR breaks it. Fixed here, along with `docs/04-Captures/03-crd.md` L45 (pasted NOTES output). The Makefile helm targets pass `--set image.tag`, so they're unaffected.

Hubble block: dropped L133/136/139, render is identical.

PR #2592, opened 2026-07-28
feat(chart): make retina-operator resources configurable

darkweaver87:
# Motivation
The retina-operator container's resources are hardcoded in the deployment template, so they can't be tuned without patching the rendered manifest — while the agent, relay and ui all expose `resources` in values. This adds `operator.resources`, defaulted to the current values, so rendered output is unchanged.


darkweaver87:
@microsoft-github-policy-service agree company="TraefikLabs"

rbtr:
lgtm but same signing feedback as the other PR

darkweaver87:
Thanks @rbtr, done :-)

rbtr (approved):

PR #953, opened 2024-11-06
fix: reduce cardinality for tcp remote address metrics

vakalapa:
# Description

This PR makes the following changes:

- **Remove Port Label**: The port label is removed from the TCP remote address metrics.

- **Aggregate IP Addresses**: Instead of tracking individual IP addresses, the metrics now use a generalized `AllIPs` string to represent all IP addresses.

image

## Related Issue

If this pull request is related to any issue, please mention it here. Additionally, make sure that the issue is assigned to you before submitting this pull request.

## Checklist

- [ ] I have read the [contributing documentation](https://retina.sh/docs/contributing).
- [ ] I signed and signed-off the commits (`git commit -S -s ...`). See [this documentation](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) on signing commits.
- [ ] I have correctly attributed the author(s) of the code.
- [ ] I have tested the changes locally.
- [ ] I have followed the project's style guidelines.
- [ ] I have updated the documentation, if necessary.
- [ ] I have added tests, if applicable.

## Screenshots (if applicable) or Testing Completed

Please add any relevant screenshots or GIFs to showcase the changes made.

## Additional Notes

Add any additional notes or context about the pull request here.

---

Please refer to the [CONTRIBUTING.md](../CONTRIBUTING.md) file for more information on how to contribute to this project.

PR #2608, opened 2026-08-04
docs: correct tcp_connection_remote label documentation

### LINKED ISSUE #2559: networkobservability_tcp_connection_remote only emits address=AllIPs instead of per-remote address labels
## Summary

`networkobservability_tcp_connection_remote` is documented as a per-remote IP/port metric, but the current implementation only emits a single aggregated series with `address="AllIPs"`.

We are specifically interested in having this supported in the lighter Hubble-based Retina deployment as well, since that is the mode we are using in production.

## Expected behavior

The metric description and docs suggest this should expose one series per active remote address/port pair, for example:

- `networkobservability_tcp_connection_remote{address="10.0.0.1",port="443"}`
- `networkobservability_tcp_connection_remote{address="10.0.0.2",port="5432"}`

From our perspective, this would be especially valuable if available in the Hubble-based Retina deployment, not just in a heavier/full Retina setup.

## Actual behavior

The current code sums all valid remote addresses into a single total and exports only:

- `networkobservability_tcp_connection_remote{address="AllIPs"}`

Relevant code:

- https://github.com/microsoft/retina/blob/577d9179b53c61697b5caf309fb4508244083146/pkg/plugin/linuxutil/netstat_stats_linux.go#L236-L245

The constant is also defined here:

- https://github.com/microsoft/retina/blob/577d9179b53c61697b5caf309fb4508244083146/pkg/plugin/linuxutil/netstat_stats_linux.go#L19-L23

## Why this looks incomplete

The implementation already builds `socketByRemoteAddr` in `processSocks()`, but `updateMetrics()` only validates and sums those entries instead of exporting them individually.

It seems likely the missing piece is handling series cleanup across polls, i.e. setting previously seen remote address/port label combinations to `0` or otherwise removing stale series when they disappear from the current snapshot.

## Impact

This makes the metric much less useful for diagnosing network issues. Dashboards that group `by(address)` end up with a single `AllIPs` series instead of a breakdown by remote endpoint.

In our case, this is affecting a Hubble-based Retina deployment where we want a lightweight but still actionable TCP remote-endpoint diagnostic signal.

## Docs mismatch

These docs describe the metric as per remote IP/port:

- https://github.com/microsoft/retina/blob/main/docs/03-Metrics/02-hubble_metrics.md
- https://github.com/microsoft/retina/blob/main/docs/03-Metrics/modes/basic.md

## Suggested fix

Either:

1. Implement per-remote address/port series export for `networkobservability_tcp_connection_remote`, including stale-series handling between polls

or

2. Update the docs/metric description to reflect that the metric is intentionally aggregated into `AllIPs`

A real per-remote series would be much more useful if cardinality is acceptable or can be gated/configured, and we would especially like to see that supported in the Hubble-based Retina deployment path.

xRookieFight:
# Description

Fixes #2559.

The docs describe `tcp_connection_remote` as a per-remote-endpoint metric with `address` (IP:port) labels, but that hasn't been true since #953, which deliberately dropped the port label and folded every remote socket into a single `address="AllIPs"` series to keep the metric's cardinality bounded.

So this is a docs bug rather than a code bug - the aggregation is the intended behavior. This PR takes option 2 from the issue and makes the documentation match what the exporter actually emits:

- `docs/03-Metrics/02-hubble_metrics.md` and `docs/03-Metrics/modes/basic.md`: label column now says `address` is always `AllIPs`, plus a short note in the basic-mode label values section explaining why individual remotes aren't exported.
- `docs/06-Troubleshooting/basic-metrics.md`: the sample scrape output still showed per-IP series with a `port` label that no longer exists. Replaced with the single aggregated series.
- `pkg/metrics/types.go`: the gauge's HELP string said "by remote address", which is the same misleading claim shipped to every Prometheus scrape. Reworded.
- `pkg/plugin/linuxutil/netstat_stats_linux.go`: added a note on the `addrDefaultTCPRemote` constant so the aggregation doesn't read like an unfinished implementation next time.

Worth calling out for the issue author: this closes the docs mismatch but does not give you the per-remote breakdown you were after. If there's appetite for that as an opt-in (default stays aggregated, a config flag turns on per-endpoint series with stale-series cleanup between polls), I'm happy to follow up with a separate PR - but that seemed like a decision for the maintainers rather than something to sneak into a docs fix.

## Related Issue

#2559

## Checklist

No behavior change, so nothing new to test - `go build ./pkg/metrics/... ./pkg/plugin/linuxutil/...` passes and the only runtime-visible difference is the HELP text on the gauge.


xRookieFight:

PR #2609, opened 2026-08-04
fix(utils): jitter Retry backoff so agents do not retry in lockstep

1991santhu:
## The problem

`utils.Retry` sleeps for exactly `2^attempt` seconds:

```go
t := int64(math.Pow(2, float64(i)))
time.Sleep(time.Duration(t) * time.Second)
```

The delay is a pure function of the attempt number, so every caller computes the same sequence — 1s, 2s, 4s, 8s.

That matters here more than it would in a single-process tool, because Retina runs as a **DaemonSet**: an agent on every node executes this loop. And the failures it retries are generally not node-local. The clearest case is the apiserver host lookup in `pkg/watchers/apiserver/apiserver.go`:

```go
// Retry the lookup for hostIPs in case of failure.
err = utils.Retry(retryFunc, hostLookupRetries)
```

If DNS resolution is degraded, it is degraded for the whole cluster, so every agent enters the loop at roughly the same moment and every agent then retries at exactly the same instants. The retries arrive as synchronised waves whose size scales with node count — and they land on the resolver precisely when it is already struggling, which is the load pattern backoff exists to prevent.

The same reasoning applies to the other two callers (`filtermanager`, `pktmon`), since both retry against node-local subsystems that tend to fail for cluster-wide reasons.

## The change

Equal jitter: keep half the computed interval as a floor, randomise the remainder.

```go
base := time.Duration(int64(math.Pow(2, float64(attempt)))) * time.Second
half := base / 2

return half + time.Duration(rand.Int63n(int64(base-half)+1))
```

Half is retained rather than using full jitter so a retry is never issued immediately after a failure. The total wait never exceeds what it was before, so no caller waits longer than it does today.

## Testing

`go test -count=2 ./pkg/utils/...` passes, `go vet` clean.

`pkg/utils` had no test file for `Retry`, so this adds one covering: the delay always falls within `[base/2, base]` across 500 samples for attempts 0–5; the delay varies across 500 draws, which fails if the implementation regresses to deterministic; and `Retry`'s existing contract — returns immediately on success, stops retrying once a call succeeds, and returns the last error when attempts are exhausted.

## Note

`math/rand` rather than `crypto/rand`, with a `//nolint:gosec` — this is load spreading, not a security boundary. Happy to switch if the project prefers otherwise.

### DIFF
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"math/rand"
 	"net/http"
 	"time"
 )
@@ -43,16 +44,33 @@ func StringPtr(v string) *string {
 }
 
 // Exponential backoff retry logic.
+//
+// The delay carries equal jitter: half the computed interval is retained as a
+// floor and the remainder is randomised. Retina runs as a DaemonSet, so an agent
+// on every node executes this loop. The conditions that make a call fail —
+// apiserver or DNS degradation — are typically cluster-wide, meaning every agent
+// enters the loop at once. Without jitter each of them would sleep for exactly
+// 1s, 2s, 4s and so on, so their retries would arrive at the dependency in
+// synchronised waves whose size scales with the node count, which is the load
+// pattern backoff exists to avoid.
 func Retry(f func() error, retry int) (err error) {
 	for i := 0; i < retry; i++ {
 		err = f()
 		if err == nil {
 			return nil
 		}
-		t := int64(math.Pow(2, float64(i)))
-		time.Sleep(time.Duration(t) * time.Second)
+		time.Sleep(backoffWithJitter(i))
 	}
 	return err
+}
+
+// backoffWithJitter returns the delay for the given zero-based attempt: an
+// exponential interval of 2^attempt seconds with the upper half randomised.
+func backoffWithJitter(attempt int) time.Duration {
+	base := time.Duration(int64(math.Pow(2, float64(attempt)))) * time.Second
+	half := base / 2
+
+	return half + time.Duration(rand.Int63n(int64(base-half)+1)) //nolint:gosec // spreading load, not a security boundary
 }
 
 func CompareStringSlice(a, b []string) bool {
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -45,14 +45,16 @@ func StringPtr(v string) *string {
 
 // Exponential backoff retry logic.
 //
-// The delay carries equal jitter: half the computed interval is retained as a
-// floor and the remainder is randomised. Retina runs as a DaemonSet, so an agent
-// on every node executes this loop. The conditions that make a call fail —
-// apiserver or DNS degradation — are typically cluster-wide, meaning every agent
-// enters the loop at once. Without jitter each of them would sleep for exactly
-// 1s, 2s, 4s and so on, so their retries would arrive at the dependency in
-// synchronised waves whose size scales with the node count, which is the load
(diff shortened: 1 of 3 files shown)


1991santhu:

1991santhu:
Correcting my own patch, and the reason is worth stating because I missed it first time round.

The first revision used equal jitter, giving `[base/2, base]`. That silently undercut a documented budget. `pkg/watchers/apiserver` declares:

```go
hostLookupRetries = 6 // 6 retries for a total of 63 seconds.
```

and 63 is exactly `1+2+4+8+16+32` — the sum of the deterministic schedule. Centring the jitter made the real total anywhere between roughly 31 and 63 seconds, so an agent could give up on resolving the apiserver about twice as fast as that comment promises, without the comment changing.

The jitter is now added rather than centred, giving `[base, 2*base]`:

```go
base := time.Duration(int64(math.Pow(2, float64(attempt)))) * time.Second

return base + time.Duration(rand.Int63n(int64(base)+1))
```

so no caller ever gives up sooner than the deterministic schedule would have. The documented total becomes a lower bound rather than an exact figure, and I have updated that comment to say so:

```go
hostLookupRetries = 6 // 6 retries, waiting at least 63 seconds in total (jittered, so up to ~126s).
```

I have also noted the convention on `Retry` itself, so the other two callers know that a documented total becomes a floor.

Decorrelation is unaffected — what matters is that agents differ from one another, not the direction.

`go test -count=2 ./pkg/utils/` passes. Note that `pkg/watchers/apiserver` and `pkg/watchers/endpoint` do not build on darwin (`fm.Init` and `initNewCache` undefined); I confirmed that reproduces on a clean tree without my changes, so it is unrelated.

1991santhu:
Adding a measurement, because "this is a real problem" is worth more as a number than as an argument.

I simulated the arrival times this backoff produces for a DaemonSet of 500 agents whose apiserver host lookups fail together — which is the case the retry exists for, since they fail together precisely because the thing they share became unreachable. The metric is the peak number of lookups landing inside any 100ms window, which is what the resolver actually feels.

Schedule is retina's own: `1s, 2s, 4s, 8s, 16s, 32s`.

```
attempt           none         equal         added
1                500.0         119.2          66.5
2                500.0          62.5          37.1
3                500.0          36.1          22.4
4                500.0          22.8          14.7
5                500.0          13.6           9.7
6                500.0           9.3           6.8
```

Without jitter every one of the 500 agents retries in the same 100ms window, at every attempt, indefinitely — the peak never decays, because a deterministic schedule cannot decorrelate anything. With the jitter in this PR the first retry peaks at ~67 and falls to ~7 by the sixth, a **7.5x reduction on the first retry and 73x by the last**.

The `equal` column is the equal-jitter variant I originally submitted. It also decorrelates, but less, because `[d/2, d]` is a narrower window than `[d, 2d]` — which is a second reason to add the jitter rather than centre it, on top of the documented-budget argument above.

One result worth calling out: I re-ran it with the agents' initial failures scattered over 50ms rather than landing on the same instant, and the deterministic column is unchanged at 500. A deterministic backoff neither creates nor destroys correlation — it *preserves* whatever correlation exists, forever. That is the mechanism, and it does not depend on the failures being perfectly simultaneous.

To be straight about what this is and is not: it is a simulation of the arrival process, not a production incident report. I have not observed this in a running cluster, and I am not claiming anyone has. It quantifies the mechanism the code implies. The harness is ~150 lines of dependency-free Go and I am happy to attach it if that is useful.

```
go run . -clients=500 -schedule=1s,2s,4s,8s,16s,32s -window=100ms
```

nddq:
PR is automated and low effort even if AI was used, closing...

PR #2625, opened 2026-08-07
feat: Add windows support and cli changes 

lakshk98:
This pr rebases dev/v0.0.33 onto main resolving conflicts across Dockerfiles, workflow files, Helm values, E2E tests, and auto-generated protobuf files.

Merge conflict resolutions

For most conflicts, upstream changes were taken as-is. Notable exceptions:

•  .github/workflows/e2e-test-event-writer.yml  — kept our version, which includes Windows-specific build steps (LLVM, MSBuild, eBPF store, Docker daemon) absent from upstream's skeleton.
•  deploy/.../values.yaml  — introduced  enablePodLevel_win: true  to decouple Windows pod-level metrics from the upstream Linux default ( enablePodLevel: false ), following the existing  enabledPlugin_linux / enabledPlugin_win  pattern. Also kept  enabledPlugin_win  with  ebpfwindows  and  enableAnnotations: true .
•  pkg/utils/metadata_*.pb.go  — regenerated from the auto-merged  .proto  (which combined our Cilium eBPF drop reasons with upstream's HNS/NDIS drop reasons and sampling fields).
•  test/e2e/retina_e2e_test.go  — kept Windows  UnLoadAndPinWinBPFJob  step; dropped stale  ValidateHubble  reference (renamed upstream); took upstream's  InstallAndTestHubbleMetrics  and  ValidateCapture .
•  controller/Dockerfile  /  cli/Dockerfile  — took upstream's  bpf-gen  stage refactor and Azure Linux 3.0 base image upgrade.

Post-merge fixes

The final commit removed usage of the hubble parser options addresses build failures introduced by the merge — upstream changes that conflicted with our usage of hubble parser options required removing that dependency and fixes lint errors surfaced by the linter

## Related Issue

## Checklist

## Additional Notes

---


deploy/standard/manifests/controller/helm/retina/values.yaml
```diff
@@ -53,8 +53,9 @@ image:
 enableConntrackMetrics: false
 conntrackReportInterval: 30s
 enablePodLevel: false
+enablePodLevel_win: true
 remoteContext: false
-enableAnnotations: false
+enableAnnotations: true
 bypassLookupIPOfInterest: false
```
alexcastilio, line 58: I don't think we want to change the default behavior of enableAnnotations to true

deploy/standard/manifests/controller/helm/retina/values.yaml
```diff
@@ -53,8 +53,9 @@ image:
 enableConntrackMetrics: false
 conntrackReportInterval: 30s
 enablePodLevel: false
+enablePodLevel_win: true
```
alexcastilio: `enablePodLevelWin` is in line with the naming pattern in the file

.github/workflows/e2e-test-event-writer.yml
```diff
@@ -32,4 +35,109 @@ jobs:
...
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        uses: actions/checkout@v4.2.2
+
+      - name: Azure Login
+        uses: azure/login@v2
+        if: ${{ env.IS_MERGE_GROUP == 'true' }}
+        with:
+          client-id: ${{ vars.AZURE_CLIENT_ID }}
+          tenant-id: ${{ vars.AZURE_TENANT_ID }}
+          subscription-id: ${{ vars.AZURE_SUBSCRIPTION_ID }}
```
alexcastilio:
```suggestion
          client-id: ${{ secrets.AZURE_CLIENT_ID }}
          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION }}
```

deploy/standard/manifests/controller/helm/retina/templates/configmap.yaml
```diff
@@ -53,11 +53,9 @@ data:
     metricsInterval: {{ .Values.metricsInterval }}
     metricsIntervalDuration: {{ .Values.metricsIntervalDuration }}
     enableTelemetry: {{ .Values.enableTelemetry }}
-    enablePodLevel: {{ .Values.enablePodLevel }}
+    enablePodLevel: {{ .Values.enablePodLevel_win }}
```
alexcastilio: `enablePodLevelWin` is in line with the naming pattern in the file

test/e2e/yaml/windows/non-hpc-pod.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: non-hpc-pod
+  labels:
+    app: non-hpc
+spec:
+  nodeSelector:
+    "kubernetes.io/os": windows
+  containers:
+  - name: non-hpc-container
+    image: ghcr.io/microsoft/retina/test/e2e-test-event-writer:latest
```
alexcastilio: image is coming from ghcr.io, but when it's running for merge queue, image is pushed to ACR. This should be set accordingly.
I also suggest not to use the latest tag, as it can be using some image different than expected. Make it configurable and set to same value that is pushed by build pipeline

test/e2e/yaml/windows/install-ebpf-xdp.yaml
```diff
@@ -0,0 +1,37 @@
...
+    matchLabels:
+      name: install-ebpf-xdp
+  template:
+    metadata:
+      labels:
+        name: install-ebpf-xdp
+    spec:
+      containers:
+      - name: install-ebpf-xdp-container
+        image: ghcr.io/microsoft/retina/test/e2e-test-event-writer:latest
```
alexcastilio: image is coming from ghcr.io, but when it's running for merge queue, image is pushed to ACR. This should be set accordingly.
I also suggest not to use the latest tag, as it can be using some image different than expected. Make it configurable and set to same value that is pushed by build pipeline

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,360 @@
...
+			} else if lostEventsCount > prevLostEventsCount {
+				// The lost events count is cumulative, so we need to calculate the difference
+				counterToAdd := lostEventsCount - prevLostEventsCount
+				metrics.LostEventsCounter.WithLabelValues(utils.Kernel, name).Add(float64(counterToAdd))
+				prevLostEventsCount = lostEventsCount
+			}
+
+		case <-ctx.Done():
+			p.l.Error("ebpfwindows plugin canceling", zap.Error(ctx.Err()))
+			err := eventsMap.UnregisterForCallback()
```
alexcastilio: Successful registration already defers unregister, but ctx.Done() unregisters again before returning. The return runs the defer, releasing the same native handle twice. 

Please keep one cleanup path.

test/e2e/framework/kubernetes/load-winbpf.go
```diff
@@ -0,0 +1,130 @@
...
+
+type LoadAndPinWinBPF struct {
+	KubeConfigFilePath                 string
+	LoadAndPinWinBPFDeamonSetNamespace string
+	LoadAndPinWinBPFDeamonSetName      string
+}
+
+func WaitForPodReadyWithTimeOut(ctx context.Context, kubeConfigFilePath, namespace, labelSelector string, timeout time.Duration) error {
+	config, _ := clientcmd.BuildConfigFromFlags("", kubeConfigFilePath)
+	clientset, _ := kubernetes.NewForConfig(config)
```
alexcastilio: both errors are discarded. This can cause nil dereference. Please handle both errors

alexcastilio (commented):
I added some comments.

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,355 @@
...
+		e, err := p.parser.Decode(&observer.MonitorEvent{
+			Payload: &observer.PerfEvent{
+				Data: perfData,
+			},
+		})
+		if err != nil {
+			return fmt.Errorf("could not convert pktmon dropnotify event to flow: %w", err)
+		}
+		ext := utils.NewExtensions()
+		utils.AddPacketSize(ext, size-uint32(unsafe.Sizeof(DropNotify{})))
```
alexcastilio: The packet header sizes are applied to the wrong event types. The ordinary Cilium drop branch subtracts the 57-byte pktmon header, while the pktmon branch still subtracts the 36-byte DropNotify header. This under-reports ordinary drop sizes and over-reports pktmon sizes by 21 bytes. Please use dropNotifyV1Len at line 282 and dropPktmonNotifyV1Len at line 337, and add packet-size assertions for both event types.

.github/workflows/e2e-test-event-writer.yml
```diff
@@ -33,3 +36,113 @@ jobs:
...
+        id: tag
+        shell: pwsh
+        run: |
+          $TAG = $(git tag --points-at HEAD)
+          if (-not $TAG) {
+            $TAG = $(git rev-parse --short HEAD)
+          }
+          echo "tag=$TAG" | Out-File -FilePath $env:GITHUB_OUTPUT -Encoding utf8 -Append
+
+      - name: Build and push images
```
alexcastilio, line 130: The fixtures now use the commit-specific image tag, but this image is published by a separate merge_group workflow from the E2E consumer. There is no dependency guaranteeing that this push completes before InstallEbpfXdp attempts to pull the tag, so merge-queue E2E can fail nondeterministically. Please orchestrate publication and E2E in one workflow or reusable-workflow dependency, or add a bounded registry-existence wait before E2E starts.

test/e2e/framework/kubernetes/load-winbpf.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	var windowsPod *v1.Pod
+	for i := range pods.Items {
+		pod := &pods.Items[i]
+		if pod.Spec.NodeSelector["kubernetes.io/os"] == "windows" &&
+			pod.Status.Phase == v1.PodRunning {
+			// Optionally, check for Ready condition here
+			windowsPod = pod
+			break
```
alexcastilio, line 121: This selects the first running Windows pod matching the label, independent of the node hosting the non-HPC traffic pod. On a multi-Windows-node cluster, the interface index can be obtained on one node and used by the event-writer or Retina pod on another. Please accept the target node name and select the event-writer and Retina pods on that same node

pkg/plugin/ebpfwindows/ebpf_windows_test.go
alexcastilio, line 1: The PR adds 24 Windows-only Go tests, but no required workflow runs go test ./pkg/plugin/ebpfwindows; Linux test discovery excludes this package. Please add a required Windows unit-test job so these tests can protect the new parser and lifecycle code.

pkg/plugin/ebpfwindows/ebpf_windows_test.go
```diff
@@ -0,0 +1,967 @@
...
+const (
+	pktSizeBytes = 100
+)
+
+var (
+	errTestFailure    = errors.New("test failure")
+	errModuleNotFound = errors.New("module not found")
+)
+
+func makeMockEthernetIPv4TCPPacket() []byte {
```
alexcastilio, line 63: The existing valid cases cover both Ethernet and raw-IP pktmon payloads, but all packet fixtures are IPv4/TCP. NewParser registers IPv4/IPv6 and TCP/UDP/SCTP/ICMP decoders, and decodeLayers maps each of those protocols into the flow model. Please add one focused IPv6 case and one UDP case to exercise those currently uncovered primary branches.

pkg/plugin/ebpfwindows/ebpf_windows_test.go
alexcastilio, line 1: These tests assert basic addresses, ports, event type, and selected reasons, but not the user-visible flow semantics that feed advanced metrics. Please also assert verdict, subtype/reason, observation point/direction, packet-size extension, and source/destination endpoint metadata. Packet-size assertions would have caught the swapped header constants in the current patch.

pkg/plugin/ebpfwindows/ebpf_windows_test.go
```diff
@@ -0,0 +1,967 @@
...
+	//nolint:gosec // ignore G115 -- data length is guaranteed to be within uint32 bounds in test context
+	err = p.handleTraceEvent(unsafe.Pointer(&data[0]), uint32(len(data)))
+	if err != nil {
+		t.Fatalf("expected no error for unknown event type, got: %v", err)
+	}
+}
+
+// TestHandleTraceEvent_InvalidTraceNotify tests the behavior of the handleTraceEvent function
+// when an invalid TraceNotify event is received.
+func TestHandleTraceEvent_InvalidTraceNotify(t *testing.T) {
```
alexcastilio, line 611: The current negative tests below cover only a few top-level events. Since this feature decodes packed external data through unsafe pointers, please add direct decoder cases for length-1, exact supported length, unsupported version/type, malformed packet payload, and invalid IP-version nibble, asserting an error and no panic.

pkg/plugin/ebpfwindows/ebpf_windows_test.go
```diff
@@ -0,0 +1,967 @@
...
+	if err == nil {
+		t.Fatalf("expected error - invalid size 0")
+	} else if err.Error() != "invalid size: 0" {
+		t.Fatalf("expected error - invalid size: 0, got %v", err)
+	}
+}
+
+// TestMetricsMapIterateCallback_DropEgress tests the behavior of the metricsMapIterateCallback function
+// when a drop event is received for egress traffic.
+func TestMetricsMapIterateCallback_DropEgress(t *testing.T) {
```
alexcastilio, line 746: These callback tests below obtain metric handles but do not assert that the callback wrote the expected Count and Bytes, so they can pass even if the metric updates are removed. Please assert exact values for ingress/egress forward/drop and include the pktmon extended-reason label.

pkg/plugin/hnsstats/hnsstats_windows.go
```diff
@@ -210,7 +212,21 @@ func notifyHnsStats(h *hnsstats, stats *HnsStatsData) {
...
+	ciliumEnabled, err := plugincommon.IsCiliumOnWindowsEnabled()
+	if err != nil {
+		h.l.Error("Error while checking if Cilium is enabled on Windows", zap.Error(err))
+		return fmt.Errorf("failed to check if Cilium is enabled on Windows: %w", err)
+	}
+
+	if ciliumEnabled {
+		h.l.Warn("Cilium is enabled on Windows, skipping hnsstats plugin initialization")
+		return nil
+	}
```
alexcastilio: The new registry value controls mutually exclusive ebpfwindows and hnsstats startup, but there is no missing/0/1/error test matrix. Please add injectable registry tests proving both plugins make complementary start/skip decisions

pkg/plugin/ebpfwindows/ebpf_windows.go
```diff
@@ -0,0 +1,355 @@
...
+			return fmt.Errorf("%w", errNilDropNotifyEvent)
+		}
+		if fl.GetIP() == nil {
+			return fmt.Errorf("%w; perfdata: %v;", errNilDropNotifyEvent, perfData)
+		}
+		// Set the drop reason.
+		eventType := fl.GetEventType().GetSubType()
+		utils.AddDropReason(fl, ext, uint16(eventType)) //nolint:gosec // eventType values are bounded drop reason codes that fit in uint16
+		utils.SetExtensions(fl, ext)
+		p.enricher.Write(e)
```
alexcastilio: These writes send every Windows event into the shared advanced-metrics pipeline, but the expected admission behavior differs by context mode (see [doc](https://retina.sh/docs/Metrics/modes/). Existing Linux producers use the selected-IP filter to bound local-context events to observed pods; remote context then intentionally builds source/destination-pair series, including external IPs. Please add mode-aware admission before these handoffs: in local context reject flows when neither IP is selected and allow only the observed local pod to become the metric identity; in remote context preserve both sides so the common module can generate pair labels. The Windows filter manager currently discards AddIPs, so it must retain the selected set or the native API must expose equivalent filtering. Add some tests for local selected-source, selected-destination, neither-selected and remote bypass

pkg/plugin/ebpfwindows/parser_windows.go
```diff
@@ -0,0 +1,663 @@
...
+	if dropReason > math.MaxInt32 {
+		return fmt.Errorf("%w: %d", errDropReasonOverflow, dropReason)
+	}
+	decoded.DropReasonDesc = pb.DropReason(int32(dropReason))
+	decoded.File = decodeFileInfo(dn)
+	decoded.Ethernet = decodedpacket.Ethernet
+	decoded.IP = decodedpacket.IP
+	decoded.L4 = decodedpacket.L4
+	decoded.Source = srcEndpoint
+	decoded.Destination = dstEndpoint
```
alexcastilio: The parser assigns non-nil endpoint objects to both sides before Kubernetes enrichment, even when an IP is external or otherwise unresolved. Existing Linux producers leave endpoints nil; the enricher populates only known objects. That distinction is required in both modes: local context uses non-nil endpoints to choose the local pod, while remote context retains the raw IP separately and should emit unknown Kubernetes metadata for an unresolved peer. Please keep the parser mode-neutral by leaving Source/Destination nil until enrichment.

alexcastilio (changes requested):
Added some comments regarding metric modes and labels

pkg/plugin/ebpfwindows/ebpf_windows.go
lakshk98: swapped size and modified existing test to validate size.

test/e2e/framework/kubernetes/load-winbpf.go
lakshk98: ebpf loading and pinning is now done on all available windows nodes. Metric validation commands are targeted to the node with non-hpc-pod on it.

deploy/standard/manifests/controller/helm/retina/values.yaml
```diff
@@ -53,6 +53,7 @@ image:
 enableConntrackMetrics: false
 conntrackReportInterval: 30s
 enablePodLevel: false
+enablePodLevelWin: true
```
alexcastilio: this should be false by default

deploy/standard/manifests/controller/helm/retina/values.yaml
```diff
@@ -96,7 +97,7 @@ apiServer:
 logLevel: debug
 
 enabledPlugin_linux: '["dropreason","packetforward","linuxutil","dns"]'
-enabledPlugin_win: '["hnsstats"]'
+enabledPlugin_win: '["hnsstats", "ebpfwindows"]'
```
alexcastilio: I think we don't want to enable bpfwindows by default at this moment. Please remove the plugin from this list.

alexcastilio (dismissed):

lakshk98:
@microsoft-github-policy-service agree company="Microsoft"

ibezrukavyi (approved):

PR #2679, opened 2026-08-23
fix(metrics): emit a port value for non-TCP/UDP flows to keep cardinality aligned

### LINKED ISSUE #2678: A port metric context panics the metrics module on any non-TCP/UDP flow (ICMP, SCTP)
With a `port` metric context configured (`sourceContext`/`destinationContext` including `port`), the metric vector is registered with a `port` label, but the value side does not emit a value for a flow whose L4 is neither TCP nor UDP, so the metrics module panics.

`getLabels()` adds the label whenever `Port` is set (`pkg/module/metrics/types.go:175`):

```go
if c.Port {
	labels = append(labels, prefix+portCtxOption)
}
```

`getByDirectionValues()` only appends a value for TCP, UDP, or a nil L4 (`pkg/module/metrics/types.go:301`):

```go
if c.Port {
	l4 := f.GetL4()
	if l4 != nil {
		if tcp := l4.GetTCP(); tcp != nil {
			... // one value
		} else if udp := l4.GetUDP(); udp != nil {
			... // one value
		}
		// no else: a non-nil L4 that is not TCP/UDP appends nothing
	} else {
		values = append(values, "unknown")
	}
}
```

So an ICMP, ICMPv6, or SCTP flow (a non-nil `Layer4` whose `GetTCP()` and `GetUDP()` are both nil) yields one fewer value than the label set, and the next `WithLabelValues(labels...)` in `drops.go` / `forward.go` panics:

```
panic: inconsistent label cardinality: expected N label values but got N-1
```

taking down the metrics goroutine. A single ping is enough once `port` is enabled.

**How to reproduce**

Enable a `port` context and send ICMP traffic through a watched pod. As a unit test:

```go
c := NewCtxOption([]string{"ip", "port"}, source)
f := &flow.Flow{
	IP: &flow.IP{Source: "10.0.0.1"},
	L4: &flow.Layer4{Protocol: &flow.Layer4_ICMPv4{ICMPv4: &flow.ICMPv4{}}},
}
// len(c.getLabels()) == 2, len(c.getValues(f)) == 1  ->  WithLabelValues panics
```

**Expected**

The value count matches the label count for every flow, as it already does for a nil L4 with the `"unknown"` placeholder.

Same class as #561 and #838. `port` is opt-in, so this only affects operators who enable it.

thc1006:
`getLabels()` adds the `port` label whenever the `port` context is set, but `getByDirectionValues()` only emitted a value for TCP, UDP, or a nil L4. A non-nil `Layer4` that is neither TCP nor UDP (ICMP, ICMPv6, SCTP) appended nothing, so the value slice was one short of the label slice and the next `WithLabelValues` panicked the metrics module with `inconsistent label cardinality`. A single ICMP flow triggers it once `port` is enabled.

This adds the missing `else`, emitting the same `"unknown"` placeholder the nil-L4 branch already uses, so the counts stay aligned.

Verified: a new `TestNewCtxOps` case with an ICMP flow, plus `TestPortContextCardinalityAcrossL4` which brute-forces every `Layer4` variant (TCP, UDP, SCTP, ICMPv4, ICMPv6, nil) in both directions, fail without the fix (`getValues` returns one fewer value than `getLabels`) and pass with it. `gofmt` and `go vet ./pkg/module/metrics/` are clean.

Fixes #2678

### DIFF
--- a/pkg/module/metrics/types.go
+++ b/pkg/module/metrics/types.go
@@ -314,6 +314,9 @@ func (c *ContextOptions) getByDirectionValues(f *flow.Flow, dest bool) []string 
 				} else {
 					values = append(values, fmt.Sprintf("%d", udp.GetSourcePort()))
 				}
+			} else {
+				// Non-TCP/UDP L4 (e.g. ICMP) has no port; keep values aligned with labels.
+				values = append(values, "unknown")
 			}
 		} else {
 			values = append(values, "unknown")
--- a/pkg/module/metrics/types_test.go
+++ b/pkg/module/metrics/types_test.go
@@ -159,6 +159,17 @@ func TestNewCtxOps(t *testing.T) {
 			f:            &flow.Flow{},
 			expectedVals: []string{"unknown"},
 		},
+		{
+			name:     "port option with a non-TCP/UDP flow keeps values aligned with labels",
+			opts:     []string{"ip", "port"},
+			ctxType:  source,
+			expected: []string{"source_ip", "source_port"},
+			f: &flow.Flow{
+				IP: &flow.IP{Source: "10.0.0.1"},
+				L4: &flow.Layer4{Protocol: &flow.Layer4_ICMPv4{ICMPv4: &flow.ICMPv4{}}},
+			},
+			expectedVals: []string{"10.0.0.1", "unknown"},
+		},
 	}
 
 	for _, tc := range tt {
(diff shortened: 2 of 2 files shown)


thc1006:

nddq:
@thc1006 fix looks right to me, please sign your commits so we can get this in.

nddq (approved):
thanks!

PR #2688, opened 2026-08-25
fix(build): bpf generation improvements

alexcastilio:
# Description

This PR hardens BPF object generation during Retina image builds, particularly for ARM64 builds running through QEMU.

The BPF generation stage could fail while compiling its temporary Go generator, leave an empty architecture-specific object file, and still allow the image build to continue. The resulting image then failed at runtime while loading the embedded BPF object.

## What changed

- Make the BPF generation stage fail immediately when a command fails.
- Verify that the generated architecture-specific conntrack object exists and is non-empty before packaging the generated plugin artifacts.
- Stop the `retina-image` target if either the init or agent image build fails, preventing a later loop iteration from masking an earlier failure.
- Disable CGO only in the temporary BPF generator stage and enable Microsoft' no-CGO OpenSSL system-crypto backend. This avoids QEMU-emulated GCC failures during ARM64 generation while preserving system-crypto compliance.

These changes affect build-time generators only. CGO and system-crypto behavior for the Retina runtime binaries is unchanged.

## Related Issue

No linked issue.

## Checklist

- Built the exact `bpf-gen` stage for Linux AMD64.
- Built the exact `bpf-gen` stage for Linux ARM64 through QEMU.
- Confirmed both builds produce a non-empty architecture-specific conntrack BPF object.

## Additional Notes

The no-CGO configuration is intentionally scoped to the temporary generator stage in `controller/Dockerfile`; it is not a repository-wide or runtime build setting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -261,6 +261,7 @@ container-docker-windows: # util target to build Windows container images withou
 
 retina-image: ## build the retina linux container image.
 	echo "Building for $(PLATFORM)"
+	set -e; \
 	for target in $(AGENT_TARGETS); do \
 		echo "Building for $$target"; \
 		if [ "$$target" = "init" ]; then \
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -25,6 +25,8 @@ ARG GOOS=linux
 ARG GOARCH=amd64
 ENV GOOS=${GOOS}
 ENV GOARCH=${GOARCH}
+ENV CGO_ENABLED=0
+ENV GOEXPERIMENT=ms_nocgo_opensslcrypto
 RUN if [ "$GOOS" = "linux" ] ; then \
     tdnf install -y clang lld bpftool libbpf-devel; \
     fi
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -30,9 +30,12 @@ RUN if [ "$GOOS" = "linux" ] ; then \
     fi
 COPY ./pkg/plugin /go/src/github.com/microsoft/retina/pkg/plugin
 WORKDIR /go/src/github.com/microsoft/retina
-RUN if [ "$GOOS" = "linux" ] ; then \
+RUN set -eu; \
+    if [ "$GOOS" = "linux" ] ; then \
     go mod init github.com/microsoft/retina; \
     go generate -skip "mockgen" -x /go/src/github.com/microsoft/retina/pkg/plugin/...; \
+    if [ "$GOARCH" = "amd64" ]; then bpf_arch=x86; else bpf_arch="$GOARCH"; fi; \
+    test -s "./pkg/plugin/conntrack/conntrack_bpfel_${bpf_arch}.o"; \
     tar czf /gen.tar.gz ./pkg/plugin; \
     rm go.mod; \
     else \
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -32,12 +32,16 @@ RUN if [ "$GOOS" = "linux" ] ; then \
     fi
 COPY ./pkg/plugin /go/src/github.com/microsoft/retina/pkg/plugin
 WORKDIR /go/src/github.com/microsoft/retina
+# Sanity-check every arch-specific BPF object bpf2go produced is non-empty.
 RUN set -eu; \
     if [ "$GOOS" = "linux" ] ; then \
     go mod init github.com/microsoft/retina; \
     go generate -skip "mockgen" -x /go/src/github.com/microsoft/retina/pkg/plugin/...; \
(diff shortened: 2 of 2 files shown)


controller/Dockerfile
```diff
@@ -25,14 +25,19 @@ ARG GOOS=linux
 ARG GOARCH=amd64
 ENV GOOS=${GOOS}
 ENV GOARCH=${GOARCH}
+ENV CGO_ENABLED=0
+ENV GOEXPERIMENT=ms_nocgo_opensslcrypto
```
SRodi, line 29: ENV instructions persist across multi-stage, I think we should do this inline (see comment below)

controller/Dockerfile
```diff
@@ -25,14 +25,19 @@ ARG GOOS=linux
...
 RUN if [ "$GOOS" = "linux" ] ; then \
     tdnf install -y clang lld bpftool libbpf-devel; \
     fi
 COPY ./pkg/plugin /go/src/github.com/microsoft/retina/pkg/plugin
 WORKDIR /go/src/github.com/microsoft/retina
-RUN if [ "$GOOS" = "linux" ] ; then \
+RUN set -eu; \
+    if [ "$GOOS" = "linux" ] ; then \
     go mod init github.com/microsoft/retina; \
     go generate -skip "mockgen" -x /go/src/github.com/microsoft/retina/pkg/plugin/...; \
```
SRodi, line 39: Scope no-cgo OpenSSL to the generator only; QEMU-emulated gcc for arm64 fails under cgo. Runtime binaries (capture-bin, controller-bin, init-bin) must not inherit this, that's a separate policy change.

```suggestion
    CGO_ENABLED=0 GOEXPERIMENT=ms_nocgo_opensslcrypto \
         go generate -skip "mockgen" -x /go/src/github.com/microsoft/retina/pkg/plugin/...; \
     # Sanity-check every arch-specific BPF object bpf2go produced is non-empty.
     for f in $(find ./pkg/plugin -name '*_bpfel_*.o'); do test -s "$f" || \
         { echo "empty BPF object: $f" >&2; exit 1; }; done; \
```

controller/Dockerfile
```diff
@@ -25,14 +25,19 @@ ARG GOOS=linux
...
+    if [ "$GOARCH" = "amd64" ]; then bpf_arch=x86; else bpf_arch="$GOARCH"; fi; \
+    test -s "./pkg/plugin/conntrack/conntrack_bpfel_${bpf_arch}.o"; \
```
SRodi: Sanity-check every arch-specific BPF object bpf2go produced is non-empty.

```suggestion
    for f in $(find ./pkg/plugin -name '*_bpfel_*.o'); do test -s "$f" || \
         { echo "empty BPF object: $f" >&2; exit 1; }; done; \
```

controller/Dockerfile
jpayne3506, line 29: Not sure of the timeline/priority, but 1.27 removes the need/use of GOEXPERIMENT.

Makefile
```diff
@@ -261,6 +261,7 @@ container-docker-windows: # util target to build Windows container images withou
 
 retina-image: ## build the retina linux container image.
 	echo "Building for $(PLATFORM)"
+	set -e; \
```
jpayne3506, line 264: Is this for testing? Imo, this should be set exclusively in CI/CD or at least be optional, with the default having it not be used.

alexcastilio: This is used to build (both locally and in pipeliens). `set -e` will make the build fail if there's any error, instead of silently continue

controller/Dockerfile
alexcastilio: when ENV is set after FROM, it will only persist through child stages (e.g. `FROM bpf-gen AS ...`) which does not exist in this file

controller/Dockerfile
alexcastilio: Thanks for the suggestion. The sanity-check idea for every arch-specific bpf object is good, but the problem with the suggested implementation. It would fail for empty placeholders for other architectures, causing build to fail. And if find returns no file, loop would not execute and succeed. 
I have applied the suggestion fixing these points mentioned, so it will check only the right arch and fail if there's any empty file for that arch.

controller/Dockerfile
alexcastilio: ENV is already scoped (see comment above)

controller/Dockerfile
alexcastilio: I'll update to 1.27 in a separate PR and remove GOEXPERIMENT

SRodi (approved):
LGTM

PR #2707, opened 2026-09-01
fix(operator): set Pod owner reference on RetinaEndpoints for garbage collection

### LINKED ISSUE #2708: operator: does not clean up stale RetinaEndpoints across restarts
**Describe the bug**

`RetinaEndpoint` objects can leak across missed events in the operator, called out in https://github.com/microsoft/retina/blob/0a8044757e75e4104a0ca4c6ea25821854505320/pkg/controllers/operator/retinaendpoint/retinaendpoint_controller.go#L47.

**To Reproduce**

Stop operator pod, delete some pods, restart operator pod.  Notice that the `RetinaEndpoints` for the deleted pods are not cleaned up.

**Expected behavior**

Any `RetinaEndpoint` objects no longer exist for deleted pods.

**Additional context**

Fixed by https://github.com/microsoft/retina/pull/2707.

mmckeen:
# Description

`RetinaEndpoint` objects can be leaked across operator restarts or missed events right now.  This sets up owner references so they are automatically GCed when the pods they track are deleted.

## Related Issue

Solves the TODO referenced here https://github.com/microsoft/retina/blob/0a8044757e75e4104a0ca4c6ea25821854505320/pkg/controllers/operator/retinaendpoint/retinaendpoint_controller.go#L47

## Checklist

## Additional Notes

Alternatively, explored a periodic reconciliation loop that would also help create RetinaEndpoints that were missed.  This is much more expensive, and we already do that on operator restarts so this approach is far superior.

---

### DIFF
--- a/operator/cmd/standard/deployment.go
+++ b/operator/cmd/standard/deployment.go
@@ -209,6 +209,11 @@ func (o *Operator) Start() error {
 			// start reconcile the cached Pod before manager starts to not miss any events
 			go ke.ReconcilePod(ctrlCtx)
 
+			// Clean up RetinaEndpoints orphaned before owner references were set on them.
+			if err = mgr.Add(ke); err != nil {
+				return errors.Wrap(err, "unable to add retinaendpoint cleanup")
+			}
+
 			pc := podcontroller.New(mgr.GetClient(), mgr.GetScheme(), retinaendpointchannel)
 			if err = (pc).SetupWithManager(mgr); err != nil {
 				return errors.Wrap(err, "unable to create controller - podcontroller")
--- a/pkg/controllers/operator/retinaendpoint/retinaendpoint_controller.go
+++ b/pkg/controllers/operator/retinaendpoint/retinaendpoint_controller.go
@@ -15,6 +15,7 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	retinav1alpha1 "github.com/microsoft/retina/crd/api/v1alpha1"
 	"github.com/microsoft/retina/operator/cache"
@@ -44,8 +45,6 @@ func New(client client.Client, podchannel chan cache.PodCacheObject) *RetinaEndp
 	}
 }
 
-// NOTE(mainrerd): Chances are that pod cache channel lost pods events during controller manager restart, we need to
-// have full-set reconciliation to make sure all RetinaEndpoints are reconciled to Pods.
 // when a pod reaches here, it indicates that there is a metricsconfiguration that references it,
 // or doesn't and a RetinaEndpoint needs to be created or updated.
 // This is a blocking function, and will wait on the pod channel until a pod is received.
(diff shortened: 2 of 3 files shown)


pkg/controllers/operator/retinaendpoint/retinaendpoint_controller_test.go
```diff
@@ -247,6 +247,123 @@ func TestRetinaEndpointReconciler_ReconcilePod(t *testing.T) {
...
+
+	require.NoError(t, r.reconcileRetinaEndpointFromPod(context.Background(), cache.PodCacheObject{Key: key, Pod: pod}))
+
+	got := &retinav1alpha1.RetinaEndpoint{}
+	require.NoError(t, c.Get(context.Background(), key, got))
+	// requireOwnedByPod asserts exactly one Pod reference, so the stale UID is gone.
+	requireOwnedByPod(t, got, key.Name, podUID)
+}
+
+func TestRetinaEndpointReconciler_SetPodOwnerRejectsCrossNamespace(t *testing.T) {
```
nddq, line 357: nit: this is testing the controller-runtime's correctness, not retina operator's code, we can remove it

pkg/controllers/operator/retinaendpoint/retinaendpoint_controller.go
```diff
@@ -74,6 +75,26 @@ func (r *RetinaEndpointReconciler) ReconcilePod(pctx context.Context) {
...
+// drops events.
+//
+// A plain owner reference is enough for garbage collection. The controller reference is
+// deliberately not claimed, which would also block Pod deletion on this RetinaEndpoint.
+// A reference to a previous Pod of the same name is replaced in place, since owner
+// references are matched on group, kind and name.
+//
+// Note this is metadata.ownerReferences, distinct from the RetinaEndpoint's informational
+// spec.ownerReferences, which mirrors the Pod's own owners (ReplicaSet, DaemonSet, ...)
+// and is not consulted by garbage collection.
```
nddq, line 90: we can cut this down a bit by removing the parts about the previous bug/behavior (the delete event getting lost across restarts, the channel dropping events), a new reader shouldn't need that context to understand what this function does now. We can also remove the old NOTE(mainrerd) at line 48 since this change addresses it.

pkg/controllers/operator/retinaendpoint/retinaendpoint_controller.go
```diff
@@ -212,6 +238,12 @@ func (r *RetinaEndpointReconciler) reconcileRetinaEndpointFromPod(ctx context.Co
 		Annotations:     pod.Pod.Annotations,
 		OwnerReferences: refs,
 	}
+	// Adopt RetinaEndpoints created before the owner reference was set, so they become
```
nddq, line 241: same here, this reads as migration context. Something like "Set the Pod owner reference on existing RetinaEndpoints as well" is enough.

pkg/controllers/operator/retinaendpoint/retinaendpoint_controller.go
```diff
@@ -212,6 +238,12 @@ func (r *RetinaEndpointReconciler) reconcileRetinaEndpointFromPod(ctx context.Co
...
+	// garbage collectable without waiting for their Pod to be recreated.
+	if err = r.setPodOwner(new, pod.Pod); err != nil {
```
nddq, line 243: RetinaEndpoints that were already orphaned before this ships (pod gone, no ref) never get adopted since no pod event will ever come for them. We should do an one-time sweep on startup so the TODO is fully closed.

nddq (commented):
lgtm with some comments

mmckeen:
@nddq this has been updated 🙇

mmckeen:
@nddq apologies for the ping, but appreciate getting this merged if possible 🙇

nddq:
@mmckeen can you rebase with latest main? Thanks!

mmckeen:
> Matthew McKeen (@mmckeen) can you rebase with latest main? Thanks!

Done 👍

pkg/controllers/operator/retinaendpoint/retinaendpoint_controller.go
```diff
@@ -74,6 +73,67 @@ func (r *RetinaEndpointReconciler) ReconcilePod(pctx context.Context) {
...
+// Start deletes RetinaEndpoints whose Pod no longer exists, once, when the operator
+// starts. Owner references make the API server garbage collect RetinaEndpoints from then
+// on, but ones already orphaned carry no reference and would never receive a Pod event, so
+// nothing else would reclaim them.
+//
+// Start implements manager.Runnable so the manager runs it after its caches have synced.
+func (r *RetinaEndpointReconciler) Start(ctx context.Context) error {
+	endpoints := &retinav1alpha1.RetinaEndpointList{}
+	if err := r.List(ctx, endpoints); err != nil {
+		return fmt.Errorf("listing RetinaEndpoints: %w", err)
```
nddq: Ran this on a kind cluster with the CRD not yet installed (`installCRDs: true` with nothing pre-applied): the `List` here runs before the apiserver serves the new CRD, `Start` returns the error and the manager takes the whole operator down with it:

```
failed to start retina-operator: problem running manager: listing RetinaEndpoints: no matches for kind "RetinaEndpoint" in version "retina.sh/v1alpha1"
```

The operator on `main` survives the same start since the `Get` errors in `reconcileRetinaEndpointFromPod` just get requeued. The next restart works since the CRD exists by then, and helm pre-installs `crds/`, so this only affects installs that rely on `installCRDs` or a recreated CRD, but it is still a new crash path. Since this is a best-effort cleanup, we should log and swallow the errors in here instead of returning them (same for the `Delete` at line 128), the next restart sweeps again anyway.

pkg/controllers/operator/retinaendpoint/retinaendpoint_controller.go
```diff
@@ -74,6 +73,67 @@ func (r *RetinaEndpointReconciler) ReconcilePod(pctx context.Context) {
...
+			// The Pod is still around, so its next event adopts the RetinaEndpoint.
+			continue
+		}
+		if !errors.IsNotFound(err) {
+			return fmt.Errorf("getting Pod %s: %w", key.String(), err)
+		}
+
+		r.l.Info("deleting orphaned RetinaEndpoint", zap.String("name", key.String()))
+		// Refuse the delete if the RetinaEndpoint changed since it was listed.
+		if err := r.Delete(ctx, endpoint, client.Preconditions{UID: &endpoint.UID}); err != nil {
```
nddq: `Preconditions{UID}` only refuses the delete if the RetinaEndpoint was recreated under the same name, not if it changed, so the comment above doesn't match. Adding the resource version makes the `IsConflict` branch below actually reachable:

```suggestion
		if err := r.Delete(ctx, endpoint, client.Preconditions{UID: &endpoint.UID, ResourceVersion: &endpoint.ResourceVersion}); err != nil {
```

operator/cmd/standard/deployment.go
```diff
@@ -209,6 +209,11 @@ func (o *Operator) Start() error {
 			// start reconcile the cached Pod before manager starts to not miss any events
 			go ke.ReconcilePod(ctrlCtx)
 
+			// Clean up RetinaEndpoints orphaned before owner references were set on them.
+			if err = mgr.Add(ke); err != nil {
```
nddq: nit: `Start` is a bit misleading on the reconciler since `ReconcilePod` is the actual loop, we can keep the reconciler out of `manager.Runnable` with something like `mgr.Add(manager.RunnableFunc(ke.DeleteOrphanedEndpoints))`

pkg/controllers/operator/retinaendpoint/retinaendpoint_controller.go
```diff
@@ -74,6 +73,67 @@ func (r *RetinaEndpointReconciler) ReconcilePod(pctx context.Context) {
...
+	return nil
+}
+
+// Start deletes RetinaEndpoints whose Pod no longer exists, once, when the operator
+// starts. Owner references make the API server garbage collect RetinaEndpoints from then
+// on, but ones already orphaned carry no reference and would never receive a Pod event, so
+// nothing else would reclaim them.
+//
+// Start implements manager.Runnable so the manager runs it after its caches have synced.
+func (r *RetinaEndpointReconciler) Start(ctx context.Context) error {
```
nddq: nit: `Start` is a bit misleading on the reconciler since `ReconcilePod` is the actual loop, we can keep the reconciler out of `manager.Runnable` with something like `mgr.Add(manager.RunnableFunc(ke.DeleteOrphanedEndpoints))`

nddq (changes requested):
Thanks for the update! The sweep works on my kind cluster (orphan deleted, unowned endpoint with a live pod adopted, owned one kept, second start reports 0). A few more things that I've found.

mmckeen:
@nddq fixed 👍 🙇

nddq (approved):

PR #2746, opened 2026-09-08
fix(ci): run the ADO CI pipeline on the OneBranch pool

nddq:
# Description

The ADO pipeline `CI (Build Images)` runs on the Microsoft-hosted `Azure Pipelines` pool. That pool queues jobs for long periods, and merge-queue runs time out. This PR moves the pipeline to the OneBranch pool.

`.azure-pipelines/ci.yml` now extends the OneBranch NonOfficial template. Each job declares a OneBranch pool type, and the template places it on the OneBranch pool. Triggers and the PR-mode / merge-queue-mode split are unchanged.

- **Linux images**: one OneBranch docker job per component and architecture, from the new template `.azure-pipelines/jobs/onebranch-linux-image.yml`. arm64 images build on native arm64 hosts, so the QEMU and disk-space steps are gone.
- **Windows images**: Windows docker jobs. The archives are not compressed: the gzip pass alone took 20 minutes for the 5 GB agent image. The mandatory image save takes about 5 minutes on one host family of the pool and up to 17 minutes on the other. The pool provider picks the family, and no demand selects it.
- **Event writer**: a `type: windows` job builds the solution in the OneBranch container with Visual Studio 2022. NuGet restores go through the `microsoft_PublicPackages` feed, because network isolation blocks nuget.org. A Windows docker job then builds the image. The `latest` tag is gone; nothing reads it.
- **Go jobs**: Windows binaries, e2e, and perf run as `type: linux` jobs. The container ships Go 1.26, `az`, `make`, and `oras`, so this pipeline does not use `steps/setup-go.yml`. The e2e job installs `kubectl` for the capture scenario.
- **Manifests**: the e2e job runs `make manifest` with `hack/scripts/oras-index.sh` as the backend, because the container has no docker CLI. The new `MANIFEST_CREATE` variable defaults to `docker buildx imagetools create -t`, so the GitHub workflows and the release path do not change. The image and platform mapping stays in the Makefile only. In PR builds the e2e job is skipped; its check is not required on main.
- **Tag**: `<short SHA>-<build id>`, resolved at compile time, because docker jobs accept no script steps. OneBranch refuses to push a tag that already exists in the registry. The build id keeps tags unique per run, and the image jobs append the job attempt to the tags they push, so a retried job does not collide with its own earlier attempt. The e2e job reads the attempt tags of each platform from the registry's tag list and uses the highest one. A listing error fails the job, so a registry error cannot select an older attempt.
- **Check names**: GitHub names an ADO check `<pipeline> (<stage displayName> <job displayName>)`, and OneBranch makes the pipeline multi-stage. The stage and job display names compose to the names of the required checks on main, for example `CI (Build Images) (Build Agent Images - Linux amd64)`.
- **`runPerfTests`**: a new parameter. A manual run with it and `runE2EValidation` runs the perf jobs after e2e. Before, only merge-queue runs reached them.

## Related Issue

N/A — hosted-pool queue times in the ADO project.

## Checklist

- [ ] I have tested the changes locally. (N/A — pipeline-only change; the runs below are the test.)
- [ ] I have updated the documentation, if necessary. (N/A)
- [ ] I have added tests, if applicable. (N/A — pipeline-only change)

Full runs with `runE2EValidation` on this branch, all green:

| Run | Result |
| --- | --- |
| [125671](https://github-private.visualstudio.com/microsoft/_build/results?buildId=125671&view=results) | All images pushed, indexes created, e2e passed in 56 minutes at the first attempt. |
| [125726](https://github-private.visualstudio.com/microsoft/_build/results?buildId=125726&view=results) | Same, plus `runPerfTests`: e2e passed at the first attempt, then perf basic and perf advanced passed. |
| [125749](https://github-private.visualstudio.com/microsoft/_build/results?buildId=125749&view=results) | Same, with the review fixes: e2e and perf jobs gated on publish mode, guarded digest resolution in the manifest script. E2e passed in 48 minutes at the first attempt. |
| [125886](https://github-private.visualstudio.com/microsoft/_build/results?buildId=125886&view=results) | Same, with the attempt-suffixed tags and `make manifest` as the index backend. E2e passed at the first attempt. A retry of the build stage then pushed `-2` tags for every image without a collision. The retried jobs failed later at the template's drop publish, because they had already published in attempt 1; a job that fails before its publish does not hit that. |

PR builds pass in build-only mode in about 17 minutes, for example [125748](https://github-private.visualstudio.com/microsoft/_build/results?buildId=125748&view=results). The e2e and perf jobs are skipped there. The first PR build started on the OneBranch pool 50 seconds after it queued. The final head removes only the Windows host demand, which the runs above showed has no effect on placement.

## Additional Notes

- The pipeline is now authorized on the ACR docker-registry service connection, which the OneBranch image task uses for pushes.
- `go mod download` in the Dockerfiles and the `etl2pcapng.exe` download in `Dockerfile.windows-2022` reach public endpoints. OneBranch network telemetry records those. A follow-up routes Go modules through the internal proxy with a `GOPROXY` build arg.
- A `Cache@2` step for the Go caches was measured and dropped. The restore took longer than the compile it saved.
- `perf.yml` and `scale.yml` stay on the hosted pool until this pipeline is stable.
- Both Windows images are 5 GB, almost all of it the `servercore` base. Both run as HostProcess pods, so a follow-up can move them to a small base and cut the save and the e2e image pull.


.azure-pipelines/ci.yml
```diff
@@ -38,581 +49,446 @@ parameters:
...
-  - name: isMergeQueue
-    value: $[ startsWith(variables['Build.SourceBranch'], 'refs/heads/gh-readonly-queue/') ]
+  # Image tag: short commit SHA plus the build id. The SHA is resolved at
+  # compile time so that docker jobs can use it without a script step
+  # (OneBranch docker jobs accept only the image tasks). The build id keeps
+  # tags unique per run: OneBranch refuses to push a tag that already exists
+  # in the registry, so a rerun of the same commit, or the GitHub workflow
+  # pushing the same commit, must not share a tag with this run.
+  - name: TAG
+    value: ${{ substring(variables['Build.SourceVersion'], 0, 8) }}-$(Build.BuildId)
```
alexcastilio, line 76: Have you tried retrying some job execution after at least one image has been pushed? `Build.BuildId` remains unchanged when job is retried, so this task will attempt to push to same tag again. If OneBranch rejects an existing check, could we check whether the manifest already exists before running the image build/push and treat a matching existing image as success? Alternatively we would need some idempotent/overwrite implementation. Otherwise if one image fails to push (e.g. retina agent linux amd64) due to some transient issue, (e.g. network), the job that pushes both retina agent and init is marked as failed. Re-trying this job would also retry to push retina init, which would be rejected due to tag already have been pushed.

.azure-pipelines/ci.yml
```diff
@@ -38,581 +49,446 @@ parameters:
...
+            pool:
+              type: linux
+            variables:
+              ob_outputDirectory: $(Build.SourcesDirectory)/out
+              CLUSTER_NAME: retina-e2e-$(Build.BuildId)-$(System.JobAttempt)
+            steps:
+              # oras creates the multi-arch indexes from the per-platform tags
+              # that the image jobs pushed. The OneBranch Linux container has no
+              # docker CLI, so `make manifest` (docker buildx imagetools) is not
+              # usable here.
```
alexcastilio: This is duplicating the image/platform mapping logic of `make manifest` that is still used in the Release workflow. Future platform changes/additions could be updated only in one by mistake. Could we make this as a script using ORAS that both pipelines would call to prevent any future divergence?

.azure-pipelines/ci.yml
nddq: I think for merge queue, if a job failed then the PR will get kick out anyway. I can add a `$(System.JobAttempt)` to the tag if the goal is that we are able to retry jobs on manual runs + PR checks

.azure-pipelines/ci.yml
nddq: Fixed in fe2c1cef. The `manifest-*` targets take the create command from a new `MANIFEST_CREATE` variable, which defaults to `docker buildx imagetools create -t`. The ADO e2e job calls `make manifest` with `hack/scripts/oras-index.sh` as the backend, because the OneBranch container has no docker CLI. The GitHub workflows and the release path do not change.

.azure-pipelines/ci.yml
alexcastilio: I don't think `$(System.JobAttempt)` will solve the issue as this would have to be passed on to later stages/jobs regarding each build job.

nddq: i have that fix in the oras-index.sh script actually, the script will probe <tag>-<platform>-1, -2 for each platform and uses the highest tag that exists for manifest creation

hack/scripts/oras-index.sh
```diff
@@ -0,0 +1,60 @@
...
+# newest_ref <ref>: the reference with the highest existing attempt suffix,
+# or <ref> itself when no attempt tag exists. Ten attempts of one job in one
+# run is the bound; every retry is a manual action.
+newest_ref() {
+  local ref="$1" n found=""
+  for n in 1 2 3 4 5 6 7 8 9 10; do
+    if oras manifest fetch --descriptor "$ref-$n" >/dev/null 2>&1; then
+      found="$ref-$n"
+    fi
+  done
```
alexcastilio: Every failed fetch is treated as missing tag, so any auth, timeout or other error can select an older image. What if we fetch all tags (`oras repo tags <image>`), filter by the preffix and get the highest value? That might work

alexcastilio (approved):

